PR #5, opened 2017-01-23
Suggested Updates to the README

germonprez:
Working on the README to provide more detail. Things will have to be changed with respect to technical details. This also implies that we are checking the application against Travis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 branch | status
 --- | ---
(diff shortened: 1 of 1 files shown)


howderek:
@germonprez What is wrong with the pip install from github?

germonprez:
Nothing wrong with it. Happy to change. 

Yea, I think that the install script should be in this repo.

howderek:
@germonprez Sounds good!

PR #10, opened 2017-02-01
Use docstrings and ignore egg-info.

jakeharding:
Uses docstrings for setup.py documentation and adds a couple entries to .gitignore.


howderek:
Thank you!

PR #11, opened 2017-02-01
Add license

srobins259:
I added a license to the __init__.py file


howderek:
@srobins259 I can't merge this into master because it's based on exploratory and we're not ready to roll that out yet.

PR #13, opened 2017-02-01
Added License to the top of cli.py

bparish628:
Added the license that should be in every file to the `cli.py`.


howderek:
Thanks!

PR #16, opened 2017-02-11
Front-end Implemented

howderek:
Hello all!

I implemented a front-end interface to GHData! It is very basic, it only shows forks, commits, and stargazing trends, but it illustrates the direction we might head in the future. I encourage all of you to `git checkout dev && git pull` then try playing around with `ghdata`

I will work on creating a public-facing version of GHData soon!

Thanks!
Derek


howderek:
Here is a screenshot of what it looks like if you don't have `msr14` and want to know before it's live online:

screenshot 2017-02-10 20 06 12

germonprez:
Nice! Thanks for pushing this forward @howderek

PR #19, opened 2017-02-15
Added issues to 'update indicators based on LFOSLS

GeorgLink:


howderek:
@GeorgLink Looks good and feel free to push commits directly to dev!

PR #81, opened 2018-02-23
Moved logo to static directory

b-hodges:
Moved `logo.png` to its expected location


howderek:
The public directory is updated by Brunch and is only tracked by Git for an easier time for our Docker deployment. Both the Brunch server and Docker NGINX configuration will serve the file correctly. Please use `brunch watch -s` for local testing (this will also proxy your API requests so it will work, it won't work if you just open the HTML file regardless of if the paths are correct).

PR #101, opened 2018-05-02
Frontend parameter fixes

b-hodges:
Some of the frontend metric visualizations (bubble chart and stacked bar chart) didn't respect the props passed into them


howderek:
Thank you so much :)

PR #102, opened 2018-05-02
 Added simple guide for adding non-timeseries metrics

b-hodges:
The guide is currently very rudimentary and requires that a new vue be created for each type of endpoint


howderek:
Thank you!

PR #107, opened 2018-05-18
Updates to jupyter notebook and ghtorrent

kmn5409:
Hopefully this goes fine
(I completed some examples using ghtorrent.py file and added that to the visualization in BaseRepoActivityCard and also added some more examples for the jupyter notebooks)


howderek:
Look good, thanks!

PR #115, opened 2018-06-06
Update server.py

kmn5409:
Unexpected indent at line 973,
will cause an issue when running the command:
augur


howderek:
Hi @kmn5409 thank you for bring this to our attention, this was fixed in bd4d683

PR #173, opened 2018-09-18
Remove second "cd augur" in install instructions

Nebrethar:

The makefile is in the root directory;
make would not work in augur/augur.


howderek:
Thank you @Nebrethar !

PR #192, opened 2018-11-14
Create default augur.config.json

Nebrethar:
I could not find an augur.config.json file in the repository, but the dev-install guide refers to the file.

This is a default version that acts as a template for a new config.

Describe you pull request here:

### DIFF
--- a/augur.config.json
+++ b/augur.config.json
@@ -23,22 +23,11 @@
         "name": "name",
         "pass": "password",
         "port": "0000",
-        "user": "username"
-    },
-    "GHTorrentPlus": {
-        "host": "hostname.io",
-        "name": "name",
-        "pass": "password",
-        "port": "0000",
-        "user": "username"
-    },
-    "Git": {
-        "refresh": "3600",
-        "repositories": [
-            "https://github.com/*",
-            "https://github.com/*"
+        "projects": [
+            "project",
+            "project2",
         ],
-        "storage": "runtime/git_repos/"
+        "user": "username"
     },
     "GitHub": {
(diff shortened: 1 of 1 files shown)


augur.config.json
```diff
@@ -0,0 +1,47 @@
...
+            "project",
+            "project2",
+        ],
+        "user": "username"
+    },
+    "GitHub": {
+        "apikey": "0000000000000000000000000000000000"
+    },
+    "LibrariesIO": {
+        "apikey": ""
```
ccarterlandis: The connection to LibrariesIO is now through a database, so I think it would be fine to have this section mirror the earlier `Database` section, something like: 
```suggestion
                "host": "hostname.io",
         "name": "name",
         "pass": "password",
         "port": "0000",
         "user": "username"
```

augur.config.json
```diff
@@ -0,0 +1,47 @@
...
+    },
+    "Plugins": [],
+    "Server": {
+        "cache_expire": 3600,
+        "host": "0.0.0.0",
+        "port": "0000",
```
ccarterlandis: I think that 5000 is an okay default port. If we want to change this later, we can.
```suggestion
        "port": "5000",
```

ccarterlandis:
@Nebrethar if you're comfortable making the proposed changes, then I feel comfortable merging.

ccarterlandis:
@Nebrethar: as part of this effort to make everything clearer, I can write up a small section covering the configuration file in the developer docs as well. I think this is a sensible default, thanks for doing this!

PR #193, opened 2018-11-14
Move all plugins in augur/plugins/core to augur/plugins

ccarterlandis:

I propose that we move all plugins to a single directory under `augur/plugins`. This will make testing easier and be more welcoming for newcomers, since they won't have to decide where to put their plugin.


ccarterlandis:
This creates some licensing issues, of which I was unaware when making the change.

PR #213, opened 2019-01-03
Hotfixes from Risk Branch

Nebrethar:
_remove f before quotes in application.py and metrics_status.py_
Augur would not run with the `f` in place. I had to remove them to get Augur to work.

_added a .decode('utf-8') to json.loads() parameters for readable responses_
This enabled the graphs outside of git metrics. Otherwise, json.loads was trying to run a bytes object.


ccarterlandis:
I'm unable to reproduce these errors in `dev`. If you check out the latest version of `dev`, does Augur work for you @Nebrethar?

Nebrethar:
The steps I take to reproduce this issue are:

-Install fresh for Ubuntu (including necessary conda environment repair and changing pip->pip3 in makefile to avoid installing to py2.7)
-gir checkout dev
-`Make dev` or `augur` command both produce issues
-Apply these fixes to resolve runtime issues

This is performed on Ubuntu 16.04 / 18.04 ( I have performed the fix on both). I've done this mostly with my area's local augur server.

Nebrethar:
I also just recently started to run into this issue. The first time I saw it was probably late November.

ccarterlandis:
Okay, maybe the issue is related to Ubuntu then? I haven't run into this (I'm running macOS Mojave). That seems weird to me that these Python issues are only occurring for you. I've not had to change `pip` to `pip3` in the Makefile or apply any of the other changes, and everything works for me.

However, our hosting server runs Ubuntu 16.04 and has no problems with the latest version of `dev` which doesn't help my suspicions. I'll continue to look into this.

In the meantime, any ideas @howderek?

augur/datasources/metrics_status/metrics_status.py
```diff
@@ -19,15 +19,15 @@ def __init__(self):
 
     def determine_frontend_status(self, metric):
         attribute = None
-
+        
         if metric.metric_type == "timeseries":
-            attribute = next((attribute for attribute in self.timeseries if f"/api/unstable/<owner>/<repo>/timeseries/{attribute[2]}" == metric.endpoint), None)
+            attribute = next((attribute for attribute in self.timeseries if "/api/unstable/<owner>/<repo>/timeseries/{}".format(attribute[2]) == metric.endpoint), None)
```
ccarterlandis: I had to change these 3 `if` statements slightly, as removing the `f` disables the necessary string interpolation. The solution, although less readable (IMO), works just fine and is okay by me.

PR #236, opened 2019-02-13
Add preinstall script to Makefile for convenience

Nebrethar:

This is a script that attempts to run all preinstall operations for the user, and stops if a command fails to execute as expected.

In the case that this is accepted, the installation docs will also need to be updated to include this option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,9 @@ default:
 # 
 #  Installation
 #
+preinstall:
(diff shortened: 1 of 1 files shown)


ccarterlandis:
I'm not quite sure I understand the idea behind the `preinstall` concept (that's probably just a me thing). We don't explicitly depend on Anaconda to make everything work; that is, one can get the project running just fine without it, but it makes the Python version management easier. Additionally, I think that just attempting to install things and observing what fails and what doesn't is just fine. If I'm missing something, by all means please correct me!

PR #257, opened 2019-03-26
Implement repository_size in githubapi

parthsharma2:
Implemented the metric "repository size" in githubapi. This implementation returns the size of the repository in KBs.

### DIFF
--- a/augur/datasources/githubapi/githubapi.py
+++ b/augur/datasources/githubapi/githubapi.py
@@ -170,6 +170,21 @@ class GitHubAPI(object):
     ###           ACTIVITY            ###
     #####################################
 
+    @annotate(tag="repository-size")
+    def repository_size(self, owner, repo):
+      """
+      Calculate the repository size.
+      
+      :param owner: Username of the repository owner.
+      :param repo: Name of the repository.
+      :return: Size of repository in KBs.
+      """
+
+      url = 'https://api.github.com/repos/{}/{}'.format(owner, repo)
+      response = requests.get(url, auth=('user', self.GITHUB_API_KEY))
+
+      return {'size': response.json()['size']}
+
 
     #####################################
     ###         EXPERIMENTAL          ###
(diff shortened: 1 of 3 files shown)


sgoggins:
@ccarterlandis : It seems like there is an issue with database connectivity when I look at the Travis build. I just reran it. Can you assess wha'ts going on? I think its probably one thing in all cases.

parthsharma2:
@sgoggins AFAIK currently the Travis Build is configured to use Environment Variables rather than the values from the config files. This results in the following

> Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.
> Pull requests sent from forked repositories do not have access to encrypted variables or data.
https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

parthsharma2:
I think the main issues why the Travis CI build fails is environment variables. Since the database configuration is stored as environment variables and my pull request is external (i.e. not from the same repository or from someone who has write access to the repo), Augur isn't able to read the environment variables and connect to the database. @ccarterlandis @sgoggins could you try running the Travis Build using the `augur.config.json` to load the database config instead of environment variables?

ccarterlandis:
@parthsharma2 regarding the Travis build problems: the `augur.config.json` is not checked into Git, meaning that Travis will not be able to use it. We also can't unencrypt the variables, as we don't want to expose our credentials to the public! As such, I don't see a way around this issue for now, but I do very much want to investigate this further and find a solution.

sgoggins:
@parthsharma2 : congrats on having your GSOC pull request merged! I will be looking for your completed app in my augur project inbox! @ccarterlandis : Thanks for the assist!! :)

parthsharma2:
Thank you @sgoggins @ccarterlandis. I look forward to working with you all :smile:

PR #260, opened 2019-03-30
fix typo in dev-guide

bing0n3:
In [dev-guide 2.3](http://augur.augurlabs.io/static/docs/dev-guide/2-install.html#augur-installation-instructions), there is a typo about granting privileges. 

The original one in [dev-guide](https://github.com/chaoss/augur/blob/master/docs/python/source/dev-guide/2-install.rst) is:
`GRANT ALL PRIVILEDGES ON augur.* TO 'augur'@'localhost';`

However, It should be:
`GRANT ALL PRIVILEGES ON augur.* TO 'augur'@'localhost';`


ccarterlandis:
Hey @bing0n3, thanks for catching this typo. I switched the base to point to `hotfix`, as that's where we do all our releases from. I'll take a look at your other PR soon!

PR #278, opened 2019-05-03
fixing commits_by_week endpoint to be what it says it is

sgoggins:
Looking over this endpoint I think it should reflect commits_by_week. It included author and affiliation information, which makes it "noncompliant" with our API practices associated with dates. This converts the `week of the year` and `year` data in the facade schema to the Sunday of that week and year and removes the author information. It can now be visualized as "commits by week". 

Next Gen Augur will have different queries, but this one makes this endpoint useful, IMHO.


sgoggins:
I broke the API tests and the Python tests ... will change those.

PR #283, opened 2019-05-20
Broker

sgoggins:
The version of ipykernel in our requirements is no longer compatible or available. THis update makes it so augur will work.


sgoggins:
@ccarterlandis : We need update the tests to have this Jupyter version: 

 ipykernel==5.5.1

PR #297, opened 2019-06-19
Initial repo_group_id and repo_id when Repo construct

bing0n3:
In the frontend,  function `Repo(repo)` called to construct repo. 

https://github.com/chaoss/augur/blob/d2ba1806c8f510d5ea122d510cdb0322dc3f6da7/frontend/app/Augur.js#L70-L71

But `repo_group_id` and `repo_id` are not always passed, it usually only commit `girurl` or `githuburl` by Vuex. For example:
https://github.com/chaoss/augur/blob/d2ba1806c8f510d5ea122d510cdb0322dc3f6da7/frontend/app/Augur.js#L219-L222

I added a new function in function `Repo` to fetch `repo_group_id` and `repo_id`. It will query `repo_group_id` and `repo_id` from `augur_db` by `owner` and `name`.

### DIFF
--- a/augur/datasources/augur_db/augur_db.py
+++ b/augur/datasources/augur_db/augur_db.py
@@ -826,3 +826,21 @@ class Augur(object):
         """)
         results = pd.read_sql(closedIssueCountSQL, self.db)#, params={"rg_id": '{}'.format(rg_id)})
         return results
+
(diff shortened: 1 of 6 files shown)


gabe-heim:
This is very helpful, thank you Bingwen

PR #301, opened 2019-06-22
Optimize frontend to prevent duplicate requests

bing0n3:
Optimize charts and card to prevent duplicate requests:
* GitCard
* CommitChart
* HorizontalBarChar
* LinesOfCodeChart
* NormalizedStackedBarChart
* OneDimensionalStackedBarChart
* TickChart

### DIFF
--- a/frontend/app/Augur.js
+++ b/frontend/app/Augur.js
@@ -58,20 +58,25 @@ export default function Augur () {
         state.gitRepo = payload.gitURL
         state.baseRepo = payload.gitURL
         state.hasState = true
-        let repo = window.AugurAPI.Repo(payload)
-        state.baseRepo = repo.toString()
-        if (!window.AugurRepos[repo.toString()]) {
+        let repo = null
+        let repoName = gitUrlToString(payload)
+        if (!window.AugurRepos[repoName]) {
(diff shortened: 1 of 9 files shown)


sgoggins:
@gabe-heim : does the batch API implementation you did cover any of this? 
Also, I notice that the "total lines of code" on the line chart never changes no matter where you hover. : http://mudcats.augurlabs.io:3333/single/github.com%2Fropensci%2Fplotly/git#

sgoggins:
@ccarterlandis and @gabe-heim : THis looks like reasonable VUE and .jS optimizations. We don't have tests for the front end (that's hard, and most web apps have little) .. but the travis build fails on a missing library `zmq` ... ? @bing0n3 : Is this something added somewhere else in a different pull request that you know of that needss to be added to requirements.txt ?

bing0n3:
@sgoggins This [commit](https://github.com/chaoss/augur/commit/95d3f1fd36a802f6fad6a4559cd27db6b8cdff02) by Gabe imported the library  `zmq`.  I guess we need to add the library into `requirements.txt`

sgoggins:
This looks like it will reduce API calls and not change how we are getting data from the API.

PR #324, opened 2019-07-13
Fix gh_repo_info_worker

parthsharma2:
This fixes the gh_repo_info_worker to fetch tasks from its maintain queue & also send task completion confirmation to the broker allowing the broker to send more tasks to this worker.

This worker can now be set to run periodically by specifying a housekeeper job in `augur.config.json`

Signed-off-by: Parth Sharma <parth261297@gmail.com>


parthsharma2:
@sgoggins updated the readme

PR #340, opened 2019-08-03
Dev patch 2

sgoggins:
Fixes an issue with commit counting when there is more than one commit on a day, and there is more than one file committed, and there are empty commit records that do not specify files (i.e., merged pull requests, etc.).


sgoggins:
Merging my own PR so that Dev does not get out of sync during development.

PR #222, opened 2019-01-22
Implement metrics in githubapi

parthsharma2:
Implemented the following metrics:

- **Code Commits:** This implementation returns the number of code commits per day.
- **Contributors:** This implementation returns a list of contributors along with their respective contributions to a project.

PR #356, opened 2019-08-17
Added script to create augur.config.json

parthsharma2:
Created a script that allows you to interactively create `augur.config.json`. To run the script execute the following command 
```
make config
```

Signed-off-by: Parth Sharma <parth261297@gmail.com>


ccarterlandis:
So it took me entirely too long to remember this, but it's failing because we've configured the Travis build to remove our encrypted environment variables from **any** build that doesn't come from within the main augur repository, as we don't want untrusted builds to have access to our credentials. That's not to say we don't trust you @parthsharma2 - in fact, we very much trust you! It's just a matter of precaution. @sgoggins You also might recall that [we ran into this issue](https://github.com/chaoss/augur/pull/222) back in January for the same reason...my vote is we just adopt the same solution we did last time 😄

PR #359, opened 2019-08-19
API endpoints for issue comments

parthsharma2:
Added the following API endpoints:
- issue-comments-mean: returns the mean/average of issue comments
- issue-comments-mean-std: returns the mean & standard deviation of issue comments


parthsharma2:
@sgoggins added tests

gabe-heim:
Thank you Parth! Looks great. We will eventually want to add a 'day' group by option for the issue_comments_mean metric, but we can worry about that later.

Issue #366, opened 2019-08-31
Updates to creation of augur.config.json on install. 


sgoggins:
**Is your feature request related to a problem? If so, please describe the problem:**
Install script should create default worker and housekeeper blocks. 

**Describe the solution you'd like:**

    "Facade": {
        "check_updates": 1,
        "clone_repos": 1,
        "create_xlsx_summary_files": 1,
        "delete_marked_repos": 0,
        "fix_affiliations": 1,
        "force_analysis": 1,
        "force_invalidate_caches": 0,
        "force_updates": 1,
        "limited_run": 0,
        "multithreaded": 0,
        "nuke_stored_affiliations": 0,
        "pull_repos": 1,
        "rebuild_caches": 1,
        "run_analysis": 1
    },
    "Workers": {
        "facade_worker": {
            "port": 51246,
            "switch": 0,
            "workers": 1,
            "repo_directory": "**WE NEED TO PROMPT FOR THIS"
        },
        "pull_request_worker": {
            "port": 51252,
            "switch": 0,
            "workers": 1
        },
        "github_worker": {
            "port": 51238,
            "switch": 0,
            "workers": 1
        },
        "insight_worker": {
            "port": 51244,
            "switch": 0,
            "workers": 1
        },
        "repo_info_worker": {
            "port": 51242,
            "switch": 0,
            "workers": 1
        }
    },
    "Housekeeper": {
      "jobs": [
      	{
                "delay": 150000,
                "given": ["git_url"],
                "model": "issues",
                "repo_group_id": 0
        },
	{
                "delay": 150000,
                "given": ["git_url"],
                "model": "repo_info",
                "repo_group_id": 0
        },
	{
                "delay": 150000,
                "given": ["git_url"],
                "model": "pull_requests",
                "repo_group_id": 0
        }
      ]
    } 
}

**Describe alternatives you've considered:**
N/A

parthsharma2:
I'll work on this

gabe-heim:
Defaults added in #373

PR #375, opened 2019-09-03
add default worker & housekeeper block values in make_config.py

parthsharma2:
- Rename `make-config.py` to `make_config.py`
- Add default worker & housekeeper block values in `make_config.py` as specified in #366 
- Removes user input prompt for GitHub API Key


ccarterlandis:
LGTM! Thanks again Parth!

PR #384, opened 2019-09-10
Fix error messages in Vue compilation

Nebrethar:
Signed-off-by: Matt Snell <msnell@unomaha,edu>


Nebrethar:
Closing because I realized this is no longer a necessary development. It looks like this problem has already been taken care of

PR #450, opened 2019-10-29
Facade, documentation, and installation updates

sgoggins:


util/scripts/install/install.sh
```diff
@@ -12,18 +12,25 @@ if [[ $? -ne 0 ]]; then
   exit 1
 fi
 
+mkdir logs
+
 # backend
-util/scripts/install/backend.sh
+util/scripts/install/backend.sh > logs/backend.log
```
ccarterlandis: We need to use a different file for the backend install log. `logs/backend.log` is used as the log file for the backend server process while it runs.

util/scripts/install/setup_db.sh
```diff
@@ -44,10 +44,14 @@ function get_api_key_and_repo_path() {
   read -p "Facade repo path: " facade_repo_path
   echo
 
+  while [ ! -d "$facade_repo_path" ]; do
```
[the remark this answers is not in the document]
ccarterlandis: @sgoggins Let's discuss when we meet next. This comment just now came through, this is the one I send you on Slack earlier

util/scripts/install/install.sh
```diff
@@ -12,18 +12,25 @@ if [[ $? -ne 0 ]]; then
   exit 1
 fi
 
+mkdir logs
```
ccarterlandis: If we are going to pipe the output to log files, then we need to add indications that we're doing so and that the process is still going while the output is being piped. It's very confusing behavior, I restarted the install process several times because I thought it was hanging before I realized that everything was being piped to log files.

util/scripts/install/nomos.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# if [[ "$(type -t gcc)" == "" ]]; then
+#     echo "[$0] Couldn't find gcc in your PATH!"
+#     echo "[$0] Please install gcc before continuing."
+#     exit 1
+# fi
+
+REPO=$(mktemp -d)
+echo "[$0] git clone https://github.com/fossology/fossology $REPO"
+git clone https://github.com/fossology/fossology $REPO || exit 1
```
ccarterlandis: Is cloning fossology the only way to install `nomos`? Since it's so massive, it makes the install take a really long time for me, and like I mentioned above I stopped and restarted several times because I thought there was an error. If it's the only way to install it, then as a I'd like to see some indication that progress is being made and that it might take a while to finish.

ccarterlandis: BTW, this is not me suggesting we find another way, I'm just curious as I'm unfamiliar with `nomos` and `fossology`. 😄

util/scripts/install/install.sh
```diff
@@ -12,18 +12,25 @@ if [[ $? -ne 0 ]]; then
...
+
 # backend
-util/scripts/install/backend.sh
+util/scripts/install/backend.sh > logs/backend.log
+
+# nomos scanner for dosocs
+util/scripts/install/nomos.sh > logs/nomos.log 2>logs/nomos.err
```
ccarterlandis: `nomos` does not get installed correctly using this script on my machine. I get [this](https://pastebin.com/qSJKKwpX) following stack trace. I'm running OSX 10.14.6 and I'm using the `fish` shell version 3.0.2.

util/scripts/install/setup_db.sh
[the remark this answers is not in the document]
sgoggins: I commented this out @ccarterlandis

util/scripts/install/install.sh
sgoggins: I changed the logfile name to backend-install.log

util/scripts/install/install.sh
sgoggins: I think its confusing only to us, those who test the install script. I think to ordinary newbies it reduces a lot of overwhelming output.

util/scripts/install/nomos.sh
sgoggins: That is the way. And there is no automation at this point, and because of the complexity I plan to write the docs with this as a clearly optional module.

util/scripts/install/install.sh
sgoggins: Yeah, right now its a shit show on the Mac. I am working through those issues. I want to include the script in dev without calling it automatically in order to enable deployment of augur-sbom on Ubuntu servers.

sgoggins: Pushing it to master will help us get others involved in evaluating the new functionality. Because its experimental until we resolve the mac issues, is not automatically installed, and increasingly vital on Ubuntu, I would like to push this forward.

util/scripts/install/setup_db.sh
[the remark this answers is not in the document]
sgoggins: @ccarterlandis : would it be acceptable if we told the user that we are going to create the subdirectory if it does not exist? Thus far we are the only people who run the install where the repos are actually already cloned. Its a rather unusual use case.

util/scripts/install/nomos.sh
ccarterlandis: Gotcha. Unfortunate as it may be, we'll work with what we've got! 👍

util/scripts/install/install.sh
ccarterlandis: While I do agree that displaying **all** output is too overwhelming for most users, when I'm installing something I expect a certain level of feedback that at least something is happening. Starting the installation and then seeing nothing happening on my screen makes me think nothing is actually happening. I think the solution could be as simple as adding an `echo` statement before each call to an installation section (like `echo "Installing backend dependences..."`) and then adding another when it executes (like `echo "Done!"`). The output would then look something like this:
```
Installing backend dependencies...
Done!
Installing workers...
Done!
Installing nomos...
Done
etc, etc
```
which provides a basic level of feedback that tells people it is in fact working without being overwhelming.

util/scripts/install/setup_db.sh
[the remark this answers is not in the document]
ccarterlandis: I think that's the perfect solution @sgoggins 👍

util/scripts/install/install.sh
ccarterlandis: FWIW, I've never been able to get `nomos` installed on macOS. Given, I did try almost a year ago, but the error I got now does look very similar to the one I got then. It makes me extremely uneasy to push something that doesn't work to `master` but since it's non-mission-critical (well at least probably not, for most users) and this PR contains a lot of other risk work, I'm willing to put aside my fears for now. 

That being said, IMO our number one focus after the call today ought to be either getting `nomos` to install on macOS, or to make the install optional and put in a caveat that it won't work on macOS, at least for the time being.

ccarterlandis:
@sgoggins once the tests pass I will merge!

Issue #466, opened 2019-11-13
JSON SPDX Document Beautification


Nebrethar:
**Describe the bug:**
This one is a continuation of #416 where the SPDX Document button works now... but the output needs beautification.

**To Reproduce:**
1. Go to zephyr-rtos Repo Group
2. Go to zephyr-rtos/zephyr
3. Click on Risk
4. Click on the SPDX Document Download button

**Expected behavior:**
Beautiful JSON output to a file.

**Unexpected behavior:**
Single-line JSON output to a file.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome 75

**Additional context**
This is mostly a "note to self"

PR #467, opened 2019-11-14
Quick Fix for JSON SPDX Document

Nebrethar:
Thanks, Gabe H!

This is a one-liner that prints the SPDX Document in the correct format for viewing.


Nebrethar:
This fixes problems introduced in issue #466

PR #470, opened 2019-11-17
Update badging worker to use JSON field in DB

ccarterlandis:
Under the previous implementation, the `linux_badge_worker` would break everytime the CII badging API would add new fields. This implementation switches to using the `jsonb` field type in the `repo_badging` table, and pulls only the columns we need from the result. This way, we are no longer bound to the structure of the CII badging API result.

I'm marking this as a draft pull request until we can confirm that the appropriate database schemas have been updated to accommodate this change. I've included the SQL I used to create the table with the correct structure below:

```
CREATE TABLE augur_data.repo_badging
(
    badge_collection_id bigint NOT NULL DEFAULT nextval('augur_data.repo_badging_badge_collection_id_seq'::regclass),
    repo_id bigint,
    data jsonb,
	created_at timestamp(0) without time zone DEFAULT CURRENT_TIMESTAMP,
    tool_source character varying(255) COLLATE pg_catalog."default",
    tool_version character varying(255) COLLATE pg_catalog."default",
    data_source character varying(255) COLLATE pg_catalog."default",
    data_collection_date timestamp(0) without time zone DEFAULT CURRENT_TIMESTAMP
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE "augur_data"."repo_badging" OWNER TO "augur";

-- ----------------------------
-- Primary Key structure for table repo_badging
-- ----------------------------
ALTER TABLE "augur_data"."repo_badging" ADD CONSTRAINT "repo_badging_pkey" PRIMARY KEY ("badge_collection_id");

-- ----------------------------
-- Foreign Keys structure for table repo_badging
-- ----------------------------
ALTER TABLE "augur_data"."repo_badging" ADD CONSTRAINT "fk_repo_badging_repo_1" FOREIGN KEY ("repo_id") REFERENCES "augur_data"."repo" ("repo_id") ON DELETE NO ACTION ON UPDATE NO ACTION;

GRANT ALL ON TABLE augur_data.repo_badging TO augur;

COMMENT ON TABLE augur_data.repo_badging
    IS 'This will be collected from the LF’s Badging API
https://bestpractices.coreinfrastructure.org/projects.json?pq=https%3A%2F%2Fgithub.com%2Fchaoss%2Faugur
';
```

The badging worker also currently does not have any idea of which augur `repo_id` values correspond to which the CII project `id` values. This is another reason this is marked as a draft; the metric will not work until this logic is implemented.

### DIFF
--- a/augur/metrics/repo_meta/repo_meta.py
+++ b/augur/metrics/repo_meta/repo_meta.py
@@ -189,8 +189,7 @@ def cii_best_practices_badge(self, repo_group_id, repo_id=None):
     # Welcome to the Twilight Zone
     if repo_id:
         cii_best_practices_badge_SQL = s.sql.text("""
-            SELECT repo_name, rg_name, repo_badging.badge_level, achieve_passing_status,
-                achieve_silver_status, tiered_percentage, repo_url, id, repo_badging.updated_at as date
+            SELECT repo_name, rg_name, repo_badging.data, repo_badging.created_at as date
             FROM repo_badging, repo, repo_groups
             WHERE repo.repo_group_id = repo_groups.repo_group_id AND repo.repo_id = repo_badging.repo_id
             AND repo_badging.repo_id = :repo_id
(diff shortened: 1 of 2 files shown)


sgoggins:
@ccarterlandis : I changed the repo_badging structure in the augur-db on neko. I think that should cause tests to pass here.  ANd fail on dev. So if we could get this out of draft when you're ready. Over.

Nebrethar:
Thanks for taking care of this @ccarterlandis. LGTM. I will work on adding this to the frontend once we have some data in there.

PR #570, opened 2020-03-02
Added unit test & Helper Module[Github Worker Function]

mrsaicharan1:
The `check_duplicates` function was tightly coupled to the GitHub Worker which proved as a herculean task to test the function(THe Github Worker would require the initialization of the postgres DB engine, broker and requires API calls to a couple of OAuth rate-limiting endpoints). 
    Further, we might need additional helper functions in the future which might be used by other modules and workers. This would also support DRY principles. Henceforth, I suggest that we separate the helper methods to promote a smooth & swift testing process.

- Added unit test for check duplicates. Tested the function locally with pytest
- Separated`check_duplicates` into a helper function module


gabe-heim:
The unit test is a great idea! And it is something we have wanted to do for a long time. For the `check_duplicates` helper function, I agree that this should be brought out as a helper method that can be used across workers. However, if it can be used across workers, it cannot be isolated in the `github_worker` directory. My suggestion would be to move it to the `workers/standard_methods.py` file, which is where other functions with the purpose of being used across multiple different workers are kept. The unit test for any of the functions in `standard_methods.py` should also be in the `workers` directory (`workers/tests`) to keep the hierarchy in order, but I also believe we have the need for different other unit tests that should be isolated in the `github_worker` directory, such as for ensuring the whole issue collection process is functioning properly.

mrsaicharan1:
> The unit test is a great idea! And it is something we have wanted to do for a long time. For the `check_duplicates` helper function, I agree that this should be brought out as a helper method that can be used across workers. However, if it can be used across workers, it cannot be isolated in the `github_worker` directory. My suggestion would be to move it to the `workers/standard_methods.py` file, which is where other functions with the purpose of being used across multiple different workers are kept. The unit test for any of the functions in `standard_methods.py` should also be in the `workers` directory (`workers/tests`) to keep the hierarchy in order, but I also believe we have the need for different other unit tests that should be isolated in the `github_worker` directory, such as for ensuring the whole issue collection process is functioning properly.

I totally agree with the decision to move it to the `standard_methods`. Thanks for the review! 
And yes, if we could increase the test coverage, it would be great. I'll start working on it soon after this :)

ccarterlandis:
@mrsaicharan1 sounds like a great plan! When you do get to the other tests, just worry about making them work on your local machine for right now. I will try to have a Travis workaround working before you're done, but we can cross that bridge when we get there!

mrsaicharan1:
I've shifted the method to the appropriate place. I believe this is okay now :D. @gabe-heim @ccarterlandis

gabe-heim:
Looks great, thanks for the contributions! @mrsaicharan1

Issue #581, opened 2020-03-05
Mouse flickers and tooltip not displayed


aksh555:
**Describe the bug:**
I was trying to solve #541 when I noticed that when hovering over the "View Insight Details ->" button in the Insights page, the mouse kept flickering and the tooltip wasn't shown. It shows only when the mouse is placed in the left corner of the button, and **thus the button cannot be clicked from anywhere else.**

**To Reproduce:**
1. Go to 'Insights Page'
2. Hover over any button that says "View Insight Details ->" and try clicking from the right side.
3. Apart from the left side of the button, mouse flickers and tooltip not shown, unable to click the button.

**Expected behavior:**
Just like the other buttons "Evolution" and "Risk", the tooltip should be visible when the mouse hovers over any part of the button and be clickable, and the mouse should not flicker. 

**Screenshots**
I have attached a video which would show the issue best.
https://vimeo.com/395679729

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser: Chrome Version 80.0.3987.122

**Additional Context**
Tested on http://augur.osshealth.io/

gabe-heim:
Fixed in #588 by @aksh555 , thank you again!

PR #582, opened 2020-03-05
Fixed tooltip  click issue

aksh555:
I think #581 can be resolved by removing the placement option of the tooltip.
This causes the tooltip to appear above the button, but no issue in clicking the button.


gabe-heim:
But regardless, good catch! Thanks so much @aksh555 , make the change to the dev branch and I will merge it

aksh555:
> I approved changes too quickly, this PR needs to be into the `dev` branch, not `master`

Oh yes, will make the change in `dev` branch.

PR #596, opened 2020-03-09
Added unit tests for Standard Worker Methods

mrsaicharan1:
Wrote tests for methods used across multiple workers
Segregated the `dump_queue` method from the workers. Also enforced the DRY principle(the same method was written 3 times across 3 worker classes).


mrsaicharan1:
Screenshot 2020-03-09 at 11 22 16 PM

ccarterlandis:
I'm running some data collection right now, when it's finished I will checkout your PR and test it locally @mrsaicharan1. I've made some changes to the workers in the past few days and I want to make sure they still work in our new Docker containers correctly 😁

mrsaicharan1:
> I'm running some data collection right now, when it's finished I will checkout your PR and test it locally @mrsaicharan1. I've made some changes to the workers in the past few days and I want to make sure they still work in our new Docker containers correctly 😁

Definitely! If there’s any task I could take up to enhance the workers,  I’d be happy to do it

PR #603, opened 2020-03-13
Refactor augur process control commands

ccarterlandis:
Augur's process control commands are now written in Python to enhance their portability.


scripts/install/install.sh
```diff
@@ -21,11 +21,11 @@ target=${1-prod}
 
 if [[ $target == *"dev"* ]]; then
```
sgoggins: ok .. this is good to know .... that the default is not dev is something to keep in mind ... what's the syntax?

PR #608, opened 2020-03-13
Update Travis build to use test data Docker image

ccarterlandis:
This PR switches all the testing IDs to be the ones of the new testing database, and refactors the Travis CI build to use the testing database Docker image as well. This makes the testing environment available to all developers, both locally and in any PRs they submit.

### DIFF
--- a/augur/metrics/issue/issue.py
+++ b/augur/metrics/issue/issue.py
@@ -111,7 +111,7 @@ def issues_first_time_closed(self, repo_group_id, repo_id=None, period='day', be
                     AND repo.repo_id = issues.repo_id
                     AND issues.pull_request IS NULL 
                     AND issues.issue_id = issue_events.issue_id
-                    And issue_events.created_at BETWEEN :begin_date AND :end_date
+                    And issue_events.created_at BETWEEN to_timestamp(:begin_date, 'YYYY-MM-DD HH24:MI:SS') AND to_timestamp(:end_date, 'YYYY-MM-DD HH24:MI:SS')
                     GROUP BY issue_events.cntrb_id, repo_name
                 ) AS iss_close
             GROUP BY issue_date, repo_name
@@ -133,7 +133,7 @@ def issues_first_time_closed(self, repo_group_id, repo_id=None, period='day', be
                     AND action = 'closed'
                     AND repo.repo_id = issues.repo_id
                     AND issues.issue_id = issue_events.issue_id
-                    And issue_events.created_at BETWEEN :begin_date AND :end_date
+                    And issue_events.created_at BETWEEN to_timestamp(:begin_date, 'YYYY-MM-DD HH24:MI:SS') AND to_timestamp(:end_date, 'YYYY-MM-DD HH24:MI:SS')
                     GROUP BY issue_events.cntrb_id, repo.repo_id, repo_name
                 ) AS iss_close
             GROUP BY repo_id, repo_name,issue_date
--- a/augur/metrics/pull_request/pull_request.py
+++ b/augur/metrics/pull_request/pull_request.py
@@ -372,7 +372,7 @@ def pull_request_acceptance_rate(self, repo_group_id, repo_id=None, begin_date=N
                     WHERE action = 'merged'
                     AND issues.pull_request IS NOT NULL 
                     AND repo_group_id = :repo_group_id
-                    AND issue_events.created_at BETWEEN :begin_date AND :end_date
+                    AND issue_events.created_at BETWEEN to_timestamp(:begin_date, 'YYYY-MM-DD HH24:MI:SS') AND to_timestamp(:end_date, 'YYYY-MM-DD HH24:MI:SS')
                     GROUP BY accepted_on
                     ORDER BY accepted_on
                 ) accepted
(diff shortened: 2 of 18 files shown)


augur/metrics/issue/issue.py
```diff
@@ -111,7 +111,7 @@ def issues_first_time_closed(self, repo_group_id, repo_id=None, period='day', be
                     AND repo.repo_id = issues.repo_id
                     AND issues.pull_request IS NULL 
                     AND issues.issue_id = issue_events.issue_id
-                    And issue_events.created_at BETWEEN :begin_date AND :end_date
+                    And issue_events.created_at BETWEEN to_timestamp(:begin_date, 'YYYY-MM-DD HH24:MI:SS') AND to_timestamp(:end_date, 'YYYY-MM-DD HH24:MI:SS')
```
sgoggins: have we tested that the issue model still works with this change?

augur/metrics/issue/issue.py
```diff
@@ -133,7 +133,7 @@ def issues_first_time_closed(self, repo_group_id, repo_id=None, period='day', be
                     AND action = 'closed'
                     AND repo.repo_id = issues.repo_id
                     AND issues.issue_id = issue_events.issue_id
-                    And issue_events.created_at BETWEEN :begin_date AND :end_date
```
sgoggins: have we tested that the issue model still works with this change

augur/metrics/pull_request/pull_request.py
```diff
@@ -372,7 +372,7 @@ def pull_request_acceptance_rate(self, repo_group_id, repo_id=None, begin_date=N
                     WHERE action = 'merged'
                     AND issues.pull_request IS NOT NULL 
                     AND repo_group_id = :repo_group_id
-                    AND issue_events.created_at BETWEEN :begin_date AND :end_date
```
sgoggins: have we tested that the pull request model still works with this change?

augur/metrics/pull_request/pull_request.py
```diff
@@ -385,7 +385,7 @@ def pull_request_acceptance_rate(self, repo_group_id, repo_id=None, begin_date=N
                     WHERE action = 'ready_for_review'
                     AND issues.pull_request IS NOT NULL 
                     AND repo_group_id = :repo_group_id
-                    AND issue_events.created_at BETWEEN :begin_date AND :end_date
```
sgoggins: have we tested that the pull request model still works with this change?

augur/metrics/pull_request/pull_request.py
```diff
@@ -405,7 +405,7 @@ def pull_request_acceptance_rate(self, repo_group_id, repo_id=None, begin_date=N
                     WHERE action = 'merged'
                     AND issues.pull_request IS NOT NULL
                     AND repo_id = :repo_id
-                    AND issue_events.created_at BETWEEN :begin_date AND :end_date
```
sgoggins: etc

augur/metrics/pull_request/pull_request.py
```diff
@@ -417,7 +417,7 @@ def pull_request_acceptance_rate(self, repo_group_id, repo_id=None, begin_date=N
                     WHERE action = 'ready_for_review'
                     AND issues.pull_request IS NOT NULL 
                     AND repo_id = :repo_id
-                    AND issue_events.created_at BETWEEN :begin_date AND :end_date
```
sgoggins: etc

workers/standard_methods.py
```diff
@@ -212,6 +212,7 @@ def read_config(section, name=None, environment_variable=None, default=None, con
     :param section: location of given variable
     :param name: name of variable
     """
+    config_file_path = os.getenv("AUGUR_CONFIG_FILE", config_file_path)
```
sgoggins: this is for docker i assume?

augur/metrics/issue/issue.py
gabe-heim: This shouldn't affect any of the workers, but @ccarterlandis I am wondering why we made this change? I am assuming to make user inputs of begin and end dates more flexible?

augur/metrics/issue/issue.py
gabe-heim: @sgoggins The issue model would not be affected, this change is just in one of our issue metrics in our api, and the issue model does not hit our own api at all

augur/metrics/pull_request/pull_request.py
gabe-heim: Same situation here, the model will still work, but I do not know why this change was made for multiple metrics @ccarterlandis

augur/metrics/issue/issue.py
ccarterlandis: Yes, without adding the timestamp the dates were not being parsed correctly and the data was not being returned even it was there. However there have also been problems getting the data to gather correctly, which might have actually been the case

augur/metrics/pull_request/pull_request.py
ccarterlandis: See my comment above

test/api/test_issue_routes.py
```diff
@@ -6,233 +6,233 @@ def metrics():
...
     assert len(data) >= 1
     assert data[0]["count"] > 0
 
-def test_issues_first_time_closed_by_repo(metrics):
-    response = requests.get('http://localhost:5000/api/unstable/repo-groups/24/repos/22997/issues-first-time-closed')
-    data = response.json()
-    assert response.status_code == 200
-    assert len(data) >= 1
-    assert data[0]["count"] > 0
+# def test_issues_first_time_closed_by_repo(metrics):
```
gabe-heim: Why is this test commented out? If we do not need it anymore than we should delete it, if we need to reference it in the future, then we can look at commit history

PR #613, opened 2020-03-18
handle nonexistent groups

maxbalk:
for michael


michaelwoodruffdev:
Thanks Max, I'm going to add the necessary lines to manager.py on my branch since I've made a few other changes to the file since I've pushed to dev

PR #621, opened 2020-03-22
GSoC 2020: Adding Gitlab worker: Microtask-1

abhinavbajpai2012:
Adding Gitlab worker which fetches details of each Issue, Branch, Commit and Merge Request for all owned projects and stores them as JSON in directory named after parent project in the 'output' directory.

Signed-off-by: Abhinav <abhinavbajpai2012@gmail.com>

### DIFF
--- a/workers/abhinavbajpai2012/constants.py
+++ b/workers/abhinavbajpai2012/constants.py
@@ -1,0 +1,2 @@
+GITLAB_ADDRESS = 'https://gitlab.com/'
+OUTPUT_PATH = 'outputs'
--- a/workers/abhinavbajpai2012/utils.py
+++ b/workers/abhinavbajpai2012/utils.py
@@ -1,0 +1,95 @@
+from datetime import datetime
+from dateutil.parser import parse
+
+
+def flatten_json(json):
+    '''
+
+    Flattens a json response
+    :param json: Dictionary: json object
+    :return:
+    '''
+    out = {}
+
+    def flatten(field, name=''):
+        if type(field) is dict:
+            for item in field:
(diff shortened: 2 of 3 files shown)


workers/abhinavbajpai2012/constants.py
```diff
@@ -0,0 +1,2 @@
+GITLAB_ADDRESS = 'https://gitlab.com/'
+OUTPUT_PATH = 'outputs'
```
ccarterlandis: Hi @abhinavbajpai2012, thanks for completing this microtask! Your code looks good to me - however, since these files are generated by the script and not actual source code, can you please remove them from version control? You'll need to first remove them from Git, then add the folder to the `.gitignore`, and then add the changes again. You should be able to do this with something like:
```
# remove from git
$ git rm --cached -r workers/abhinavbajpai2012/outputs/

# then add workers/abhinavbajpai2012/outputs/ to the .gitignore
$ echo workers/abhinavbajpai2012/outputs/ >> .gitignore

# add the .gitignore changes
$ git add .gitignore
```
This should remove the files from version control, allowing others to edit them. It also prevents unnecessary merge conflicts - if I were to run the scripts on my machine and commit the changes to the gathered data, and then you try to do the same, we could possibly gather different data (there might have been a commit added after I ran my scripts), and then we would get a merge conflict in the files which would be tricky to resolve. By removing them from Git, however, they are freshly generated every time and there will be no merge conflicts. Hopefully that all made sense, let me know if you have any questions and I would be glad to help.

abhinavbajpai2012: Cool. Got it :)

PR #624, opened 2020-03-24
Gsoc sso

Nayan-Das:
I am creating a new pull request for adding the Developer Certificate of Origin.
I have added two pages login.vue and secure.vue. I also made changes in route.ts file to make my login page as home page.
Please suggest if need to update this code in a separate branch or should I keep the home page as it was earlier.
Then We have to go to the /login route to check the functionality.


Nayan-Das:
DCO not updated properly. So, I am canceling this request.

Issue #557, opened 2020-02-26
GSOC & Outreachy SSO Microtasks


sgoggins:
## Here's some ideas: 

1. Microtask 0: Familiarize yourself with augur by downloading and configuring the dev branch. For a little more context about what we are trying to accomplish with Augur's prototyping, checkout Augur's documentation at : https://oss-augur.readthedocs.io/en/dev/   Fork Augur into your own GitHub Account. Work in the dev branch. 
2. Microtask 1: Using the Augur system, implement any SSO technology of your choosing. This is a completely "proof of concept" without any expectation of a finished work. Could be a social media or github based oauth. Does not need to "protect pages" (i.e., manage the entitlements part of sign on). Its enough to have to think through SSO in a Flask environment with something potentially as simple as a "mickey mouse" sign in (i.e., username and password of "mickey mouse", but processing it through an actual authentication system).  
3. Microtask 2: Make a pull request to update the dev branch with the work from Microtask 1. If you have questions about getting started in the steps above, open an issue in the Augur repository.

Chinmay4400:
Thanks @sgoggins Getting right on it.

Nayan-Das:
Hi,
I am working on this task.I have some doubts like-
Which page should I use for SSO or should I create a separate page a redirect from there.
Could you please guide me where I can add my vue code.

Thanks,
Nayan

ccarterlandis:
Hi @Nayan-Das, I would say that for now, work on a separate page in whatever way you see fit. Since it's a microtask, we are looking more for a proof of concept/demo than a fully fledged integration. Best of luck and please reach out if there's anything else we can help with!

rajrohit10:
Hello, 
I went through this idea. I love it.I have experience in python and blockchain. Right now working on Microtask 0. What more should I do to let you guys know that I have completed microtask 0??

Nayan-Das:
Hi,

I have completed GSOC & Outreachy SSO Microtasks #557 and raised a pull request.
Could you please suggest if I should proceed with the application.

Thanks and regards,
Nayan

ccarterlandis:
Hi @Nayan-Das, we will review your pull request soon. In the meantime please feel to get started on the application.

ccarterlandis:
[...]

PR #627, opened 2020-03-24
GSOC & Outreachy SSO Microtasks #557

Nayan-Das:
**Files Added**: I have added two file login.vue and secure.vue.
**Files Modified**: I have also changed the route.ts file to make login.vue  file as home page.

**Functionality** : There is only facebook SSO implemented and one normal login with user name and password "nayan". After the login, it will navigate to the secure page. 

**Note**: _This may break the current dev.
Please suggest if I should update to a different branch or keep the route file as it is.
If I do not change the route file then we need to traverse to the /login route to check the functionality._


sgoggins:
@Nayan-Das : would you mind reissuing this PR into the GSoC branch of augur? There are a lot of changes and we want to test them before moving them to dev. I'm eager to try them out!

ccarterlandis:
@sgoggins I just changed the base on this PR so @Nayan-Das does not have to open a new one...feel free to revert if you'd rather see a new PR!

PR #628, opened 2020-03-24
Updated  Installation and Getting Started Documentation

Pogayo:
1.  Replaced the broken link in docs/source/getting-started/database.rst of unified model with [this old working link](https://oss-augur.readthedocs.io/en/dev/architecture/data-model.html).
2.  Shortened the wording in docs/source/development-guide/installation.rst on _Setting up the database_ section.

Next thing that I would like to work on is to make [these list items](https://oss-augur.readthedocs.io/en/dev/architecture/data-model.html#augur-data) to be displayed as list items (They are currently displayed as a paragraph).


docs/source/getting-started/database.rst
```diff
@@ -1,7 +1,7 @@
 Database Setup
 ===============
 
-One of the reasons that Augur is so powerful is because of our `unified data model <../schema/data-model.html>`_.
+One of the reasons that Augur is so powerful is because of our `unified data model <https://oss-augur.readthedocs.io/en/dev/architecture/data-model.html>`_.
```
ccarterlandis: This link to the schema model will actually not work once this gets merged into `dev`. Our documentation is built by branch, and this link points to the version that's currently in `dev`. Because of this, the page the link points to will get deleted when the new documentation is merged into `dev`, and since that page would not exist the link would be defunct again. We should either get rid of the link entirely and just have it be plain text, or re-implement the page in its new form under `docs/source/schema/`.

Pogayo: I think implementing it in schema is a better option because the link provides really useful information? Can I go ahead and implement it in schema?

docs/source/schema/data-model.rst
```diff
@@ -0,0 +1,100 @@
...
+Example (note this will **not**) work\:
+
+.. code:: bash
+
+    psql -h example.com -p 5432 -c "CREATE DATABASE augur;"
+    psql -h example.com -p 5432 -c "CREATE USER augur WITH ENCRYPTED PASSWORD 'password123';"
+    psql -h example.com -p 5432 -c "ALTER DATABASE augur OWNER TO augur;"
+    psql -h example.com -p 5432 -c "GRANT ALL PRIVILEGES ON DATABASE augur TO augur;"
+    psql -h example.com -d augur -U augur -p 5432 -a -w -f /Users/carterlandis/augur/schema/create_schema.sql
+
```
ccarterlandis: You can go ahead and remove this section and then just add a brief point about following the instructions in the getting started for setting a database up. This way we'll only have one place to worry about when changing the documentation for how the schema is created. The link reference should be this: `../getting-started/database.html`

docs/source/schema/data-model.rst
```diff
@@ -0,0 +1,100 @@
...
+2. ``workers/facade_worker``: Based on
+http://www.github.com/brianwarner/facade, but substantially modified in
+the fork located at http://github.com/sgoggins/facade. The modifications
+include modularization of code, connections to Postgresql data instead
+of MySQL and other changes noted in the commit logs.
+3. ``workers/insight_worker``: Generates summarizations from raw data
+gathered from commits, issues, and other info.
+4. ``workers/linux_badge_worker``: Pulls data from the Linux Foundation’s
+badging program.
+5. ``workers/code_analysis``: Populates the table
```
ccarterlandis: This worker has been renamed to the `value_worker`

docs/source/schema/data-model.rst
```diff
@@ -0,0 +1,100 @@
...
+``repo_labor`` using the “SCC” tool provided the
+https://github.com/boyter/scc project. This worker is presently in
+development and not deployed.
```
ccarterlandis: It's now finished, but requires Go to be installed on a user's system (since `scc` is a Go program`), so it's an optional dependency.

docs/source/schema/data-model.rst
```diff
@@ -0,0 +1,100 @@
...
+
```
ccarterlandis: We also have a new `pull_request` worker, but if you want to just add a list item in I can fill in the technical details myself if you like.

docs/source_backup/architecture/data-model.rst
```diff
@@ -59,15 +59,17 @@ Presently this is focused on issues, including issue_comments,
...
-badging program. 5. ``workers/code_analysis``: Populates the table
+of MySQL and other changes noted in the commit logs.
+3. ``workers/insight_worker``: Generates summarizations from raw data
+gathered from commits, issues, and other info.
+4. ``workers/linux_badge_worker``: Pulls data from the Linux Foundation’s
+badging program.
+5. ``workers/code_analysis``: Populates the table
 ``repo_labor`` using the “SCC” tool provided the
 https://github.com/boyter/scc project. This worker is presently in
 development and not deployed.
```
ccarterlandis: Anything in the `source_backup` folder will not be used as documentation, it is just a convenient storage location for the old docs in case we need to reference something. The folder will be deleted before we officially release

docs/source/getting-started/database.rst
ccarterlandis: I see you already started, which is great!

ccarterlandis:
Looping in @sgoggins on this since we are discussing the data model documentation

docs/source/schema/data-model.rst
Pogayo: Thinking of adding a resource with instructions on how to install go. For those using docker do the images come with Go?

docs/source_backup/architecture/data-model.rst
Pogayo: Alright. I think now that the working version is under schema and I have unlinked it, it is okay now.

PR #635, opened 2020-03-26
Moving the repetitive tests to a common function (DRY)

mrsaicharan1:
To enforce DRY principles, moved the repetitive code to a common function. It also promotes code reusability.


sgoggins:
@ccarterlandis : can you take a quick look at this? I think its a nice optimization of how we write our tests, but you are the Travis master, so I'd like your opinion before merging. 

@mrsaicharan1 : Looks really good to me!

Issue #615, opened 2020-03-19
Top Commiters API problem.


ilam56:
**Describe the bug:**
the "top-commiters" get request in the api doc http://augur.osshealth.io/api_docs/#api-Experimental-top_committers_repo doesn't seem to work for many repos. I think the repo-group version doesn't work right either. For some repos it will just give an internal server error and for others the number of commits don't seem to add up to the right number.

**To Reproduce:**
1. Go to http://czi2.osshealth.io:5153/api/unstable/repos/26732/top-committers
2. Go to http://czi2.osshealth.io:5153/api/unstable/repo-groups/252/top-committers
3. Go to http://czi2.osshealth.io:5153/api/unstable/repo-groups/201/top-committers

**Expected behavior:**
For the first and second link I get internal server errors despite the aggregate summary for those ids showing that they have commits. They shouldn't have resulted in internal errors. The third link is an example where the number of commits don't add up. Bases on the aggregate summary the Astral repo group has 43 commits but the top commiter + others only adds up to 3 commits

**Desktop (please complete the following information):**
 - OS: Windows 10 version 1903(OS Build 18362.720)
 - Browser: Chome Version 80.0.3987.149 (Official Build) (64-bit)

**Additional context**
I am a student of Goggins when I told him about this problem he said to submit an Issue, so here it is.

ccarterlandis:
Hi @ilam56, thanks for bringing this to our attention. I'll get started on a fix.

pratikmishra356:
@ccarterlandis I found that "top-commiters" metric endpoint were not mentioned in metrics/commits/routes.py .Instead it is mentioned in metrics/contributor/routes.py .This might be the cause of error.Can I work on this issue and resolve this??

pratikmishra356:
@ccarterlandis I think there might be some errors in the calculation part , thats why in the third link it is showing partial data.Currently none of the link is working.

pratikmishra356:
@ccarterlandis @ilam56 This issue occurs due to the following reason
 First and second link is showing internal errors because the "sql" queries not giving desired output, it returns None Type.
![Screen Shot 2020-03-29 at 1 09 27 AM](https://user-images.githubusercontent.com/43684300/77832244-9dffbe00-715a-11ea-8cfd-96d6d3c96bfd.png)
[...]

PR #646, opened 2020-03-29
Fixes bug in API top-committers metric ,Added Notebook and Metric

pratikmishra356:
@ccarterlandis @gabe-heim 
As mentioned in #615 Internal server error cause when the given year has no commits so far, this is because it returns None type .
When there is no commits for the given year ,the value of year will be changed to previous year.


augur/metrics/commit/commit.py
```diff
@@ -307,6 +312,11 @@ def top_committers(self, repo_group_id, repo_id=None, year=None, threshold=0.5):
 
         results = pd.read_sql(total_commits_SQL, self.database,
                             params={'year': year, 'repo_id': repo_id})
+        if(results.iloc[0]['sum']==None):
+            year -=1
+            results = pd.read_sql(total_commits_SQL, self.database,
+                                params={'year': year, 'repo_id': repo_id})
+
 
     total_commits = int(results.iloc[0]['sum'])
```
mrsaicharan1: @pratikmishra356 Hi! Umm, I believe that when we're trying to pull commits for a certain year, augur users would expect that it returns the commits for that very year(even if it has zero commits!).
Ergo, I think that the valid fix would be as follows:

 ```
total_commits = 0 if results.iloc[0]['sum'] is None else int(results.iloc[0]['sum'])
```

Comment (mrsaicharan1) [reply augur/metrics/commit/commit.py]: The repercussions of your current fix would lead to another bug namely, **When commits are zero for a current year, it shows commits for the previous year**.
Just a heads up!

Comment (pratikmishra356) [reply augur/metrics/commit/commit.py]: @mrsaicharan1 you are right but then in the later code it returns error
```suggestion
    cumsum = 0
    for i, row in results.iterrows():
        cumsum += row['commits']
        if cumsum >= threshold_commits:
            results = results[:i + 1]
            break

    if not repo_id:
        rg_name = results.iloc[0]['repo_group_name']
        results.loc[i+1] = [repo_group_id, rg_name, 'other_contributors',
                            int(total_commits - cumsum)]
```

```suggestion
    rg_name = results.iloc[0]['repo_group_name']
```

mrsaicharan1: https://github.com/chaoss/augur/issues/615#issuecomment-605511412

This just shows one error. Please put up the error pertaining to the one mentioned above.

pratikmishra356: Screen Shot 2020-03-29 at 12 35 13 PM
it will show errors for every columns in results ,another way would be to set value "0" for every columns in results if there is no data available.

pratikmishra356:
@sgoggins @gabe-heim Please review  Sample Notebook which contains the basic prototype implementation for Time series data mentioned in my gsoc_proposal.I used stacked LSTM and created Pie Chart which shows contribution of every metrics in an anomaly.

pratikmishra356:
@sgoggins @gabe-heim @ccarterlandis Currently we have no metrics for messages and titles , so I thought to create new metrics for messages and titles.Any suggestions would be helpful for me to create those metrics.

mrsaicharan1:
Mentors and many other contributors are located in the US. It's a weekend, therefore your patience would be appreciated :D

pratikmishra356:
@mrsaicharan1 Ohh I see , but we are here to keep work mode on.

pratikmishra356:
@ccarterlandis @gabe-heim Created new metrics for pull requests with messages

PR #701, opened 2020-04-27
Changes in front-end | Implementation of suggestions

NodiraIbrogimova:
For the last weeks, I have been on reviewing and familiarizing myself with the front-end of Augur.

Shortlist of work done in this PR according to my observations in best-practices and **previous projects**:
- In `assets`:
1. Separate words in names of files with “-”
2. Better to name all pictures in lowercase
3. Pictures in avatars should be named according to their purpose
4. All images should be in the images folder
5. Image extensions do not matter, not necessary to cluster into folders for different extensions

- It's easier to find and change when all _public scope_ `images`, `CSS`/`SCSS`, and `fonts` are inside `assets/<extension-type>` folders. 
- **Do not add text unwrapped** in a template, **wrap it into semantical tags**: span, p, etc.
- For conditional rendering (and for some other cases) use built-in tags of `vue`
- As for comments, _it's better_ when each member/participant delete log messages after the usage of their cases. Otherwise, it will **mess** the **code** and its **readability.** It’s **easier** and **cleaner** to read-only code **without extra parts**. 

I will look forward for review and feedback. Thanks.

_Note: I mentioned some suggestion in issues_


NodiraIbrogimova:
I hope everyone is doing good in this not an easy period related to virus. I would like to have a review and feedback for this PR as soon as possible.
Cheers!

NodiraIbrogimova:
Dear @gabe-heim thank you for the reply. Happy to read this feedback. I hope to get back with new updates soon and looking forward for the support of outstanding team!

PR #710, opened 2020-05-17
Spike out automatic standard route generation

ccarterlandis:
This is my first attempt at doing automatic route generation for all `standard` metrics. 

## Overview
Any metric not that does not explicitly define its type when calling `@register_metric()` will be assigned the `standard` type. Upon instantiation, Augur's `Server` will create automatically generate the routes for any `standard` metric and will skip all others. All non-standard metrics will need to hand-implement their own routes under `augur/routes/nonstandard_metrics.py`.

Each standard metric's route will be at `.../<tag>`, where `<tag>` is the name of the metric function transformed to all lowercase and all underscores (`_`) are each replaced by a dash (`-`). For example, the `pull_requests_merge_contributor_new` standard metric will have two endpoints generated:
```
.../repo-groups/<repo_group_id>/pull-request-merge-contributor-new
```
and
```
.../repos/<repo_id>/pull-request-merge-contributor-new
```

I double checked and each metric that already had an endpoint was following this naming schema to begin with, so we should see no breakage, and in fact a few standard metrics that did not have endpoints yet are now available. This greatly simplifies the metrics addition process, as now to add a standard metric, the only code one needs to write the definition of the metric itself.


ccarterlandis:
@sgoggins feel free to test and merge at your leisure

sgoggins:
@ccarterlandis : Tests passed. I verified it on the augur-twitter instance and the augur-spos instance.

PR #726, opened 2020-05-27
Add releases info and metrics

ivanayov:
This adds data and metrics for the number of releases in the repo as well as release details


sgoggins:
@ccarterlandis : I do not see where the `augur.config.json` is made aware of this new worker. I think that's something we can address in a different pull request. @gabe-heim : Can you show @ivanayov, if she does not know, how to add a worker block to `augur.config.json` ... I think its not hard. :)

sgoggins:
Other PR I just merged is unrelated. So merging.

PR #723, opened 2020-05-25
WIP: Testing architecture updates

ccarterlandis:
### **THIS PR IS A WORK IN PROGRESS. DO NOT MERGE.**

For visibility into the current state of our testing architecture update efforts.

PR #729, opened 2020-05-29
Testing architecture updates

ccarterlandis:
For visibility into the current state of our testing architecture update efforts. Duplicate of #723, just from a more aptly-named source branch.


augur/application.py
```diff
@@ -4,92 +4,69 @@
 """
 
 import os
-import time
-import multiprocessing as mp
```
sgoggins, line 8: I am assuming these imports are refactored elsewhere. Facade for sure needs multiprocessing.

augur/application.py
```diff
@@ -105,47 +82,5 @@ def __connect_to_database(self, include_spdx=False):
...
-                value = default_config[section][name]
-        else:
-            try:
-                value = self.config[section]
-            except KeyError as e:
-                value = default_config[section]
-
-        return value
-
-    def load_env_configuration(self):
```
sgoggins, line 130: Looks like a lot of cleanup. :)

augur/cli/run.py
```diff
@@ -4,187 +4,148 @@
...
 @click.option("--disable-housekeeper", is_flag=True, default=False, help="Turns off the housekeeper")
 @click.option("--skip-cleanup", is_flag=True, default=False, help="Disables the old process cleanup that runs before Augur starts")
+@pass_application
 @click.pass_context
-def cli(ctx, disable_housekeeper, skip_cleanup):
+def cli(ctx, augur_app, disable_housekeeper, skip_cleanup):
     """
     Start Augur's backend server
     """
+    reset_logfiles()
```
sgoggins, line 29: Definitely easier to read @ccarterlandis !

augur/server.py
```diff
@@ -10,50 +10,40 @@
...
 from flask_cors import CORS
 import pandas as pd
 
 import augur
-from augur.util import logger
 from augur.routes import create_routes
 
 AUGUR_API_VERSION = 'api/unstable'
 
-class VueCompatibleFlask(Flask):
```
sgoggins, line 24: @ccarterlandis : looks like we are removing a vue dependency from the server.py. duh ... curious how this works, or if it being in there was a legacy.

augur/application.py
ccarterlandis: No, they're not. We just don't need these libraries for anything in the `Application()` class. This is entirely separate from Facade

augur/server.py
ccarterlandis: It's legacy

PR #745, opened 2020-06-02
Adding platform column to Augur tokens table

sgoggins:
For Gitlab worker.


sgoggins:
@gabe-heim : that's in a different schema, so I don't think we should.

PR #759, opened 2020-06-08
Add config for release worker

### LINKED ISSUE #736: Need Model and worker info for releases worker
The new releases worker needs to have sections added for the model and the worker in the augur.config.json in order to run.

ivanayov:
Release worker was previosly missing in the configuration and not enabled

Resolves #736

### DIFF
--- a/augur/cli/configure.py
+++ b/augur/cli/configure.py
@@ -164,6 +164,11 @@ default_config = {
                 "switch": 1,
                 "workers": 1
             },
+            "release_worker": {
+                "port": 51100,
+                "switch": 1,
+                "workers": 1
+            },
         },
         "Facade": {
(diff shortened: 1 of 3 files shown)


augur/cli/configure.py
```diff
@@ -164,6 +164,11 @@
                 "switch": 1,
                 "workers": 1
             },
+            "release_worker": {
+                "port": 52000,
```
ccarterlandis, line 168: Can you set this default port to be `51100`? We try to leave 100 ports open after each worker and we try to keep them close together if we can, it's a bit easier to keep track of that way IMO.

ivanayov: Thanks for noticing @ccarterlandis

PR #771, opened 2020-06-12
Create SECURITY.md

sgoggins:
Adding security file in master back to dev. We need this to be "compliant" and promoted on GitHub.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,16 @@
+# Security Policy
+
+## Supported Versions
+
+These versions of Augur are currently supported with security updates.
+
+| Version | Supported          |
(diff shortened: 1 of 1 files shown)


ccarterlandis:
Ignore that force push. In the future, all code should flow *strictly* from `dev` to `master`, and if for some reason it can't, then we need to make a separate branch and make sure to update them both together. As much as possible, **`master` should never be ahead of `dev`**. Otherwise, there is a possibility someone might get confused and accidentally merge `dev` into `master` because the branch is behind, realize what they did mere seconds after doing it, and then scramble to reset `master` before someone else realizes what happened. Just a hypothetical...🤣

PR #784, opened 2020-06-19
Updating Test Branch

sgoggins:
Latest version of dev for release


workers/gitlab_issues_worker/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
sgoggins, line 1: at a later time, @ccarterlandis , this should be the SPDX license identifier and not the entire MIT License. Also, its 2020.

PR #789, opened 2020-06-20
Merge Request Worker

abhinavbajpai2012:
Adding gitlab-merge-request-worker with the following functionalities -

1. Query and store merge requests
2. Query and store MR labels
3. Query contributors (using worker_base)
4. Query MR events
5. Query MR comments/Notes


sgoggins:
You don’t need the whole MIT license in the file. Just the SPDX identifier. @Nebrethar can point you in the right direction.  

If this PR requires any schema changes you should create a branch based on dev called gitlab-worker. 

I don’t see any schema changes. If there are any we need to get that file into this PR as well. @gabeheim or I can help with that.

workers/gitlab_merge_request_worker/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2019 Augur Team
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
```
sgoggins, line 10: Change this to the SPDX license identifier.

ccarterlandis: We just shouldn’t include this file. I don’t see why we would need the whole license again below the root directory where it‘s already specified, but maybe I’m missing something.

mrsaicharan1: This might've been transferred from another repository. Ergo, the license! I guess we wouldn't require an additional one, but not sure..

abhinavbajpai2012: Cool. I have removed the LICENSE file.

sgoggins:
@gabe-heim / @ccarterlandis : This lgtm. See any worries?

PR #806, opened 2020-06-27
Update installation.rst

maxbalk:
fix broken link for Go download (additional https)


sgoggins:
URL was wrong. Thanks, @maxbalk!

PR #830, opened 2020-07-06
Fixing `make install` , `augur util kill`, updating frontend

sgoggins:
See commit comments.

### DIFF
--- a/augur/cli/util.py
+++ b/augur/cli/util.py
@@ -7,6 +7,7 @@ import os
 import signal
 import logging
 from subprocess import call, run
+import time
 
 import psutil
 import click
@@ -52,9 +53,28 @@ def cli_kill_processes():
     if processes != []:
         for process in processes:
             if process.pid != os.getpid():
-                logger.info(f"Killing process {process.pid}")
+                logger.info(f"Terminating process {process.pid}")
                 try:
                     process.send_signal(signal.SIGTERM)
+                    logger.info(f"sending SIGTERM Signal to {process.pid}")
+                except psutil.NoSuchProcess as e:
+                    pass
+
+            logger.info(f"Waiting to check if processes terminated.")
+
+    time.sleep(15)
+    logger.info(f"Checking on process termination.")
+
+    processes = get_augur_processes()
+
+    if processes != []:
+        for process in processes:
+
+            if process.pid != os.getpid():
+                logger.info(f"Killing process {process.pid}")
+                try:
+                    process.send_signal(signal.SIGKILL)
+                    logger.info(f"sending SIGKILL Signal to {process.pid}")
                 except psutil.NoSuchProcess as e:
                     pass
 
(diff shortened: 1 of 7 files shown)


augur/cli/util.py
```diff
@@ -64,11 +84,29 @@ def kill_processes():
     if processes != []:
         for process in processes:
             if process.pid != os.getpid():
-                logger.info(f"Killing process {process.pid}")
+                logger.info(f"Terminating process {process.pid}")
                 try:
                     process.send_signal(signal.SIGTERM)
+                    logger.info(f"sending SIGTERM Signal to {process.pid}")
```
sgoggins, line 90: I could have saved a bunch of lines of code and just put the SIGKILL here by default, but I think there is merit to allowing some small amount of time for a softer termination first.

scripts/install/frontend.sh
```diff
@@ -24,6 +24,11 @@ read -r -p "Would you like to install Augur's frontend dependencies? [Y/n] " res
 case "$response" in
   [yY][eE][sS]|[yY]) 
     echo "Installing..."
+
```
sgoggins, line 27: I sent @ccarterlandis and @gabe-heim a message in slack about this. 

I fixed the augur install-dev and augur install frontend.sh issue … tl;dr, I hacked it a bit because when the backend finishes by calling augur configure generate, it overwrites the logs directory, removing the install folder. My hack for now is to recreate it at the beginning of install/frontend.sh, which allows that to work, and the install to finish.  We are losing the install/backend.log file when augur configure generate is called, which is not ideal, but its good enough for now. There’s enough happening with the GSoC students that getting a release out so we aren’t tripping over each other as much is more important than perfection.  We should fix my hack in a future release.

augur/cli/util.py
```diff
@@ -52,9 +53,28 @@ def cli_kill_processes():
     if processes != []:
```
sgoggins, line 53: see my note under kill_processes ... I am not exactly sure why we have two nearly identical commands, but I suspect its backward compatibility related. :)

scripts/install/frontend.sh
ccarterlandis: Hmmmm, this didn't even cross my mind, that's definitely what's happening here. Really, the install log files are not very helpful since if the install failed one would want to know right away rather than go check a logfile. They're also immediately discarded the first time Augur is run, so it's not like we even persist them anywhere so they could be useful later. Everything should just be logging to `stdout` but this is fine for now

augur/cli/util.py
ccarterlandis: This needs to be 2 separate commands, one for a graceful exit and one for a not-so-graceful exit. Having the kill command sleep each time when it's likely not even going to be needed in 99% of cases will just cause a lot of confusion. 15 seconds is also a really long time to be waiting in this case.

augur/cli/util.py
ccarterlandis: The `cli_kill_processes` really needs to be changed to just be a wrapper for the actual `kill_processes` function. There are two version because `kill_processes` needs to be imported elsewhere in the code, and with how Augur is set up we can't use the `cli_kill_processes` command to run it has to be repeated. This will be fixed later but this is why there are 2 versions.

frontend/src/components/layout/MainFooter/MainFooter.vue
```diff
@@ -1,34 +1,35 @@
...
-  to: '#',
+  title: 'Augur Documentation',
+  to: 'https://oss-augur.readthedocs.io',
 }, {
-  title: 'Products',
-  to: '#',
-}, {
-  title: 'Blog',
-  to: '#',
+  title: 'Submit Augur Issue',
```
ccarterlandis, line 31: I think `Submit an issue` is a clearer title here.

gabe-heim:
@sgoggins I changed the "Submit Augur Issue" title button to be "Submit an Issue" per @ccarterlandis 's suggestion. If you do not agree it can easily be changed again to something we all agree with. The rest of this PR looks good to me other than the issues that Carter brought up, after those are addressed then I think it is good to be merged.

augur/cli/util.py
sgoggins: OK ... we can do this in the next release.

PR #835, opened 2020-07-08
Updating package-lock.json and package.json to what is current on the…

sgoggins:
… servers.


sgoggins:
@gabe-heim : This is the current state of package-lock.json and package.json on our servers. I figure its a good idea to keep things up to date. It makes the front end work faster actually.

PR #872, opened 2020-08-13
Message Insights worker - Tests & Documentation

aksh555:
tests & documentation for message worker


sgoggins:
It looks like you need to add skimage as a library in the setup.py for your worker @aksh555 ... The python tests are all failing and that looks like the error.

aksh555:
@sgoggins, Yes I have added scikit-image, the library package needed in setup.py. But I don't think it gets installed during the build..

aksh555:
@sgoggins Yes, actually this includes the tests, https://github.com/chaoss/augur/pull/872/files#diff-c003279e53abba2232bdc4c9ab585f59R1. The tests are passing locally

PR #873, opened 2020-08-13
Discourse Analysis Worker

sarit-adh:


sgoggins:
@sarit-adh : It looks like you have not written tests, and that's why its failing in TRAVIS.  Can you write the tests?

PR #895, opened 2020-08-27
Update release worker

ccarterlandis:
Now that `dev` has been properly reset and the build is fixed, let's get ready to merge these into `dev`. @gabe-heim these changes LGTM but it would make me feel better if you reviewed them as well.


sgoggins:
@ccarterlandis : Is there a reason we are not merging this?

ccarterlandis:
Like I mentioned above, I would like @gabe-heim to review it first.

PR #904, opened 2020-09-02
Fix Augur process control commands

ccarterlandis:
At somepoint in the summer, a patch was introduced that introduced a 15 second wait time to the `augur util kill` command that occured after the first attempt to gracefully stop the process with `SIGTERM`, after which it would attempt to kill all the process with `SIGKILL`. While this fallback makes sense, it in inadvertently introduced this artifical 15 second wait time to the startup of the main server. This was due to poor design during the initial implementation, which copied code blocks that should have been refactored. The 15 second delay was added to both of these code blocks, one of which was used in `augur run`, thus causing the slowdown.

This was fixed by refactoring out the copied code blocks to a singular underlying method call, removing the 15 second wait altogether, and exposing a new command as well changing the meaning of an existing one.

The `augur util stop` command was added, which will attempt to gracefully terminate any existing Augur process. This was how `augur util kill` previously behaved; `augur util kill` will now send a `SIGKILL` signal instead of a `SIGTERM`, in order to forcefully halt the execution of the process. This should only be used when Augur has crashed hard, and is not responding to attempts to use `augur util stop`.

There is further refactoring to be done, but this is better than where it was before!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ default:
 #
 #  Installation
 #
-.PHONY: install install-dev 
-.PHONY: install-spdx install-spdx-sudo install-augur-sbom 
+.PHONY: install install-dev
+.PHONY: install-spdx install-spdx-sudo install-augur-sbom
 .PHONY: clean rebuild
 install:
 	@ ./scripts/install/install.sh prod
@@ -63,12 +63,12 @@ rebuild-dev:
 #
 .PHONY: dev-start dev-stop dev monitor-frontend monitor-backend monitor frontend backend-stop backend-start backend-restart backend clean rebuild
 
-dev-start: 
+dev-start:
 	@ scripts/control/start_augur.sh
 	@ scripts/control/start_frontend.sh
 
-dev-stop: 
-	@ augur util kill
+dev-stop:
+	@ augur util stop
 	@ scripts/control/kill_frontend.sh
 
 dev: dev-stop dev-start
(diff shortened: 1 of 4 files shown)


sgoggins:
@ccarterlandis : I did that.  As long as we are using SIGKILL instead of SIGTERM, the change is fine.  I assumed you switched from SIGKILL for a reason, so I let it try that first, wait a short time, then do SIGKILL … SIGTERM was not and does not kill all of the augur processes for some reason.

sgoggins:
@ccarterlandis : SIGTERM was CONSISTENTLY not working on ALL OF THE SERVERS … we either need the pause OR we need to use SIGKILL ... basically, until I made the change, `augur util kill` did NOT WORK on any of the servers.

ccarterlandis:
We were never using `SIGKILL`, only `SIGTERM`. This is the first time I've heard of this version not working on the servers, but either way, neither adding a wait nor using `SIGKILL` is the right choice. For what it's worth, I've never run into any problems when I use `augur util kill` on the servers, it always works just fine for me. But, if it was actually failing on the servers, then there is a deeper reason that needs to be addressed, because it works fine elsewhere. Additionally, is it JUST the **servers**, or any **machine running the same Linux distro version**? If so, that is a BIG difference that has a lot of implications.

sgoggins:
@ccarterlandis : OK. I get it now. As long as there's a way to trigger SIGKILL, which there is, I'm good with this. 
First: `augur util stop` 
Second, after some reasonable period, if there are still augur processes running for the instance `augur util kill`

THANK YOU!!!

sgoggins (approved):

PR #905, opened 2020-09-04
Updates lost fixing GSOC-Dev and Necessary default config changes

sgoggins:


.travis.yml
```diff
@@ -41,8 +41,7 @@ env:
 before_install:
   - docker run -d -p 5432:5432 --name augur_test_database augurlabs/augur:test_data@sha256:fd2d9a178a9fee7cd548bd40a16e08d4611be22892491e817aafd53502f74cd0
 install:
-  - ./scripts/install/backend.sh
-  - ./scripts/install/workers.sh
+  - pip install .[dev]
```
ccarterlandis, line 44: You have an old version of this file in your branch. This change (getting rid of the shell scripts again) reverts a bug fix I made. If merged, this change will break the worker builds again. This change needs to be removed.

docs/source/development-guide/toc.rst
```diff
@@ -17,6 +17,7 @@ If you have questions or would like to help please open an issue on GitHub_.
    make/toc
    logging
    documentation
+   workers/toc
```
ccarterlandis, line 20: There is no docs for the workers, it does not make sense to include this TOC entry. This might be an old change too, but it needs to be removed.

setup.py
```diff
@@ -37,7 +37,7 @@
         "sqlalchemy",
         "flask_login",
         "flask",
-        "pandas==1.0.5",
+        "pandas",
```
ccarterlandis, line 40: This "change" also reverts a date precision bug fix that causes 3.6 unit tests to fail due to altered `pandas` behavior since its most recent update. This change needs to be removed.

ccarterlandis (changes requested):
There are several changes that need to be removed, the history of this branch has been messed up somehow and you do not have the most up to date versions of some of these files. Please address the requested changes.

sgoggins:
Oops.

PR #906, opened 2020-09-04
Updating sean-dev from dev

sgoggins:


sgoggins:
@ccarterlandis : I think the 3.6 python tests are failing because of that date precision issue we have discussed.

ccarterlandis:
@sgoggins, yes, they are, but that's because you have the wrong version of `setup.py` in your branch. See my review comments on https://github.com/chaoss/augur/pull/905#pullrequestreview-483007348 for details

PR #907, opened 2020-09-05
update to config.py / add $AUGUR_HOME/scripts/install/git.sh

sgoggins:


sgoggins:
@ccarterlandis I have not called `git.sh` from anywhere yet, but it should be called during installation.  Large repos need the first two lines. And everyone needs the last two for when repos move or get deleted. For every platform.

sgoggins:
@ccarterlandis : This contains an urgent fix we need to get into Master. See related issue #912

scripts/install/backend.sh
```diff
@@ -8,7 +8,7 @@ echo
 target=$1
 
 if [[ $target == *"prod"* ]]; then
-    pip install .
+    pip install -e .
```
ccarterlandis, line 11: Please undo this change. We should not be installing for production with the `-e` flag.

scripts/install/frontend.sh
```diff
@@ -10,6 +10,7 @@ function install_deps() {
     echo
     npm install;
     npm install brunch canvas vega @vue/cli;
+    npm run build;
```
ccarterlandis, line 13: Please undo this change. Unless we are deploying for production, there is no reason we need to build the frontend statically; given that installing for dev vs. prod does not exist right now, it doesn't make sense to do this yet.

scripts/install/git.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+git config --global diff.renames true
```
ccarterlandis, line 2: Why are these global settings? We do not need to be overwriting a **global** Git config without **VERY** good reason, and we need to tell them about it loud and clear. Can these be changed to a `user`, `local`, or even `system` scope? And this script needs to print out info about what's being set at what level.

ccarterlandis (changes requested):
I can't recreate the issue in #912. Also, I don't see how anything in this PR would fix that issue, none of the files changed are related to the config at all.

There are other changes in this PR that need to be addressed first. See comments below

scripts/install/backend.sh
sgoggins: @ccarterlandis : I will see if taking that change out results in the `augur.config.json` not getting written out with `make install` ... I was able to replicate the problem on the current state of `master` using python 3.8.5 on mac OSX catalina.

scripts/install/frontend.sh
sgoggins: So, I do `install-dev` on most of our instances because we use them for debugging.  I can do `npm run build` myself in those cases. I see your point.

scripts/install/git.sh
sgoggins: pretty simple: if a repo moves and credentials are not cached, all platforms prompt the facade worker for a login, basically stopping the process.

sgoggins: @ccarterlandis : Large repos need large renames numbers.

scripts/install/backend.sh
ccarterlandis: Then this is a red herring, if this "fixes" that issue then there is something else going horribly wrong with the installation that is the real issue. Adding the flag to install the package in editable mode is not and should not be required for production, only for development.

scripts/install/git.sh
ccarterlandis: I understand why we need the settings, but I still don't understand why these need to be **global** settings. If they HAVE to be global, then this is not going in an automated script. Changing global Git settings with out a user's consent or knowledge is a really quick way to piss them off :)

sgoggins:
@ccarterlandis : I did revert my changes to the files in question. 
Except the `git` one. Those settings are "global" for the user, and I cannot imagine where they would cause harm. But I might not have a good imagination.

scripts/install/git.sh
sgoggins: Well, if they are not "global" they have to be set on every repository... or am I being `git dumb`?

sgoggins:
@ccarterlandis : I don't automatically call the git.sh file anywhere. I think we would document what its for, or add it as a choice, explaining why these settings are needed, in the installation script at some point.

sgoggins (commented):
@ccarterlandis : For now, the config.py changes and existence of git.sh as a utility are helpful and important for "the way things work today". If the script `git.sh` is *there*, its also a helpful reminder to make a note of what its about in the documentation. Its not an ideal solution, but I think its one that many augur users will appreciate.

ccarterlandis:
This is definitely not an ideal solution, as the Facade worker should have more intelligent handling than to just error out when this sort of thing happens. We can work these as in *temporary* changes for now, but I'm not willing to approve these changes until we are absolutely 100% certain that the `--global` flag is required, and that `--system` doesn't suffice. We need to avoid changing global Git settings unless there really is no workaround, and if we do have to use `--global`, then a `Yes/No` prompt and some informational output and **MUST** be added to the script to cover our tracks.

ccarterlandis (changes requested):
See above comment for changes

sgoggins:
@ccarterlandis There's a lot in here now that we clearly don't want. I am going to make another pr for the `git.sh` file because on large scale implementations it is presently necessary.

PR #916, opened 2020-09-11
Update default config file location

ccarterlandis:
Currently, on all branches, using `augur configure generate` (the augur lib command to create a config file) behaves in an undesirable way if installed using `make install`. The issue is that after running `augur configure generate`, if you used `make install` to install Augur, your config file is not written to root of your cloned source code directory. It is instead written to the `site-packages/augur/` folder that lives with your Python binary inside your virtualenv folder. The reason that it writes to this location is due to how I implemented the config generation script. The script uses a package-level variable called `ROOT_AUGUR_DIRECTORY` to determine the location of the root of the file currently that is currently running; when installed in non-editable mode (without using the `-e` flag, like in `make install`), `pip` will copy ONLY the files defined in your `MANIFEST.in` from your cloned source code directory (if that's what you're installing) into that `site-packages` folder; **these are the files that are "running" when using augur installed in non-editable mode** - that is, without `-e`.

However, when installing the package in editable mode with `-e` (like `pip install -e .` in `make install-dev`), the files are **not** copied into that folder. Instead, some stuff happens behind the scenes, and `pip` links the contents of your local cloned directory to the `site-packages` folder; this is what allows you to make changes, save, and then run the commands again. That means when we use `install-dev`, **these are the files that are running when we execute augur commands. This is why the augur.config.json file is always installed to your cloned source code directory; that's the "root" of the currently running Augur binary.

There's a very important caveat to using `make install`, however; _it has no way of knowing where you installed the binary from_. This means it has _no way of installing the config file to your cloned source code directory_. In fact, it's not even guranteed that the source is on your machine; all the other packages you install with `pip` like `pandas` only have the files in their `MANIFEST.in` copied as well. If we want the config file to only ever be in one location, then this is problem; it can't be in the source code directory like it has been. We need to pick a standard location on the file system to store the config file outside the cloned source code tree.

To this end, this patch changes the default config file location `$HOME/.augur/augur.config.json` for all production installations. However, if it detects a config file in the root Augur directory, then it will use that instead, for backwards compatibility.

### DIFF
--- a/augur/cli/configure.py
+++ b/augur/cli/configure.py
@@ -7,14 +7,14 @@ import os
 import click
 import json
 import logging
+from pathlib import Path
 
-from augur.config import default_config, ENVVAR_PREFIX
+from augur.config import default_config, ENVVAR_PREFIX, CONFIG_HOME
 from augur.cli import initialize_logging
 from augur.logging import ROOT_AUGUR_DIRECTORY
 
 logger = logging.getLogger(__name__)
 ENVVAR_PREFIX = "AUGUR_"
-
 
 @click.group('configure', short_help='Generate an augur.config.json')
 def cli():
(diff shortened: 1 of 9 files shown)


gabe-heim (approved):
lgtm

PR #918, opened 2020-09-11
Update installation logging for clarity

ccarterlandis:
## Changelog

This patch:

- updates the installation process to write the all logs (including errors) to both `stdout` and to the `logs/` directory
- hides user input when entering their database password during the installation process
- adds a warning in the documentation to not use `sudo` when running or installing Augur

PR #920, opened 2020-09-13
Revert "Update installation logging for clarity"

ccarterlandis:
Reverts chaoss/augur#918 which was merged early without the appropriate review.


sgoggins:
@ccarterlandis ... ok ... it was out of "draft" status, and it looked good to me. We are requiring 1 review for merging in the rules. Is the practice different?

ccarterlandis:
I explicitly requested @Nebrethar's review, which he did not have a chance to provide. We do "require" only 1 review, but really, I should have removed you as a reviewer so that only @Nebrethar's review would allow the PR to pass - that's my fault.

ccarterlandis:
Opening a new PR for Matt to review

PR #923, opened 2020-09-21
Add ability to create repo groupgs from GH orgs on the CLI

ccarterlandis:
Like the title says. Takes a GitHub org name, and (blindly, without dupe checking) creates a repo group with the same name. It then inserts all the repos from GitHub into that repo group.

### DIFF
--- a/augur/cli/db.py
+++ b/augur/cli/db.py
@@ -96,16 +96,15 @@ def add_github_org(augur_app, organization_name):
         logger.fatal(f"No organization with name {organization_name} could be found")
         exit(1)
 
-    page = 1
-    headers = {'Authorization': 'token %s' % augur_app.config.get_value("Database", "key")}
     all_repos = []
-    repo_query_response = requests.get(org_query_response['repos_url'] + f"?per_page=100&page={page}", headers=headers).json()
+    page = 1
+    repo_query_response = None
+    headers = {'Authorization': 'token %s' % augur_app.config.get_value("Database", "key")}
     while repo_query_response != []:
+        repo_query_response = requests.get(org_query_response['repos_url'] + f"?per_page=100&page={page}", headers=headers).json()
         for repo in repo_query_response:
             all_repos.append(repo)
-
         page+=1
-        repo_query_response = requests.get(org_query_response['repos_url'] + f"?per_page=100&page={page}", headers=headers).json()
 
     insert_repo_group_sql = s.sql.text("""
     INSERT INTO "augur_data"."repo_groups"("rg_name", "rg_description", "rg_website", "rg_recache", "rg_last_modified", "rg_type", "tool_source", "tool_version", "data_source", "data_collection_date") VALUES (:repo_group_name, '', '', 0, CURRENT_TIMESTAMP, 'Unknown', 'Loaded by user', '1.0', 'Git', CURRENT_TIMESTAMP) RETURNING repo_group_id;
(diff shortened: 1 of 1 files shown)


augur/cli/db.py
```diff
@@ -77,10 +78,51 @@ def add_repo_groups(augur_app, filename):
             logger.info(f"Inserting repo group with name {row[1]} and ID {row[0]}...")
             if int(row[0]) not in repo_group_IDs:
                 repo_group_IDs.append(int(row[0]))
-                augur_app.database.execute(insert_repo_group_sql, repo_group_id=int(row[0]), repo_group_name=row[1])
+                augur_app.database.extendcute(insert_repo_group_sql, repo_group_id=int(row[0]), repo_group_name=row[1])
```
gabe-heim, line 81: extendcute does not exist

augur/cli/db.py
```diff
@@ -81,6 +82,47 @@ def add_repo_groups(augur_app, filename):
...
+    page = 1
+    headers = {'Authorization': 'token %s' % augur_app.config.get_value("Database", "key")}
+    all_repos = []
+    repo_query_response = requests.get(org_query_response['repos_url'] + f"?per_page=100&page={page}", headers=headers).json()
+    while repo_query_response != []:
+        for repo in repo_query_response:
+            all_repos.append(repo)
+
+        page+=1
+        repo_query_response = requests.get(org_query_response['repos_url'] + f"?per_page=100&page={page}", headers=headers).json()
```
gabe-heim, line 108: page incrementation should be after you hit the endpoint

gabe-heim (changes requested):
see comment

augur/cli/db.py
```diff
@@ -81,6 +82,47 @@ def add_repo_groups(augur_app, filename):
...
+    insert_repo_sql = s.sql.text("""
+        INSERT INTO augur_data.repo(repo_group_id, repo_git, repo_status,
+        tool_source, tool_version, data_source, data_collection_date)
+        VALUES (:repo_group_id, :repo_git, 'New', 'CLI', 1.0, 'Git', CURRENT_TIMESTAMP)
+    """)
+    logger.info(f"{organization_name} repo group created")
+
+    for repo in all_repos:
+        logger.info(f"Adding {organization_name}/{repo['name']} ({repo['clone_url']})")
+        result = augur_app.database.execute(insert_repo_sql, repo_group_id=new_repo_group_id, repo_git=repo['clone_url'])
```
gabe-heim, line 124: There is a way to bulk insert which is faster, but if performance is not an issue (which I don't imagine it would be because I wouldn't think there would be an insane amount of repos in any org) then this implementation is just fine. So lmk if you want me to point you to some resources

augur/cli/db.py
ccarterlandis: It increments it after because it hits it once first and then needs to get the second request inside the loop, but looking back at it now that was a really wacky way to implement the loop and I'm gonna refactor it.

augur/cli/db.py
ccarterlandis: I considered a bulk insert but since there won't be too many repos for the vast majority of cases I went with this

ccarterlandis:
@gabe-heim made changes to that request loop so it makes more sense, thoughts?

ccarterlandis:
@gabe-heim pinging 🛠️

gabe-heim (approved):

PR #933, opened 2020-09-29
Dev release schema patch

sgoggins:
Adds the tags to the schema creation, which was missed during testing because a new contributor didn't know to include that in the patch.


ccarterlandis (approved):
LGTM

PR #935, opened 2020-10-01
git.sh script is needed for large collections (500+ repos) to accomod…

sgoggins:
git.sh script is needed for large collections (500+ repos) to accomodate large scale refactoring and repos that are removed, due to platform behavior when we attempt to clone a repo that no longer exists. Long term, error handling in Facade, and addressing 2 open issues related to repository org changes, repository additions in orgs, and repository deletion will fix this issue

### DIFF
--- a/scripts/install/git.sh
+++ b/scripts/install/git.sh
@@ -1,4 +1,3 @@
 #!/bin/bash
 git config --global diff.renameLimit 200000
-git config --global credential.helper cache
-git config --global credential.helper 'cache --timeout=9999999999999'
+git config --global credential.helper store
--- a/scripts/install/git.sh
+++ b/scripts/install/git.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+git config --global diff.renameLimit 200000
+git config --global credential.helper cache
+git config --global credential.helper 'cache --timeout=9999999999999'


scripts/install/git.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+git config --global diff.renameLimit 200000
```
ccarterlandis, line 2: I am still convinced that these need to be using `--global`. Can you please test if these work with the `--system` flag, or some other less far-reaching location setting?

scripts/install/git.sh
```diff
@@ -0,0 +1,4 @@
...
+git config --global credential.helper cache
+git config --global credential.helper 'cache --timeout=9999999999999'
```
ccarterlandis, line 4: We should not be using the `cache` helper with a large timeout. There is a [`store` credential helper](https://git-scm.com/docs/git-credential-store) for this exact use case, please use that one instead.

ccarterlandis (changes requested):
See in-line comments for requested changes

scripts/install/git.sh
sgoggins: @ccarterlandis : I tested this. The --system flag requires sudo privileges on a server. It is more common that a particular user will be running augur, and so in the case of our software --global is a more appropriate solution because we want to avoid requiring sudo wherever possible.

scripts/install/git.sh
sgoggins: I made this change @ccarterlandis.

ccarterlandis (approved):
Let the record show that I am merging this PR begrudgingly, and on the condition that we **actually fix the root cause of these errors in Facade**. I think we should have actually solved these issues rather than use this stopgap, but for the time being (and ONLY the time being) this will suffice.

PR #938, opened 2020-10-01
SPDX Schema Update

sgoggins:


ccarterlandis (commented):
Why is this getting merged into `dev`? I thought all the SPDX stuff was staying out of the `dev` for the time being.

ccarterlandis:
@sgoggins I can't review this until you answer my previous question.

sgoggins:
@ccarterlandis : We are maintaining central control of the schema within Augur. Which is important for the API and Frontend in Augur. 

You an think of the augur-spdx project as a package that populates this schema. Its optional because of the operating system constraints. Ultimately I would love to have it deployed in another Docker container with Augur on any OS, but for now this is "good", and I don't want to go for perfect. Because the perfect is the enemy of the good. ... and other platitudes.

ccarterlandis:
I was under the impression that we still weren't shipping SPDX with Augur in any form, I thought there was still work to be done related to getting that ready. But yeah, I guess this is fine.

ccarterlandis (approved):

PR #953, opened 2020-10-12
Implemented db conn retry

### LINKED ISSUE #945: Add database connection retry logic to Augur server
Right now, the Augur server only attempts to make a connection to the database once, and if it fails, it immediately stops running and exits. Normally, this is fine, but sometimes when using Docker, the server container might become available before the database container is ready. To prevent this from happening more frequently, it would be good to have some retry logic in place (maybe 3 retries, first waiting 1s, then 3s, or 5s) for making this database connection. The exact number of retries or the timeout is up for debate

chinvib66:
Addresses #945 

- Implemented tenacity module for retrying database connection
- We have 5 retry attempts, one every 2 seconds. Refer [tenacity documentation](https://github.com/jd/tenacity) for different possible options.


sgoggins (approved):
@ccarterlandis and @gabe-heim : This implements some retry logic in the main Augur application, which could be useful under high load conditions where the database connection is interrupted or slow. The package introduced looks like a mature, maintained pypi package. 

Thoughts? 

https://pypi.org/project/tenacity/

gabe-heim:
Looking in depth at the tenacity package before I make a review, will readdress the pr shortly.

sgoggins (approved):
We are reticent to include this because of the new dependency. Do we need the tenacity library? It may be cleaner to simply do a retry in the native Python.

chinvib66:
Ok, then I guess it will be better if I create another pull request with the retry implementation, right @sgoggins ?

sgoggins:
@ccarterlandis / @gabe-heim : I think this solves a problem and the dependency is well maintained.

ccarterlandis:
@chinvib66 is there a specific reason you chose to use `tenacity`? We are just trying to get a better understanding of your thought process before deciding if we would want to use it or not.

chinvib66:
My main reason was the ease of integration and flexibility `tenacity` offers. If you refer to the documentation, we have multiple configurable options for retries, delays, and their conditions. Since it is actively maintained till date, I felt it will be safe to use. I have also seen it being used in few decent sized private projects I had worked with in the past.

sgoggins:
Refactoring beyond this code, and into workers, would add limited benefit.

PR #977, opened 2020-10-21
Add logs error parsing

ivanayov:
This change allows to run augur with `augur run --logstash` and observe
errors from workers in http://localhost:8003.


sgoggins:
@ccarterlandis : I know you've replaced `augur run` with `augur backend start`. This log monitor is important for people running augur in production environments. Can you advise on anything necessary to get it merged as soon as possible?

.gitignore
```diff
@@ -152,3 +152,4 @@ workers/spdx_worker/dosocs2.conf
 workers/spdx_worker/augur-spdx/
 workers/spdx_worker/ex-raw.txt
 workers/spdx_worker/ex.json
+backup_test
```
ccarterlandis, line 155: Was this intentional? I'm not familiar with the SPDX worker, but I'm also not sure how this fits into the rest of the PR.

augur/cli/run.py
```diff
@@ -32,13 +36,29 @@ def cli(disable_housekeeper, skip_cleanup):
     else:
         logger.debug("Skipping process cleanup")
 
+    if logstash or logstash_with_cleanup:
+        augur_home = os.getenv('AUGUR_HOME', "")
```
ccarterlandis, line 40: There is a package-wide variable called `ROOT_AUGUR_DIRECTORY` that the rest of Augur uses to determine where it's currently located. You can import it with `from augur import ROOT_AUGUR_DIRECTORY`; can you use that instead of this environment variable, since the package variable already exists?

augur/cli/run.py
```diff
@@ -32,13 +36,29 @@ def cli(disable_housekeeper, skip_cleanup):
...
+        if logstash_with_cleanup:
+            print("Cleaning old workers errors...")
+            with open(augur_home + "/log_analysis/http/empty_index.html") as f:
+                lines = f.readlines()
+            with open(augur_home + "/log_analysis/http/index.html", "w") as f1:
+                f1.writelines(lines)
+            print("All previous workers errors got deleted.")
+
+        elasticsearch_path = os.getenv('ELASTIC_SEARCH_PATH', "")
```
ccarterlandis, line 49: Are there any sensible defaults for this environment variable? Like running on localhost, for example? (Same for the `LOGSTASH_PATH`)

augur/cli/run.py
```diff
@@ -6,6 +6,8 @@
 from copy import deepcopy
```
ccarterlandis, line 6: The `augur run` command got moved to `augur backend start`, in a new `augur/cli/backend.py` file. The structure of the new file is almost identical to the old one; I think all you'll need to do is copy the relevant code over to the `start()` method defined in `backend.py`.

docs/source/getting-started/command-line-interface/run.rst
```diff
@@ -63,3 +63,21 @@ To start the backend server and skip the process cleanup::
```
ccarterlandis, line 63: This file has been renamed to `backend.rst` to update the new command name; just c/p'ing the relevant docs over to the new file should be fine.

docs/source/getting-started/command-line-interface/run.rst
```diff
@@ -63,3 +63,21 @@ To start the backend server and skip the process cleanup::
 
   # successful output looks the same as without the flag
 
+To enable log parsing for errors, set ``ELASTIC_SEARCH_PATH`` and ``LOGSTASH_PATH`` variables to point to elasticsearch and logstash binaries. For example:
```
ccarterlandis, line 66: Can you add a section to the optional dependencies section of the installation docs (`docs/source/getting-started/installation.rst`) about Elasticsearch and Logstash, including links to their respective installation guides? You don't need to go in depth, just the links + maybe a sentence or two explaining what they're used for.

docs/source/getting-started/command-line-interface/run.rst
```diff
@@ -63,3 +63,21 @@ To start the backend server and skip the process cleanup::
...
+Start the http server with:
+
+  $ cd $AUGUR_HOME/log_analysis/http
+  $ python http_server.py
+
+Then run augur with ``logstash`` flag:
+  $ augur run --logstash
+
+If you'd like to clean all previously collected errors, run:
+  $ augur run --logstash-with-cleanup
```
ccarterlandis, line 81: Please wrap all code examples in a `.. code-block:: bash` directive so it's formatted properly. You could also shortcut this by using 2 colons at the end of the line, like:
```
Then run augur with ``logstash`` flag::
  $ augur run --logstash
```
This will format the lines below as a code block. There are lots of other examples of both styles throughout the docs if you need a reference.

log_analysis/http/empty_index.html
```diff
@@ -0,0 +1,85 @@
...
+   var i;
+   for (i = 0; i < ids.length; i ++) {
+       var elements = document.getElementsByClassName(ids[i]);
+       var j;
+       for (j = 0; j < elements.length; j++) {
+           elements[j].style.visibility="hidden";
+       }
+   }
+}
+function myFunction() {
```
ccarterlandis, line 45: Can you give this a more descriptive function name?

log_analysis/http/index.html
```diff
@@ -0,0 +1,85 @@
...
+<select id="jobs_list" onchange="myFunction()" class='form-control'>
+<option value="none">-- Select --</option>
+<option value="github_worker">github_worker</option>
+<option value="facade_worker">facade_worker</option>
+<option value="linux_badge_worker">linux_badge_worker</option>
+<option value="pull_request_worker">pull_request_worker</option>
+<option value="release_worker">release_worker</option>
+<option value="repo_info_worker">repo_info_worker</option>
+</select>
+</html>
```
ccarterlandis, line 79: This code seems functionally very similar to what's in `empty_index.html`. Is there any way to consolidate the logic so there is no code unnecessarily repeated?

ccarterlandis (changes requested):
@ivanayov @sgoggins see my comments inline. Overall, it looks good, just had a few minor asks + some guidance on how to move the functionality to the new `augur backend start` command.

sgoggins:
@ivanayov : if you merge `dev` into your branch, you will see that `augur backend start` replaces `augur run`; `augur backend stop` replaces `augur util stop`, and `augur backend kill` replaces `augur util kill`.  That should fix the merge conflict.

.gitignore
ivanayov: It's by mistake.

augur/cli/run.py
ivanayov: @ccarterlandis this is the path to the local elasticsearch/logstash bin. I left it as a variable so that one will not need to stick to specific convention. If you prefer I can document how to install and set a default.

log_analysis/http/empty_index.html
ivanayov: It's funny that I forgot it like that 😄

log_analysis/http/index.html
ivanayov: @ccarterlandis this page is being populated while running logstash and `empty_index.html` is the "backup", i.e. the empty file that contains no errors data. I found it better to do a cleanup this way, rather than delete specific lines from `index.html`, as the latter may result in unexpected issues (f.e. if someone changes the script, or adds new workers and the line numbers are shifted forward, they will need to update the code, so that the cleanup doesn't delete lines accidentally. 
May be it's better to generate `index.html` during execution and add it to `.gitignore`.

augur/cli/run.py
ccarterlandis: I think documenting a default would be fine!!

log_analysis/http/index.html
ccarterlandis: Ah okay, I see what you were going for. Since these are pretty small pages, and this is an MVP, I think that your solution is fine. We can additional functionality as we need it, and re-evaluate if we want to change how this is implemented then.

sgoggins:
@ivanayov : We are moving toward a release in the next day or so. Would love to have this updated and used as part of the release!

ivanayov:
@sgoggins Great, I'm taking it with first priority then and updating today.

sgoggins:
@ivanayov : Checking status.

PR #1037, opened 2020-12-02
Ml sentiment patch

sgoggins:


sgoggins:
@gabe-heim : The LSTM insight model removed from the insight worker ... did we keep that somewhere? Is it in another worker?

PR #1039, opened 2020-12-03
Automatically add repos to org

### LINKED ISSUE #914: Automatically add new repos to Augur when a repo is added to a GitHub org that is already in Augur
We frequently add new repositories for new open source projects within existing GitHub orgs that are already tracked in Augur. It would be great if these new repositories were automatically discovered and added to the appropriate repo group within Augur. 

The problem is that I don't always know when we create new repos within our existing GitHub orgs, so these repos don't have any metrics unless I happen to notice it and get the repo added manually to Augur. 

For example, the vmware GitHub org (https://github.com/vmware) is already in the pivotal.osshealth.io Augur instance, but you can see that this recently added repository is missing from the org: https://github.com/vmware/go-ipfix

ivanayov:
Automatically add repos to org
  
This change allows automatic updates of newly added repositories.
The housekeeper does those updates, but one can run them manually as well
with `augur db update-org-repositories <org>`

In order to run housekeeper job, ad the following config 
to `augur.config.json`:

```
            {
                "delay": 1000000,
                "given": [
                    "github_url"
                ],
                "model": "update_org_repositories",
                "repo_group_id": <your_repo_group_id>
            }
```

Resolves #914

### DIFF
--- a/augur/cli/db.py
+++ b/augur/cli/db.py
@@ -1,4 +1,5 @@
 #SPDX-License-Identifier: MIT
+from conftest import augur_app
 from os import walk, chdir, environ, chmod, path
 import os
 import logging
@@ -13,9 +14,11 @@ import click
 import sqlalchemy as s
 import pandas as pd
 import requests
+import json
 from sqlalchemy import exc
 
 from augur.cli import pass_config, pass_application
+from augur.platform_connector import PlatformConnector
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


ccarterlandis:
@ivanayov: first, I would like to apologize for taking so long to get to this pull request! Super excited about this functionality, it will be really useful for automating Augur and doing large-scale data collection, so thanks for submitting the patch! 🚀 

However, after talking about this PR with @sgoggins and @gabe-heim, we want to know a little more about the implementation and reasoning behind the `PlatformConnector` class. From what we could gather, it seemed like you wanted to use both the GitHub rate limit update functions and the DB connection functions. However, there are already methods in place for getting a database connection in a command line, and the housekeeper also already has access to database connections too, so using the worker methods to create new connections when there are already ones open is something we want to avoid. Was there a specific reason you didn't use those connections - perhaps you didn't know there were there? I can show you exactly how to use them if you like, but this comment is long enough already so I won't include it here 😉 

Additionally, while we do agree that the GitHub and GitLab API helper methods definitely need to be revisited and cleaned up, we would rather that change it's own patch, as it will likely be a significant change to the workers and to how the methods are used. At the moment we're just not comfortable with moving all of those important methods (as well as the DB connection functions) out of the worker base without more planning and design work. This is something we are also actively working on, and if you would like to help I'd be more than happy to go into greater detail!

I hope my comment made sense, and once again this functionality is really awesome. I just want to make sure we get the implementation as good as we can on the first try 😄

augur/cli/db.py
```diff
@@ -16,6 +16,7 @@
 from sqlalchemy import exc
 
 from augur.cli import pass_config, pass_application
+from augur.platform_connector import PlatformConnector
```
sgoggins, line 19: This adds features for collecting new repos added to an organization.  It is not a duplicate of existing features or functions.

augur/cli/db.py
```diff
@@ -331,3 +332,188 @@ def check_pgpass_credentials(config):
         else:
             logger.info("Credentials found in $HOME/.pgpass")
 
+@cli.command('update-org-repositories')
```
sgoggins, line 335: The reason to have this in the CLI is for sophisticated users to update their repositories when a large number of them have changed, without having to run all of augur. 

It does *not* duplicate existing functions.

augur/cli/db.py
```diff
@@ -331,3 +332,188 @@ def check_pgpass_credentials(config):
...
+
+    repos = get_all_repos(org)
+    repo_group_id = get_org_id(augur_app, org)
+    existing_repos = get_existing_repos(augur_app, repo_group_id)
+
+    for repo in repos:
+        if repo['node']['name'] not in existing_repos:
+            insert_repo(augur_app, repo['node'], org, repo_group_id)
+
+def generate_repo_id(augur_app):
```
sgoggins, line 351: There is a sequence in the database that automatically increments the repoID, and this is the preferred way to do incrementing of repo_id. 

This is a **very** weird SQL way for a relational database to work.  And Postgres does a poor job of making this clear.

augur/cli/db.py
```diff
@@ -331,3 +332,188 @@ def check_pgpass_credentials(config):
...
+    repos = augur_app.database.execute(repos_sql).fetchall()
+
+    if len(repos) == 0:
+        logger.info("Did not find any repositories stored in augur_database for repo_group_id {}\n".format(repo_group_id))
+
+    repos_list = list(map(get_value_from_fetch_data, repos))
+
+    return repos_list
+
+def get_org_id(augur_app, org):
```
sgoggins, line 389: This exists so you can get the organization ID for which organization the repo belongs to. We don't make it easy to get this in any other way.

sgoggins: (i.e., I think it would be confusing, and introduce breaking cases for people who used this feature otherwise.  For example, repo name alone does not need to be unique.

augur/cli/db.py
```diff
@@ -331,3 +332,188 @@ def check_pgpass_credentials(config):
...
+
+        l = len(data['data']['organization']['repositories']['edges'])
+        last_el = data['data']['organization']['repositories']['edges'][l-1]
+        after = ", after: \"" + last_el['cursor'] + "\""
+
+    data['data']['organization']['repositories']['edges'] = nodes
+
+    return data['data']['organization']['repositories']['edges']
+
+def get_repos_total_count(org):
```
sgoggins, line 447: This function tells you if you have all the repos or not.  If you want to gather all of the repositories, FIRST, you have to get a list of repos that exist, compare that to the repos you have, and gather the delta.

sgoggins: If you added a function to see if the repo count matches, you could skip the org.

augur/cli/db.py
```diff
@@ -331,3 +332,188 @@ def check_pgpass_credentials(config):
...
+    return data['data']['organization']['repositories']['totalCount']
+
+def get_data(query, org):
+
+    # Hit the graphql endpoint and retry 3 times in case of failure
+    pc = PlatformConnector()
+    pc.logger = logging.getLogger()
+    pc.initialize_database_connections()
+
+    url = 'https://api.github.com/graphql'
```
sgoggins, line 469: Double check if you need this or if its in the platform connector.

augur/platform_connector.py
```diff
@@ -0,0 +1,663 @@
+#SPDX-License-Identifier: MIT
```
sgoggins, line 1: As I noted in our discussion, I think this is a *really* wise refactoring. This puts all the API logic and a lot of the *shared* worker database logic *that is related to the API connection* and puts it in one place. 

If CLI logic, or other CORE Augur functions require API access in the future, (i.e., if we wanted to make PyPi packages out of all the workers, having this logic here would make that a **lot** easier.(

You know from our meeting last week that I like this change/refactor for those reasons.

sgoggins (approved):

ivanayov:
Hi @ccarterlandis, 
Thank you for the review.

Regarding the DB connections, those included in the `PlatformConnector` are only related to the GitHub/GitLab API access, as the `AccessToken`s  are stored in the database. As there are no general methods to get them and they are only needed in this case (so don't need to be put somewhere else) I added them to the `PlatformConnector`, so that it takes full care of accessing the GitHub/GitLab API. Those are the only connections available there.

The reason this is moved to an external class is that the API can be accessed out of the workers, like in this case from the CLI and/or the housekeeper for doing checks and updates that are more general and don't need to be part of a worker. This also allows using already implemented logic in the CLI.
The other benefit, I think, is for the `WorkerBase`, as this way the platform logic that's not specific for the workers is moved out of it and so its support. It makes the class smaller and easier to read, as well as related to the workers' logic only.

I'm happy to revisit if you have any proposals that would make the change better.
Also happy to help with the workers' refactoring if you need any help on it.

augur/cli/db.py
ivanayov: It's how all the workers approach that. Can be improved for sure and put into a config, as the number of workers using it grows. 
If it's going to turn into a too-much-refactoring for this PR, this can be done separately.

ccarterlandis (changes requested):
@ivanayov thanks for clarifying! The DB connections that are in the code you copied into `PlatformConnector` use SQLAlchemy's ORM. This is specific to how @gabe-heim  implemented the workers, as it made sense to do so; however, the rest of Augur does not use SQLAlchemy's ORM. Additionally, there is also already code that makes use of GitHub API keys (specifically in [`augur/routes/manager.py`](https://github.com/chaoss/augur/blob/860d72b81c0fe04dd7c67e7d1be97bc73e7c4ca4/augur/routes/manager.py#L207)) that does not use ORM either. So, even in existing places where we already need this shared API code, we cannot use the `PlatformConnector` code in its current form. Additionally, there are several significant performance and security issues we've identified with the worker OAuth logic in its current state, so that's even more reason to not use that code as-is. 

I will admit that mixing ORM and non-ORM database access is not a best practice most of the time and can be confusing, so we should discuss standardizing on whether we use ORM across the board or not - but that's another topic for another time. In the meantime, I'm thinking a good solution here would probably be to find a way that both the workers and the rest of Augur can use a common set of GitHub/GitLab API methods, whether they use ORM or not. That way we only have to implement it once, and can use the functionality everywhere. The config logic is already shared between the workers and the rest of Augur, so this is definitely a very do-able task. We can work out specific implementation details later.

I also don't think it makes the `WorkerBase` any easier to read, at least for me; in fact, I think it makes it harder. It only separates the code into two files without changing how anything functions in the worker, so now in order to understand the worker base one must flip back and forth between these two files. I would also argue that the functionality you copied IS specific to the workers, especially the `check_duplicates` function and the fact that the workers use ORM, while the rest of Augur does not. Those are two very key worker behaviors that I think make more sense when defined as part of the base worker class as they are vital to its functionality, and I'm also not certain that the API connector needs to be handling its own database connections.

Finally, I also think the worker's OAuth logic is simply over-engineered for the rest of Augur. For the most part, Augur only needs one - MAYBE two - keys to do some basic insertion or queries. The workers, however, are "power users" of the GitHub API, and churn through keys much faster than the rest of Augur, and also have to deal with many more edge cases and special assumptions. The rest of Augur simply doesn't need that logic right now, so I think we should follow YAGNI (You Aren't Gonna Need It) and not try to refactor all the complex logic for components that will most likely never take advantage of it.

Whew, that was a lot more than I expected...I know that was a lot, and I also left a lot of comments on your PR. Sorry about that! If you would like to discuss this over Zoom or Slack or something, to make it easier, I am happy to do that. @gabe-heim may also have ideas for the OAuth refactor; maybe we should open a discussion thread, so we can hash out those ideas together and come up with a more defined plan of action. I also really appreciate you adding this feature, I don't want to make it seem like I don't want this - I very much do! I just want to make sure we implement it in a way that will be the most useful and set us up for success in the future.

sgoggins:
@ivanayov : Checking on this isssue.

ivanayov:
@sgoggins I addressed the comments from our discussion, so should be ready for review. 
Will double-check and do some more testing tomorrow.

sgoggins (approved):

PR #1049, opened 2020-12-15
Machine Learning Workers and docs

sgoggins:
From Google Summer of Code 2020.


test.md
```diff
@@ -0,0 +1 @@
+dddd
```
ccarterlandis, line 1: This file needs to be deleted.

scripts/install/backend.sh
```diff
@@ -12,3 +12,20 @@ if [[ $target == *"prod"* ]]; then
...
+    python -m nltk.downloader stopwords > logs/stopwords.log
+    python -m nltk.downloader punkt >logs/punkt.log
+    python -m nltk.downloader popular >logs/popular.log
+    python -m nltk.downloader universal_tagset >logs/universal_tagset.log
+    echo "Done!"
+    ;;
+  *)
+    echo "Skipping frontend dependencies..."
+    ;;
+esac
```
ccarterlandis, line 31: There are several issues with this script that need to be addressed.
1. This section of code **CANNOT** go in `backend.sh`, for several reasons. First, it has nothing to do with Augur's backend; it's related to the workers. Second, this script is run as part of several automated builds (like the Docker build, the Travis build, and the `make rebuild` command) that do not have access to `stdin`, so reading user input is impossible and will cause those builds to fail. I recommend moving this code block to its own file, and running it **DURING THE REGULAR INSTALL PROCESS ONLY,** or when the stopwords are required.
2. The `logs/` directory that these words are written to gets completely cleared every time Augur starts, meaning that the words will be deleted before they are ever even used. They need to go in a different location, preferably somewhere close to where they are used in the relevant workers.
3. The output of those commands should not go into `.log` files. The `make clean` process will see those files and delete them; they need to be in `.txt` or some other form.
4. Line 29 is incorrect - these are not frontend dependencies. I assume this is left over from a c/p, but it sill needs to be fixed.
5. This section should output where it stored the words, for posterity's sake.

tests/test_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+def test_discover_config_file_env_no_exception():
+    test_config = default_config
+    with open(os.path.join(temp_dir, "augur.config.json"), "w") as f:
+        pass
+    config_object = AugurConfig(temp_dir, test_config)
+    assert config_object.discover_config_file() == os.path.join(temp_dir, "augur.config.json")
+
+
+
+
```
ccarterlandis, line 63: This file needs to be deleted. These tests break **EVERYTHING** and we do not use them.

workers/clustering_worker/trained_crf_model/directory.md
```diff
@@ -0,0 +1 @@
+for models
```
ccarterlandis, line 1: Why is this file here?

workers/discourse_analysis_worker/trained_crf_model/directory.md
```diff
@@ -0,0 +1 @@
+For trained models
```
ccarterlandis, line 1: Why is this file here?

ccarterlandis (changes requested):
Some comments inline, the rest below:
1. Everytime `.ipynb` file needs to be deleted.
2. Everything in the `time_series_notebooks` folder needs to be deleted.
3. There are a variety of miscellaneous model files (`.pkl`, `.h5`, etc) in the `message_insights` worker. All these unnecessary model files & binaries need to be deleted.

sgoggins:
> Some comments inline, the rest below:
> 
>     1. Everytime `.ipynb` file needs to be deleted.
> 
>     2. Everything in the `time_series_notebooks` folder needs to be deleted.
> 
>     3. There are a variety of miscellaneous model files (`.pkl`, `.h5`, etc) in the `message_insights` worker. All these unnecessary model files & binaries need to be deleted.

All are deleted. The `pkl` and `h5` files were added to the .gitignore at the root of hte project. These are modeling files that we would never want to commit. @aksh555 : Confirm if I am correct.

tests/test_helpers.py
sgoggins: Can you clarify how they break everything? Worker tests are generally a good thing, and we don't have enough of them. I assume they break everything because they are organized around the way we operated the augur.config.json and the old general worker. Thanks, @ccarterlandis

workers/clustering_worker/trained_crf_model/directory.md
sgoggins: Its a GitHub "trick".  If a directory is empty, it will not push it to the repository. However, we need this directory because that's where the models are generated. A better solution would be a `mkdir` command in the worker, however the risk there is installation on a file system where the user has restricted rights for making directories.  We have 2 secure installations where this is the case.

workers/discourse_analysis_worker/trained_crf_model/directory.md
sgoggins: @ccarterlandis Its a GitHub "trick". If a directory is empty, it will not push it to the repository. However, we need this directory because that's where the models are generated. A better solution would be a mkdir command in the worker, however the risk there is installation on a file system where the user has restricted rights for making directories. We have 2 secure installations where this is the case.

scripts/install/backend.sh
sgoggins: @ccarterlandis : I figured it was wrong, but was not certain why. Thanks for the clear explanation. 
I am working on a solution that addresses the installed base, who will only do `make rebuild-dev`, and do not want to introduce anything that makes installation or maintenance more complex. Nor do I want to install word like these without giving the user a choice. By default, they are placed in a user's home directory. I could specify a directory in `/usr/local/`, which sounds like it would be your preference. Its a worker's issue for sure. 

These CANNOT be installed using Python, at least according to my internet searches. I will find another way and see if it meets the salient criteria.

aksh555:
> > Some comments inline, the rest below:
> > ```
> > 1. Everytime `.ipynb` file needs to be deleted.
> > 
> > 2. Everything in the `time_series_notebooks` folder needs to be deleted.
> > 
> > 3. There are a variety of miscellaneous model files (`.pkl`, `.h5`, etc) in the `message_insights` worker. All these unnecessary model files & binaries need to be deleted.
> > ```
> 
> All are deleted. The `pkl` and `h5` files were added to the .gitignore at the root of hte project. These are modeling files that we would never want to commit. @aksh555 : Confirm if I am correct.

@sgoggins, Some of the model files/ datasets need to be persisted: workers/pull_request_analysis_worker/trained_pr_model.pkl, workers/message_insights_worker/train_data/doc2vec.model, workers/message_insights_worker/train_data/custom_dataset.xls, workers/message_insights_worker/train_data/EmoticonLookupTable.txt
The other files of the novelty model need not be committed. They get generated when the workers are run

workers/clustering_worker/trained_crf_model/directory.md
aksh555: Can have a .gitkeep too

workers/pull_request_analysis_worker/pull_request_analysis_worker.py
```diff
@@ -0,0 +1,268 @@
...
+        
+        df['merge_prob'] = [item[1] for item in merge_prob]
+
+        self.logger.info(f'Analysis done!')
+
+        # DEBUG:
+        # df.to_csv(f'PRA_{repo_id}.csv',index=False)
+
+        '''
+        # Insertion of merge probability to pull_request_analysis table
```
aksh555, line 208: @sgoggins This PR worker file is the older version, which doesn't insert the insights into the table. The updated one is in the gsoc-dev branch here https://github.com/chaoss/augur/blob/gsoc-dev/workers/pull_request_analysis_worker/pull_request_analysis_worker.py

test.md
sgoggins: @ccarterlandis : Done

scripts/install/backend.sh
sgoggins: I have a fix. I am testing it now.

workers/clustering_worker/trained_crf_model/directory.md
sgoggins: @aksh555 : Can I just put that in the root, or is it better in the subdirectory? We do *not* want to keep models from instance to instance. Unless my understanding of the models is wrong? I think they are tuned to local data.

workers/pull_request_analysis_worker/pull_request_analysis_worker.py
sgoggins: Thanks, @aksh555 !!!! I put the most recent code into the python file. 
👍🏼💯🏆

sgoggins: @aksh555 : If you look in the `gsoc-ml` branch (or this pR) you *should* see the right version.  If you have a minuet, please confirm.

aksh555: Yes, it's the right one!

workers/clustering_worker/trained_crf_model/directory.md
sgoggins: @aksh555 : I put a .gitkeep into the local worker directories that are empty when we begin. I think that fixes the issue.

test.md
sgoggins: removed

workers/pull_request_analysis_worker/pull_request_analysis_worker.py
sgoggins: @ccarterlandis : tl;dr -- some of the .pkl files and .h5 files are necessary for generally trained models. Some of the directories are needed, but empty at the start. For those, I added a .gitkeep file instead of just "some file" on Akshara's advice, which as I googled it is REALLY good advice. I chose to use them locally in the directories we need to keep.

scripts/install/backend.sh
sgoggins: @ccarterlandis : OK, I fixed this, and specified a directory as you suggested. 
On install: The user is given a choice. 
On rebuild-dev: The user is given a message informing them what is being done. 

In both cases I have created a new script under `scripts/install` to address the installation consistently. 

I'm not sure its perfect, so I look forward to your next review!

sgoggins: I have tested the solution twice locally.

workers/discourse_analysis_worker/trained_crf_model/directory.md
sgoggins: fixed with the .gitkeep file @aksh555 mentioned in the end. @ccarterlandis

sgoggins:
> > > Some comments inline, the rest below:
> > > ```
> > > 1. Everytime `.ipynb` file needs to be deleted.
> > > 
> > > 2. Everything in the `time_series_notebooks` folder needs to be deleted.
> > > 
> > > 3. There are a variety of miscellaneous model files (`.pkl`, `.h5`, etc) in the `message_insights` worker. All these unnecessary model files & binaries need to be deleted.
> > > ```
> > 
> > 
> > All are deleted. The `pkl` and `h5` files were added to the .gitignore at the root of hte project. These are modeling files that we would never want to commit. @aksh555 : Confirm if I am correct.
> 
> @sgoggins, Some of the model files/ datasets need to be persisted: workers/pull_request_analysis_worker/trained_pr_model.pkl, workers/message_insights_worker/train_data/doc2vec.model, workers/message_insights_worker/train_data/custom_dataset.xls, workers/message_insights_worker/train_data/EmoticonLookupTable.txt
> The other files of the novelty model need not be committed. They get generated when the workers are run
> **So easier would be to gitignore the .h5 files which are generated by the novelty worker per repo, and retain the .pkl files, which are fixed in number**

I believe i fixed this so those files a preserved. I think I had already done that, and then messed it up .. but its back. @aksh555

workers/clustering_worker/trained_crf_model/directory.md
ccarterlandis: `.gitkeep` is fine, thanks for clarifying

tests/test_helpers.py
ccarterlandis: 1. These are not worker tests, they have nothing to do with the workers. These are only dealing with the config discovery and generation logic.
2. They are poorly written, which causes them to interfere with other test cases. For one, none of the test clean up after themselves, leaving their temporary directories and files. Because these configs are not deleted, and are written to a poorly chosen location, all the other tests see the temporary config files and use them instead of the actual valid config.
3. It uses old ways of interacting with the config which are not relevant anymore.

I'm not currently interested in rewriting or reworking these tests, so I'd like for them to be removed. This is maybe the 3rd or 4th time they have tried to make their way into `dev`, as they are a consequence of when you merged `gsoc-dev` into `dev` at the end of the summer.

ccarterlandis: Each time there is a PR which would merge them, we've had to delete them. Trust me, you don't want these tests in `dev`.

.gitignore
```diff
@@ -154,3 +154,6 @@ workers/spdx_worker/dosocs2.conf
 workers/spdx_worker/augur-spdx/
 workers/spdx_worker/ex-raw.txt
 workers/spdx_worker/ex.json
+
+# Model files for machine learning
+workers/message_instights_worker/message_models/*.h5
```
ccarterlandis, line 159: Why did this get added to the dosocs config?

scripts/install/install.sh
```diff
@@ -42,6 +42,24 @@ fi
 scripts/install/frontend.sh
 scripts/install/api_key.sh
 
+echo "MACHINE LEARNING WORKER NOTE: Next, you will be asked to install some wordlists for the natural language processing parts of Augur.  Please be advised that the machine learning workers, including the clustering worker, discourse worker, messsage_analysis_worker, pull_request_analysis_worker, and the LSTM model of the insight_worker will not function fully without these lists."
+
+read -r -p "Would you like to install required NLTK word lists for machine learning workers? [Y/n] " response
+case "$response" in
+  [yY][eE][sS]|[yY])
+    echo "Installing... aritificial intelligence, machine learning. deep learning, and crocheting libraries and wordlists"
```
ccarterlandis, line 50: A few things on this line:
- `Artificial` is misspelled
- Please remove the reference to "crocheting"
- Please replace the period after "machine learning" with a comma

scripts/install/install.sh
```diff
@@ -42,6 +42,24 @@ fi
...
+    echo "Please checkout https://www.ravelry.com/account/login while you wait. Sure, you may not knit yet, but you know you want to."
```
ccarterlandis, line 51: Please remove this line, it is not relevant to the install.

scripts/install/install.sh
```diff
@@ -42,6 +42,24 @@ fi
...
+    sleep 2
```
ccarterlandis, line 52: Why is there a sleep command here?

scripts/install/nltk_dictionaries.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -eo pipefail
+
+echo "MACHINE LEARNING WORKER NOTE: Next, you will be asked to install some wordlists for the natural language processing parts of Augur.  Please be advised that the machine learning workers, including the clustering worker, discourse worker, messsage_analysis_worker, pull_request_analysis_worker, and the LSTM model of the insight_worker will not function fully without these lists."
```
ccarterlandis, line 4: This line is a duplicate of line 45 of `scripts/install/install.sh`. Please remove one or the other.

scripts/install/nltk_dictionaries.sh
```diff
@@ -0,0 +1,32 @@
...
+    python -m nltk.downloader stopwords -d /usr/local/share >> ~/nltk_wordlist_installations/stopwords.log
+    cp /usr/local/share/stopwords /usr/local/share/nltk-wordlists
+    python -m nltk.downloader punkt -d /usr/local/share  >> ~/nltk_wordlist_installations/punkt.log               
+    cp /usr/local/share/punkt /usr/local/share/nltk-wordlists
+    python -m nltk.downloader popular -d /usr/local/share >> ~/nltk_wordlist_installations/popular.log
+    cp /usr/local/share/popular /usr/local/share/nltk-wordlists
+    python -m nltk.downloader universal_tagset -d /usr/local/share >> ~/nltk_wordlist_installations/universal_tagset.log
+    cp /usr/local/share/universal_tagset /usr/local/share/nltk-wordlists
+    echo "Done!"
+    echo "Logs are stored in your home directory under '/usr/local/share, /usr/local/share/nltk-worklists, and in some instances based on operating system, your home directroy'."
```
ccarterlandis, line 21: Okay, there are still problems with this section.
1. Like I mentioned last time, they should not be going into `.log` files. Those have a very specific connatation and most programs will not hestitate to delete log files if needed, as they are generally viewed as temporary. **These files do not fit this profile, we need to make sure they are persisted and there is little chance of them being deleted on accident. Please change the file extension.**
2. Storing things in `/usr/local/` is usually fine, but not when each word gets its own folder, to me that is a waste of space. On top of that, there is already a `$HOME/.augur/` folder to store Augur-specific data. I should have recommended this explicitly last time; please store the stopwords in a folder under `$HOME/.augur/` (also, use `$HOME` instead of `~`, it is more standardized across platforms). 99% of the time, the user will have the correct access to write to this location, and the 1% of times they don't is not really an issue right now as that case is EXCEEDINGLY rare.
3. Instead of writing to output files, then copying them to a new location, just write them to the output location directly. Like: `python -m nltk.downloader universal_tagset -d $HOME/.augur/nltk-wordlists/`. This will write the output folder directly to `$HOME/.augur/nltk-wordlists/`, meaning we can skip the redundant copy step. This also lets us be consistent across platforms.

scripts/install/nltk_dictionaries.sh
```diff
@@ -0,0 +1,32 @@
...
+    ;;
+  *)
+    echo "Skipping NLTK Word Vector dependencies. Please be advised that the machine learning workers, including the clustering worker, discourse worker, messsage_analysis_worker, pull_request_analysis_worker, and the LSTM model of the insight_worker will not function fully without these lists."
+    ;;
+
+esac
+
+
+echo "**********************************"
+echo "***** NLTK DICTIONARY INSTALLATION  COMPLETE *****"
```
ccarterlandis, line 31: Please fix the spacing between `INSTALLATION` and `COMPLETE`, and extend the length of the top and bottom lines to match the length of the middle line.

scripts/install/workers.sh
```diff
@@ -39,4 +39,11 @@ do
       fi
 
     fi
+
+if [[ ! -e ~/nltk_wordlist_installations/stopwords.log ]]; then
```
ccarterlandis, line 43: This should not be looking for the existence of a log file, that is not a stable test. It should be checking that the installed directory is not empty; this can be done like so:
```bash
#!/bin/bash
FILE=""
DIR="$HOME/.augur/nltk-wordlists"
if [ "$(ls -A $DIR)" ]; then
 echo "Take action $DIR is not Empty"
else
 echo "$DIR is Empty"
fi
```

setup.py
```diff
@@ -49,7 +49,10 @@
         "gunicorn",
         "six>=1.14.0",
         "bokeh",
-        "selenium"
+        "selenium",
+	"nltk",
```
ccarterlandis, line 53: Please fix the indentation here. Also, why have `h5py` and `scipy` now been added?

ccarterlandis (changes requested):
@sgoggins here is my next review. There are still some issues that need to be addressed

tests/test_helpers.py
sgoggins: @ccarterlandis : OK, I will remove the tests, and maintain them in a different branch to be written along the lines you describe at a later time. We did work hard on the tests, even though we got it wrong. So, I will declare victory in the attempt!  EVERYONE WINS A TROPHY!

sgoggins: @ccarterlandis Thanks for explaining, btw.  I sincerely wanted to understand.

scripts/install/install.sh
sgoggins: *but I really wanted to promote a random knitting website at 3:30am this morning* ... ok. Will do.

scripts/install/install.sh
sgoggins: To give people time to write down the ravelry.com website. I should not attempt humor in code. I will remove.

scripts/install/nltk_dictionaries.sh
sgoggins: @ccarterlandis : The only thing I will point out is that 
1. I don't need to copy them to their own directory.  That was more so I could keep track of what I installed
2. NLTK data sets / word sets, have a strict regimine of where they will be found in the path. 
 - The users' home directory, under `nltk_data` or one of these options: 
 -  /usr/share/nltk_data',
 - /usr/local/share/nltk_data
 - /usr/lib/nltk_data
 - /usr/local/lib/nltk_data

So, of those, which do you prefer. Remember that many people may share these datasets for reasons other than Augur, and so that's why the NLTK default does not include an application directory.

scripts/install/nltk_dictionaries.sh
sgoggins: I love you man, but I swear to got I'm getting a C- from my high school English teacher every time I get this kind of PR Feedback.

Its absolutely necessary. I am entirely undisciplined about things like this and its vital that we have discipline as a project.  I appreciate the shit out YOU and this kind of feedback!!!

But, inside, somewhere, there's a 15 year old Sean going "the rules of English are dumb! I WILL NEVER USE THEM!!!"  :D

scripts/install/install.sh
sgoggins: removed.

tests/test_helpers.py
sgoggins: I have to fix this yet.

sgoggins (commented):
The updated .gitignore should ignore teh .pkl and .h5 files.  Will test and push.

PR #1056, opened 2020-12-21
Update repo and organization redirects

### LINKED ISSUE #915: Improved handling of redirects for orgs and repos
While the GitHub API automatically follows redirects and continues gathering data from the newly redirected location, this creates unexpected behavior for users. For example, someone looking in pivotal.osshealth.io for the repo group vmwarepivotallabs would think that the data is missing. However, they could find the data by looking for the pivotalservices repo group, which was the original name of the GitHub org. In this case, all of the repositories at https://github.com/pivotalservices have been redirected to https://github.com/vmwarepivotallabs

Where this gets tricky is that GitHub redirection seems to happen at the repository level, not at the org level. Because of the reirection, https://github.com/pivotalservices looks like an empty org, but you can see that if you go to one of the original repositories, like https://github.com/pivotalservices/binder, you'll be redirected to the corresponding repo in vmwarepivotallabs, in this case: https://github.com/vmwarepivotallabs/binder

In addition to displaying the new information correctly in the front end, the corresponding database fields will also need to be updated. For example: repo_groups.rg_name, repo_groups.description, repo_groups.website, repo.repo_path, repo.repo_name, repo.url, and probably others that I'm not noticing.

There are also a few different cases for redirection that should be considered. The above example talks about moving an entire org and all of the repos within it, which is probably the most complex case, but there are other simpler cases:

- A single repo moved to a new org. Example: http://github.com/Spring-projects/spring-batch-admin-samples to https://github.com/spring-attic/spring-batch-admin-samples
- A single repo renamed within the same org. Example: http://github.com/Pivotal-Field-Engineering/word_cloud-spring-kafka-k8s-workshop to https://github.com/Pivotal-Field-Engineering/word_cloud-spring-kafka-on-k8s

ivanayov:
This change updates the Augur database in case and org or repository
change their name. Then the old url redirects to the new one, but was
still visible in the database and UI.

To run the housekeeper job, add the followng to the `augur.config.json`:

```
            {
                "delay": 1000000,
                "given": [
                    "github_url"
                ],
                "model": "update_redirects",
                "repo_group_id": <your_repo_group_id>
            }
```

Resolves #915

### DIFF
--- a/augur/housekeeper.py
+++ b/augur/housekeeper.py
@@ -13,6 +13,7 @@ import pandas as pd
 from sqlalchemy import MetaData
 
 from augur.logging import AugurLogging
+from urllib.parse import urlparse
 
 import warnings
 warnings.filterwarnings('ignore')
@@ -27,6 +28,7 @@ class Housekeeper:
         self._processes = []
         self.augur_logging = augur_app.logging
         self.jobs = deepcopy(augur_app.config.get_value("Housekeeper", "jobs"))
+        self.update_redirects = deepcopy(augur_app.config.get_value("Housekeeper", "update_redirects"))
         self.broker_host = augur_app.config.get_value("Server", "host")
         self.broker_port = augur_app.config.get_value("Server", "port")
         self.broker = broker
@@ -45,6 +47,10 @@ class Housekeeper:
         """)
         rs = pd.read_sql(repoUrlSQL, self.db, params={})
         all_repos = rs['repo_git'].values.tolist()
+
+        # If enabled, updates all redirects of repositories 
+        # and organizations urls for configured repo_group_id
+        self.update_url_redirects()
 
         # List of tasks that need periodic updates
         self.schedule_updates()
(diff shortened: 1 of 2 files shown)


ccarterlandis:
@ivanayov this looks promising! I will test this locally later - do you have any handy repos you used while testing? Also, @gabe-heim curious to get your thoughts on this. Would we want some sort of "automatic" redirect update mechanism so we don't have to rely on this job firing before all the others, or is the sequencing work you've done far enough along that we can use that? Just spitballing 😄

augur/housekeeper.py
```diff
@@ -336,3 +338,35 @@ def prep_jobs(self):
...
+
+        return repos_list
+
+    def update_repo_url(self, old_url, new_url):
+        update_sql = s.sql.text("""
+                UPDATE repo.repo_git SET VALUE = '{}
+                WHERE repo_git = '{}'
+            """.format(new_url, old_url))
+
+        repos = self.db.execute(update_sql)
```
sgoggins, line 371: This looks like its going to take the revised URL reported by GitHub and make the change in the `repo` table when a repository is moved. Looks Quite Helpful and I'd like to test it out. @gabe-heim ? @ccarterlandis ?

sgoggins (approved):

augur/housekeeper.py
```diff
@@ -336,3 +338,35 @@ def prep_jobs(self):
...
+        if len(repos) == 0:
+            logger.info("Did not find any repositories stored in augur_database for repo_group_id {}\n".format(repo_group_id))
+
+        repos_list = list(map(get_value_from_fetch_data, repos))
+
+        return repos_list
+
+    def update_repo_url(self, old_url, new_url):
+        update_sql = s.sql.text("""
+                UPDATE repo.repo_git SET VALUE = '{}
```
sgoggins, line 367: Close quote missing.

ccarterlandis:
@ivanayov: after chatting with @sgoggins and @gabe-heim, I think this functionality would work better just as a configurable part of the housekeeper initialization, and not as a regular worker job. So far, we've been following the pattern that all jobs correspond directly to at least worker, and all jobs and workers are also primarily concerned with data collection instead of cleanup. Implementing the Git URL cleaup process as a job without also implementing it in a worker would break this pattern, and I think that keeping cleanup functionality as part of the standard housekeeper API makes sense within the rest of its scope.

To that end, our proposal is to add some additional keys in the Housekeeper config:
```
"Housekeeper": {
        "update_redirects": {
            "switch": 1,
            "repo_group_id": 0
        },
        "jobs": [
            ...
        ]
    }
```

And then refactor [lines 165-166](https://github.com/chaoss/augur/pull/1056/files#diff-671267b9aedba45012cb8584bc117fab33bbae4c2867195a999bf71852c3adc9R166-R167) to use these keys, and refactor pull them out into a function that gets called before `self.schedule_updates()` in the Housekeeper's `__init__` method.

We did also briefly talking about moving into its own worker to keep the pattern, but felt that this was not really a difficult enough task to warrant a worker. Additionally, there would be a problem with making sure this job ran first, as currently the order the broker sends the jobs to the workers is non-deterministic due to the workers not having a guaranteed startup order. We would have to add some sort of job sequencing capability to ensure consistent behavior, and even though it's on our roadmap for other reasons, we don't feel ready to tackle it just yet.

Would love to hear your thoughts on this! Also, sorry it took me a while to get back to you 😅

augur/housekeeper.py
```diff
@@ -336,3 +340,36 @@ def prep_jobs(self):
                 job['repos'] = rs
             # time.sleep(120)
 
+    def update_url_redirects(self):
+        if 'switch' in self.update_redirects and self.update_redirects['switch'] == 1 and 'repo_group_id' in self.update_redirects:
```
sgoggins, line 344: Ivana is fixing.

PR #1063, opened 2020-12-22
No Data Fix: Image-Apis

ABrain7710:
Added logic so that the image api endpoints return an error, if there is no data to display


sgoggins (approved):
LGTM

PR #1067, opened 2020-12-23
Image api docs patch

ABrain7710:
Change tag for community reports endpoints to "visualizations", instead of "experimental". This helps people to be able to find the vis docs easier


sgoggins:
We need to add me as a code-owner for these API docs because these are things that end users really need, and we are not merging them quickly enough.  In this case, Andrew did exactly what we discussed, @ccarterlandis ... so I think its legit.

ccarterlandis:
> We need to add me as a code-owner for these API docs because these are things that end users really need, and we are not merging them quickly enough.

Respectfully, I am not comfortable at all with this. You have not proven to me that you understand the docs architecture well enough for me to feel comfortable letting you be a code-owner for these files. "Not merging them quickly enough" is not only **not** an acceptable reason to be merging things that need my approval before they have it, especially when this PR was literally open for **less than 5 minutes**, but something that I just don't believe, as if I haven't merged a PR, there is ALWAYS a good reason, which is usually that it requires changes. I'm not sure how fast you want things to be reviewed, but 5 minutes is not a reasonable amount of time to let me both see this PR came through and review it.

PR #1078, opened 2021-01-04
gsoc-ml+gitlab-merge+releases-patch-1

sgoggins:
This pull request fixes installation issues for the gitlab worker and releases worker.


augur/config.py
```diff
@@ -33,6 +33,34 @@
                     "model": "issues",
                     "repo_group_id": 0
                 },
+                {
+                    "all_focused": 1,
```
ccarterlandis, line 37: What does this `all_focused` key actually do? I don't see it used anywhere in the code for these workers. Also, this is a separate issue, but I don't like the name `all_focused`. It's not descriptive at all and as a new user I would have no idea what it would change. The name doesn't need to convey literally everything it does, but something like `collect_all_issues` or `complete_collection` would be far more useful IMHO.

augur/config.py
```diff
@@ -33,6 +33,34 @@
...
+                    "delay": 150000,
+                    "given": [
+                        "git_url"
+                    ],
+                    "model": "merge_requests",
```
ccarterlandis, line 42: Is this **GitLab** merge requests, or just merge requests of any kind? Context clues tell me it's just for GitLab, so I think we need to be very specific both in the model and in the given. Merge request is of course a very common FOSS term and could create confusion if we do no use it carefully.

augur/config.py
```diff
@@ -33,6 +33,34 @@
...
+                    "repo_group_id": 0
+                },
+                        
+                {
+                    "all_focused": 1,
+                    "delay": 150000,
+                    "given": [
+                        "git_url"
+                    ],
+                    "model": "merge_request_files",
```
ccarterlandis, line 61: `all_focused` and merge request terminology comments apply here too.

augur/config.py
```diff
@@ -129,125 +157,123 @@
...
+            		"model" : "discourse_analysis",
+            		"repo_group_id" : 0
+	           },
+	            {
+                    "delay": 10000,
+                    "given": [
+                        "git_url"
+                    ],
+                    "model": "clustering",
+                    "repo_group_id": 0
```
ccarterlandis, line 174: This indentation is incorrect. The brackets appear too far too the left, and the code appears to be too far to the right.

augur/config.py
```diff
@@ -129,125 +157,123 @@
...
+                    "workers": 1,
+        	    "max_df" : 0.9,
+        	    "max_features" : 1000,
+        	    "min_df": 0.1,
+        	    "num_clusters" : 4
 
-        }
-        },
+                }
+            },
```
ccarterlandis, line 276: This indentation is also incorrect. It needs to follow the formatting of the rest of the object.

schema/generate/33-schema_update_35.sql
```diff
@@ -0,0 +1,4 @@
+-- Adding GitLab Platform
+INSERT INTO "augur_data"."platform" ("pltfrm_id", "pltfrm_name", "pltfrm_version", "pltfrm_release_date", "tool_source", "tool_version", "data_source", "data_collection_date") VALUES (25151, 'GitLab', '3', '2020-12-27', 'Manual Entry', 'Sean Goggins', 'GitLab', '2020-12-27 16:07:20');
+--  # Pull request commit updates
+update "augur_operations"."augur_settings" set value = 35 where setting = 'augur_data_version';
```
ccarterlandis, line 4: This needs to be added to the oneshot schema install script. And probably some other stuff we forgot to add too...

scripts/install/nltk_dictionaries.sh
```diff
@@ -0,0 +1,25 @@
...
+  *)
+    echo "Skipping NLTK Word Vector dependencies. Please be advised that the machine learning workers, including the clustering worker, discourse worker, messsage_analysis_worker, pull_request_analysis_worker, and the LSTM model of the insight_worker will not function fully without these lists."
+    ;;
+
+esac
+
+
+echo "**********************************"
+echo "***** NLTK DICTIONARY INSTALLATION  COMPLETE *****"
+echo "**********************************"
```
ccarterlandis, line 25: This code is also present in PR #1049. I don't know which version is correct.

setup.py
```diff
@@ -49,7 +49,10 @@
         "gunicorn",
         "six>=1.14.0",
         "bokeh",
-        "selenium"
+        "selenium",
+	"nltk",
```
ccarterlandis, line 53: Please fix the indentation on this line.

ccarterlandis (dismissed):

augur/config.py
sgoggins: @gabe-heim picked the name. Its for the PR and issues workers to go make sure they are getting everything done.

augur/config.py
sgoggins: This is *GitLab*

schema/generate/33-schema_update_35.sql
sgoggins: ok but that wouldn't prevent this from being merged. Where is the oneshot?

scripts/install/nltk_dictionaries.sh
sgoggins: They are the same.  I've merged that commit in that other branch as well.

augur/config.py
ccarterlandis: Okay, then I think we should change the model to `gitlab_merge_requests` and the given to `gitlab_url` to be as clear as possible.

sgoggins (commented):
@ccarterlandis : Fixed all the indentation.

augur/config.py
ccarterlandis: So what happens when it's off? Does it just...not collect everything? Like just a subset of the data or? 
And does this worker need it? I assume so, since it's about PRs (well MRs) but I guess what I'm trying to get at is that because the name of this key is confusing, I can't tell if this is actually supposed to be here or not.

augur/config.py
ccarterlandis: I don't think is resolved, I still think this needs to be changed because it is confusing otherwise. Or at least, I haven't been convinced otherwise yet. 😉

augur/config.py
ccarterlandis: This is also unresolved, for the same reasons as above.

schema/generate/33-schema_update_35.sql
ccarterlandis: Again, this is not resolved. It needs to be in the oneshot as we made that with the intention to keep it up to date with the small update scripts, otherwise we ship 2 different versions of the schema and that's not good for anyone. It's in the `schema/` folder, same as all the other schema files. It's called `20201025-Release-0.14.0.sql`

setup.py
```diff
@@ -49,7 +49,10 @@
         "gunicorn",
         "six>=1.14.0",
         "bokeh",
-        "selenium"
+        "selenium",
+	    "nltk",
```
ccarterlandis, line 53: Please fix the formatting on this line.

scripts/install/nltk_dictionaries.sh
ccarterlandis: This is not resolved either. Please fix the formatting as outlined in [this comment](https://github.com/chaoss/augur/pull/1049/files?file-filters%5B%5D=.sh#r544597459) on PR #1049.

ccarterlandis (changes requested):
Request changes are in each comment thread.

augur/config.py
sgoggins: You're asking questions about the basic design @gabe-heim has in place. I'm not doing anything new or special here. This is a switch to *ensure* that we get everything on the first run. Its been there for over two years.

augur/config.py
sgoggins: Same answer as above. Its been there over two years. Now is not the time to stop a pull request because of it. Ask @gabe-heim .

scripts/install/nltk_dictionaries.sh
sgoggins: extra space corrected.

setup.py
sgoggins: fixed.

ccarterlandis (changes requested):
Not everything is fixed and I still have several unanswered questions. By merging `gsoc-patch` you also introduced more code that has issues, so I have added my review to this new code as well. Not much, but it still needs to be fixed.
I will approve and merge it when all my questions, comments, and requested changes have been appropriately addressed.

augur/config.py
ccarterlandis: After further thought this doesn't need to block this PR, as we should do the same thing for GitHub. We can do that later. But I still think it needs to be changed.

augur/config.py
sgoggins: @gabe-heim : Please pull this branch and fix it, or merge your PR branch and I'll merge that here first.

PR #1087, opened 2021-01-14
fixing augurface

sgoggins:
The path to the configuration file `frontend.config.json` was incorrect, and while the variable `config` was set in `src/store/UtilModule.js`, the variable `configObject` was required. 

PR also includes audit fixes from NPM.


ccarterlandis:
This fix looks good to me, but now I'm realizing that the config does not get written to the `augurface` directory when it's generated, only `augur/frontend/`. I can fix this - should both frontends use the same frontend config file, or different ones?

sgoggins:
@ccarterlandis : Per our discussion last  night, I think keeping the frontend.config.json in one place makes it easier. This PR accomplishes that goal.

sgoggins:
@ccarterlandis : Do you know why the Docker tests are not running?

PR #1093, opened 2021-01-22
Pull request reports: outliers patch

ABrain7710:
Previously two of the endpoints defaulted to removing 10 outliers, which would cause an error if there was less than 10 data points. So I created a function that removes outliers based on standard deviations. I currently have it set to remove values that are more than 3 standard deviations from the mean. I also have it set to remove outliers by default.


sgoggins (dismissed):

sgoggins:
@ABrain7710 : What changes from this PR are needed.  There's a bunch of GSOC-PATCH branch stuff in here that we can't merge yet.

PR #1122, opened 2021-02-16
Gsoc patch

sgoggins:


sgoggins:
Update to numpy, where "np" was used in code, but not used in the import statement.

Issue #1018, opened 2020-11-18
`augur backend start` prints incorrect address and port


zachs18:
Please help us help you by filling out the following sections as thoroughly as you can.

**Description:**
`augur backend start` prints the wrong address and port.

**How to reproduce:**
1. Change the backend server port in augur.config.json
2. Start the backed with `augur backend start`

**Expected behavior:**
Augur prints the address and port it is running at.

**Actual behavior:**
Line 46 of `augur/cli/backend.py` prints the string `"Augur is running at: http://0.0.0.0:5000"` no matter what port/address it is configured to use.

A temporary hack-fix is to change line 46 from
`logger.info(f"Augur is running at: http://0.0.0.0:5000")` to `logger.info(f"Augur is running at: {augur_app.gunicorn_options['bind']}")`

**Software versions:**
 - Augur: dev branch commit 77e7ea8a5ef48b71090a17e6a3f55c3f9165788c
 - OS: Ubuntu 20.04.1

Sherlock2505:
I want to work on this issue can you please assign this to me?

AllMight2099:
I feel that the temporary hack fix could actually be the permanent solution.  Could you elaborate on how you think it's a temporary fix considering that you just need to send in the port number as a variable in the print statement

sgoggins:
Closing issue. its fixed in PR #1133

PR #1133, opened 2021-02-21
Fix hardcoded port logging on startup

ccarterlandis:
Fixes issue #1018. Also contains indentation fixes for the default config, and updates some Makefile commands to print the commands they are running for transparency.


Ulincsys (approved):
Looks good to me. Maybe it would be better to specify `localhost` instead of `127.0.0.1`?

sgoggins (approved):

Issue #1174, opened 2021-03-12
Start of fresh database (via docker) fails with SQL error in `43-schema_update_45.sql`


maxhbr:
**Description:**
On latest master (https://github.com/chaoss/augur/commit/f2bafd8e7b2c2443354118ee742c129007c28099) the latest database migration (https://github.com/chaoss/augur/blob/master/schema/generate/43-schema_update_45.sql) fails with `ERROR:  zero-length delimited identifier at or near """" at character 31`.

**How to reproduce:**
```
$ make docker-build-database
[...]
$ make docker-run-database
[...]
/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/42-schema_update_44.sql
ALTER TABLE
ALTER TABLE
COMMENT
CREATE INDEX
psql:/docker-entrypoint-initdb.d/42-schema_update_44.sql:9: NOTICE:  relation "probability_idx" already exists, skipping
UPDATE 1


/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/43-schema_update_45.sql
CREATE SEQUENCE
psql:/docker-entrypoint-initdb.d/43-schema_update_45.sql:7: NOTICE:  relation "worker_oauth_oauth_id_seq" already exists, skipping
COMMENT
COMMENT
COMMENT
COMMENT
ALTER TABLE
COMMENT
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
COMMENT
COMMENT
2021-03-12 10:03:38.300 UTC [462] ERROR:  zero-length delimited identifier at or near """" at character 31
2021-03-12 10:03:38.300 UTC [462] STATEMENT:  CREATE SEQUENCE IF NOT EXISTS ""augur_operations"."affiliations_corp_id_seq""
        INCREMENT 1
        MINVALUE  1
        MAXVALUE 9223372036854775807
        START 620000
        CACHE 1;
psql:/docker-entrypoint-initdb.d/43-schema_update_45.sql:41: ERROR:  zero-length delimited identifier at or near """"
LINE 1: CREATE SEQUENCE IF NOT EXISTS ""augur_operations"."affiliati...
                                      ^
make: *** [Makefile:157: docker-run-database] Error 3
```

**Software versions:**
 - Augur: latest master https://github.com/chaoss/augur/commit/f2bafd8e7b2c2443354118ee742c129007c28099
 - OS: dockerized on NixOS

[...]

PR #1175, opened 2021-03-12
fix syntax error in generated schema

maxhbr:
closes: #1174


maxhbr:
Should that be rebased on `dev` and pointing to it?

PR #1205, opened 2021-03-23
Add '.git' suffix check for repo redirects

ivanayov:
This change checks if repo url has `.git` suffix and removes it
for the repo redirects update

Signed-off-by: Ivana Atanasova <iyovcheva@iyovcheva-a02.vmware.com>


sgoggins:
I'll merge into dev from dev-redirect branch locally after I test it. Unless that messes up the commit credit, but I don't think it does.

ivanayov:
@sgoggins It's ok, whatever works better for you.

sgoggins:
@ivanayov : I sent a slack message with details, but the update does not seem to put the changed URLs into a 'new' status.

PR #1206, opened 2021-03-23
Add '.git' suffix check for repo redirects

ivanayov:
This change checks if repo url has `.git` suffix and removes it
for the repo redirects update

Signed-off-by: Ivana Atanasova <iyovcheva@iyovcheva-a02.vmware.com>


sgoggins:
@ivanayov : I merged this into dev-redirect and will test again. Could we talk on Monday? I can be up early my time? The testing on this one is complex, and I think it requires both of our eyes to be sure we are not messing something up.

PR #1238, opened 2021-04-22
Checked if the url is a gitlab url

Dhruv-Sachdev1313:
Signed-off-by: Dhruv-Sachdev1313 <dhruvhsachdev@gmail.com>

Many API calls were being wasted as the url wasn't a gitlab url in gitlab_merge_request_worker. So I have added a if condition to check the url . I think this should work , but do go through the changes once please.


sgoggins (approved):
This fixes an issue that used to waste a lot of API calls. Thanks, @Dhruv-Sachdev1313 !!

PR #1270, opened 2021-05-31
updating pr_worker_redo_test from dev

sgoggins:


sgoggins (commented):
@gabe-heim : @ABrain7710 Found that one of the changes to worker_base made it so workers that took only a git_url did not run. That change is addressed in this PR. Please review.

sgoggins (commented):
LGTM

PR #1287, opened 2021-06-21
Branch patch facade + scipy version update

sgoggins:
Added code to the commit worker, Facade, to checkout the `main` branch if a repo no longer has a `master` branch.


sgoggins:
Verified that it works.

PR #1291, opened 2021-06-23
Refactor worker_base into two classes. Worker and WorkerGitInterfaceable

IsaacMilarky:
The base class is now split into two about equally long halves. The base class doesn't have as many attributes and offloads variables like self.headers to be defined by WorkerGitInterfaceable's constructor. The classes that require WorkerGitInterfaceable are:

gitlab_issues_worker.py
gitlab_merge_request_worker.py
pull_request_worker.py
release_worker.py
repo_info_worker.py

I couldn't get the unit tests to function properly but it seems to work with some example API queries that I have been throwing at it.


sgoggins (changes requested):
Could you revert the changes to the facade worker? I can explain, or maybe we can talk it through.  It does function slightly differently than other workers because its a refactoring of https://github.com/brianwarner/facade

PR #1307, opened 2021-07-02
Extended Deps Worker to Collect Scorecard Data 

Dhruv-Sachdev1313:
Extended Deps Worker to collect Scorecard Data - 

Things I have changed - 
Installation Scripts 
added a check that if 'go' package exists to run scorecard.
Added script to check if scorecard exists already , if not clone and build 
in deps worker used Subprocess to call scorecard and made necessary env settings 

Please NOTE: It is not populating the database table for some reason now, but the program works that is for sure


Dhruv-Sachdev1313:
Changed the Schema to v54 which adds a new table with a new format for storing Scorecard Data

PR #1364, opened 2021-07-20
Master dev merge patch 23 - selective updates from `master` back into `dev` with Isaac

sgoggins:


sgoggins:
@IsaacMilarky : This still seems like not what we did. The `new_` is still in here among other things.

PR #1392, opened 2021-08-02
Andrew and I spent a lot of time rewriting enrich_cntrb_id to be more maintainable and stop duplicate inserts.

IsaacMilarky:
Andrew and I rewrote enrich_cntrb_id to be more readable by using more for loops instead of complicated table merges. In the future it might be wise to refactor this method out into its own worker class complete with a queue. Presently, we have had issues with more than one worker using this method at a time and causing collisions (i.e. pr worker thinks a contributor wasn't in the database but the GitHub worker inserts that contributor right after the check so the pr worker inserts a duplicate and throws an integrity error.) The main errors this branch can throw now have to do with integrity violations in the bulk inserts for other models not related to enrich_cntrb_id.

### DIFF
--- a/workers/github_worker/github_worker.py
+++ b/workers/github_worker/github_worker.py
@@ -1,537 +1,1790 @@
-#SPDX-License-Identifier: MIT
-from multiprocessing import Process, Queue
-from urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


sgoggins:
Looks good to me!

PR #1393, opened 2021-08-02
Fixed edge case causing ValueErrors

IsaacMilarky:
Fixed edge case with bulk insertions. I didn't make the bug happen but my changes in the bulk insert made it happen way more often. The insertion method checks to see if source_issues['all'] isn't empty but not if source_issues['insert'] is empty. So when it tried to enrich_cntrb_id source_issues['insert'] was empty causing a ValueError.


sgoggins (approved):
LGTM

PR #1394, opened 2021-08-03
Fixing of Value Error's Caused by Shared Library

IsaacMilarky:
I am continuing to debug Dev. There are more cases where inserts aren't being checked before being sent to enrich_cntrb_id than I thought. I have added checking to all the places that call enrich_cntrb_id.


sgoggins (approved):

sgoggins:
LGTM @IsaacMilarky !

PR #1397, opened 2021-08-04
Added support for alternate identifiers for source data when enriching contributor.

IsaacMilarky:
Different types of data have different identifiers so now if the source data doesn't have a user.id the method will instead use the user.node_id to determine whether or not the user is in the contributors table. We can't just check one key for all of them either since some data doesn't have a user.node_id and some data doesn't have a user.id.


sgoggins:
LGTM

PR #1401, opened 2021-08-05
Fix typo in docs

ABrain7710:
The documentation for the committers indicated that the endpoint should be called with http://localhost:5000/api/unstable/repo-groups/1/Committers but the correct url is http://localhost:5000/api/unstable/repo-groups/1/committers. The only difference being the c in committers


Ulincsys (approved):
Approved

PR #1420, opened 2021-08-12
Made docker docs more current before I start making QoL improvements

IsaacMilarky:
Updated docker docs while I think about how the container will work when it is more usable

Signed-off-by: Isaac Milarsky <imilarsky@gmail.com>


sgoggins (approved):

sgoggins:
These silk be helpful.

Issue #1423, opened 2021-08-13
enrich_cntrb method not checking for additional tokens


sgoggins:
Please help us help you by filling out the following sections as thoroughly as you can.

**Description:**
If GH/GL worker collection fails on retrieval of contributor_id from the platform, it issues this error and stops the worker.

```
2021-08-12 16:25:43,170,170ms [PID: 279416] workers.github_worker.57516 [INFO] Contributor data: {'message': 'API rate limit exceeded for user ID 75419635.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
```

Embedded in this error: 

```
2021-08-12 16:25:42,887,887ms [PID: 279416] workers.github_worker.57516 [INFO] Hitting endpoint: https://api.github.com/users/toufikImk ...

2021-08-12 16:25:43,170,170ms [PID: 279416] workers.github_worker.57516 [INFO] Contributor data: {'message': 'API rate limit exceeded for user ID 75419635.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
2021-08-12 16:25:43,171,171ms [PID: 279416] workers.github_worker.57516 [ERROR] Worker ran into an error for task: {'job_type': 'MAINTAIN', 'models': ['issues'], 'display_name': 'issues model for url: https://github.com/angular/angular', 'given': {'github_url': 'https://github.com/angular/angular'}, 'focused_task': 1}
2021-08-12 16:25:43,171,171ms [PID: 279416] workers.github_worker.57516 [ERROR] Worker was processing this task for -2498.578952550888 seconds.
2021-08-12 16:25:43,171,171ms [PID: 279416] workers.github_worker.57516 [ERROR] Printing traceback...
2021-08-12 16:25:43,171,171ms [PID: 279416] workers.github_worker.57516 [ERROR] 'login'
2021-08-12 16:25:43,173,173ms [PID: 279416] workers.github_worker.57516 [ERROR] Traceback (most recent call last):
  File "/home/sean/github/operatefirst/workers/worker_base.py", line 180, in collect
    model_method(message, repo_id)
  File "/home/sean/github/operatefirst/workers/github_worker/github_worker.py", line 192, in issues_model
    pk_source_issues = self._get_pk_source_issues()
  File "/home/sean/github/operatefirst/workers/github_worker/github_worker.py", line 174, in _get_pk_source_issues
[...]

PR #1427, opened 2021-08-13
Patch Lack of API call checking in enrich_cntrb_id 

IsaacMilarky:
Previously, enrich_cntrb_id did not update the rate limit of the api key before it was hitting endpoints. I have restructured the api request to be pretty similar to to how it updates the rate limit in paginate_endpoint. Also there was a duplicate update_gh_rate_limit for some reason that was identical to the other one that has since been deleted. This pull request is a response to #1423.


sgoggins (approved):
LGTM!

PR #1442, opened 2021-08-19
update readme to refer to new primary branch as `main`

sgoggins:


sgoggins:
Fixed previous references to old primary branch `master`. Primary branch is now `main`.

PR #1508, opened 2021-10-29
Updates to contributor resolution

sgoggins:
Addressing a bug caused by Github user 'NaN', who is interpreted as a float instead of as a string without explicit type casting.


sgoggins (commented):
What I expected.

PR #1542, opened 2021-12-20
Documentation of enabling HTTPS for augur server

sbrennan98:
# Enabling HTTPS for an Augur Server

This change includes some files and documentation for enabling HTTPS to work with augur :)


sgoggins:
@sbrennan98 or @ABrain7710 or @IsaacMilarky : We need to test that this will run on a dev machine for a developer who runs the frontend using `npm run serve` ... I think it does, but would like some confirmation. 

@Ulincsys : We would also need to update `augur_view` to hit https endpoints. 

@ABrain7710 : Anything that can be scripted on install would be great.

augur/application.py
```diff
@@ -39,7 +39,9 @@ def __init__(self, given_config={}, disable_logs=False, offline_mode=False):
         self.gunicorn_options = {
             'bind': '%s:%s' % (self.config.get_value("Server", "host"), self.config.get_value("Server", "port")),
             'workers': int(self.config.get_value('Server', 'workers')),
-            'timeout': int(self.config.get_value('Server', 'timeout'))
```
sgoggins, line 42: @ABrain7710 : This is the section of the code with the merge conflict. @sbrennan98 has a useful comment when you look at the merge conflict, and I think its these conditions where the install script may need to be modified. If we could script generating a certbot certificate (I'm not sure we can, or that its worth the labor), then we may want to use config variables for HTTP or HTTPS.

sgoggins (changes requested):
I'm not requesting any changes from @sbrennan98 -- Thanks for the contribution!
I think the dev team just needs to sort out the different installation conditions. Of course, if @sbrennan98 wants to help, we will not refuse. :)

ayushtamra:
@sgoggins does this needs any changes? I saw this on the [projects top priorities list](https://github.com/chaoss/augur/projects/9#card-77193923)

PR #1548, opened 2022-01-05
Isaac commit resolution

IsaacMilarky:
The facade worker now uses the commit sha to attempt to resolve commit emails to contributors  before anything else. Also other smaller changes to log statements and minor refactors to make more readable.

### DIFF
--- a/workers/worker_git_integration.py
+++ b/workers/worker_git_integration.py
@@ -1001,9 +1001,12 @@ class WorkerGitInterfaceable(Worker):
                 self.oauths[0]['rate_limit'] -= 1
                 self.logger.info("Headers did not work, had to decrement")
                 time.sleep(30)
+
         self.logger.info(
             f"Updated rate limit, you have: {self.oauths[0]['rate_limit']} requests remaining."
         )
+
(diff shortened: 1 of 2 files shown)


sgoggins (approved):

workers/worker_git_integration.py
```diff
@@ -1001,9 +1001,12 @@ def update_gh_rate_limit(self, response, bad_credentials=False, temporarily_disa
                 self.oauths[0]['rate_limit'] -= 1
                 self.logger.info("Headers did not work, had to decrement")
                 time.sleep(30)
+
         self.logger.info(
             f"Updated rate limit, you have: {self.oauths[0]['rate_limit']} requests remaining."
         )
+
+        #Stalls after here for some reason.
```
sgoggins, line 1009: I have seen stalling on heavily taxed machines. Are you seeing this only in the facade worker resolution or elsewhere?

IsaacMilarky: I haven't seen it elsewhere yet and it hasn't happened with facade when I have been testing it most recently. I think that comment might be talking about a bug that has since been squashed.

PR #1559, opened 2022-01-12
Add print traceback method to pr worker

ABrain7710:
1. Add a method that will print the stack trace when an exception is registered instead of repeating this code block every time
`                self.logger.debug(f'exception: {e}.')
                stacker = traceback.format_exc()
                self.logger.debug(f"{stacker}")
`


sgoggins (approved):
LGTM. I'm going to run it in the test branch before release.

sgoggins (approved):
LGTM. Going to run it in test before release.

PR #1560, opened 2022-01-12
Remove pass from pr worker

ABrain7710:
1. Remove unneeded `pass` keywords from pr worker. 

The `pass` keyword is used as a placeholder for future code, so in these use cases they were not doing anything


sgoggins (approved):
LGTM. I'm going to run it in the test branch before release.

PR #1575, opened 2022-01-24
Small change to worker_base.py

ABrain7710:
1. Made a small change to an if else block which makes it more readable

The old version had two chains of if else blocks using python's syntax, which I found very difficult to read, so I changed it to 1 more traditional if else block

Essentially I took 2 if else chains using python syntax to 1 more traditional if else block

### DIFF
--- a/workers/worker_base.py
+++ b/workers/worker_base.py
@@ -384,14 +384,20 @@ class Worker(Persistant):
             'repo_id': repo_id,
             'job_model': model
         }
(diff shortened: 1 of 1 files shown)


IsaacMilarky (commented):
This seems better and more readable for sure. I don't know why there was some unnecessary redundancy before.

IsaacMilarky (approved):

sgoggins (approved):

PR #1608, opened 2022-02-02
CRITICAL: Tensorflow version casuing installation error

ABrain7710:
1. There was an error in the setup.py for the message insights worker that said `tensorflow==2.8.0rc0`, which needs to be `tensorflow==2.8.0`.

This typo was causing `make rebuild-dev` and `make install-dev` to both error when they made it to the message_insights_worker installation.

### DIFF
--- a/workers/message_insights_worker/setup.py
+++ b/workers/message_insights_worker/setup.py
@@ -37,7 +37,7 @@ setup(
         'emoji==1.2.0',
         'Keras>=2.8.0rc0',
         'Keras-Preprocessing==1.1.2',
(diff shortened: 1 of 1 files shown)


IsaacMilarky (approved):

sgoggins:
@ABrain7710 : Thanks!  Tensorflow dependencies have changed rapidly lately. Just a few days ago I had it building on all platforms. My guess is the release candidate went into release (obviously) and then pypi broke the dependencies. In fact I built it just yesterday on Ubuntu. THANK YOU!!

PR #1610, opened 2022-02-03
Visualization - Error Checking

ABrain7710:
1. Add method to get the repo_id, start_date, and end_date in the pr reports (this code was repeated for every single vis, so I made it into a method)
2. Add a message to the user when they hit an endpoint, but do not specify a repo_id to show data for

### DIFF
--- a/augur/routes/contributor_reports.py
+++ b/augur/routes/contributor_reports.py
@@ -576,8 +576,8 @@ def create_routes(server):
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
         if repo_id is None:
-            return Response(response="Invalid Query Params. Refer to the API docs here: https://oss-augur."
-                                     "readthedocs.io/en/main/rest-api/api.html",
+            return Response(response="repo_id not specified. Use this endpoint to get a list of available repos: "
+                                     "http://<your_host>/api/unstable/repos",
                             mimetype='application/json',
                             status=400)
 
@@ -747,8 +747,8 @@ def create_routes(server):
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
         if repo_id is None:
-            return Response(response="Invalid Query Params. Refer to the API docs here: https://oss-augur."
-                                     "readthedocs.io/en/main/rest-api/api.html",
+            return Response(response="repo_id not specified. Use this endpoint to get a list of available repos: "
+                                     "http://<your_host>/api/unstable/repos",
                             mimetype='application/json',
                             status=400)
 
@@ -951,8 +951,8 @@ def create_routes(server):
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
         if repo_id is None:
-            return Response(response="Invalid Query Params. Refer to the API docs here: https://oss-augur."
-                                     "readthedocs.io/en/main/rest-api/api.html",
+            return Response(response="repo_id not specified. Use this endpoint to get a list of available repos: "
+                                     "http://<your_host>/api/unstable/repos",
                             mimetype='application/json',
                             status=400)
 
@@ -1084,8 +1084,8 @@ def create_routes(server):
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
         if repo_id is None:
-            return Response(response="Invalid Query Params. Refer to the API docs here: https://oss-augur."
-                                     "readthedocs.io/en/main/rest-api/api.html",
+            return Response(response="repo_id not specified. Use this endpoint to get a list of available repos: "
+                                     "http://<your_host>/api/unstable/repos",
                             mimetype='application/json',
                             status=400)
 
--- a/augur/routes/contributor_reports.py
+++ b/augur/routes/contributor_reports.py
@@ -575,6 +575,12 @@ def create_routes(server):
 
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
+        if repo_id is None:
+            return Response(response="Invalid Query Params. Refer to the API docs here: https://oss-augur."
+                                     "readthedocs.io/en/main/rest-api/api.html",
+                            mimetype='application/json',
+                            status=200)
+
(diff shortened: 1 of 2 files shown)


augur/routes/contributor_reports.py
```diff
@@ -356,11 +356,14 @@ def get_repo_id_start_date_and_end_date():
...
 
-        repo_id = int(request.args.get('repo_id'))
+        repo_id = request.args.get('repo_id')
         start_date = str(request.args.get('start_date', "{}-01-01".format(now.year - 1)))
         end_date = str(request.args.get('end_date', "{}-{}-{}".format(now.year, now.month, now.day)))
 
-        return repo_id, start_date, end_date
+        if repo_id:
+            return int(repo_id), start_date, end_date
+
```
sgoggins, line 365: is the none, none, none API return set what generates the "Hey, you need a repo ID there pardner?" message.

augur/routes/contributor_reports.py
```diff
@@ -572,6 +575,12 @@ def new_contributors_bar():
 
         repo_id, start_date, end_date = get_repo_id_start_date_and_end_date()
 
+        if repo_id is None:
```
sgoggins, line 578: Seems to be "yes" to my prior question.

augur/routes/contributor_reports.py
```diff
@@ -931,10 +946,16 @@ def new_contributors_stacked_bar():
 
     @server.app.route('/{}/contributor_reports/returning_contributors_pie_chart/'.format(server.api_version),
                       methods=["GET"])
-    def returning_contributor_pie_chart():
```
sgoggins, line 934: is this simply a convention change, or is It fixing an error. LGTM, I just want to make sure I understand the change. 

Does it change the name of the existing endpoint? If so, @Ulincsys is relying on the current names and we probably need to also provide a legacy alias.

augur/routes/contributor_reports.py
```diff
@@ -1058,10 +1079,16 @@ def returning_contributor_pie_chart():
 
     @server.app.route('/{}/contributor_reports/returning_contributors_stacked_bar/'.format(server.api_version),
                       methods=["GET"])
-    def returning_contributor_stacked_bar():
+    def returning_contributors_stacked_bar():
```
sgoggins, line 1082: Same note as above.

augur/routes/pull_request_reports.py
```diff
@@ -421,14 +421,37 @@ def filter_data(df, needed_columns, not_null_columns=[]):
...
+        """ Gets the repo_id, start_date, and end_date from the GET requests array
+
+        :return: repo_id - id of the repo data is being retrieved for
+        :return: start_date - earliest time on visualization. Defaults to the January 1st of last year
+        :return: end_date - latest time on visualization. Defaults to current date
+        """
 
         now = datetime.datetime.now()
 
-        repo_id = int(request.args.get('repo_id'))
```
sgoggins, line 429: Same question about int casting. I'm just asking to make sure I understand. My guess is you have tested it, and it is not necessary. So, then no problem.

augur/routes/pull_request_reports.py
```diff
@@ -567,11 +590,14 @@ def average_commits_per_PR():
     @server.app.route('/{}/pull_request_reports/average_comments_per_PR/'.format(server.api_version), methods=["GET"])
     def average_comments_per_PR():
 
-        now = datetime.datetime.now()
```
sgoggins, line 570: What happens if somebody provides a start date, but not an end date? Is it defaulted to "now" elsewhere?

sgoggins (changes requested):
Not necessarily asking for changes. Except for one comment regarding whether or not we are changing the APIs to plural, everything else is to make sure I understand. I stopped commenting where they became repetitive in different endpoings. 

I do think we'll need to provide legacy API endpoints if the pluralization is changing them because of the augur_view interface, as well as a series of Jupyter Notebooks being developed in the CHAOSS Metrics Models working group. (Every other tuesday at 6pm. Next meeting is February 15.)

augur/routes/contributor_reports.py
```diff
@@ -356,11 +356,14 @@ def get_repo_id_start_date_and_end_date():
 
         now = datetime.datetime.now()
 
-        repo_id = int(request.args.get('repo_id'))
```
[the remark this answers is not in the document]
ABrain7710: If 'repo_id' is in the request.args array is None (which means it isn't specified in the query parameters), casting it to an int would result in an error because you cannot cast None to an int. Therefore I moved the cast to after I've checked if the 'repo_id' is specified in the request.args array

augur/routes/contributor_reports.py
ABrain7710: Yes, basically if the repo_id is not in the query params, then it returns None for all return variables. I could have just returned None, but I decided to still return None for the start_date and end_date too.

augur/routes/contributor_reports.py
ABrain7710: Yes

augur/routes/contributor_reports.py
ABrain7710: This is simply a change to the method name to match the API endpoint path. It will not change the existing endpoint.

augur/routes/contributor_reports.py
ABrain7710: Yep, it is just a change so the method name matches the current API endpoint path

augur/routes/pull_request_reports.py
ABrain7710: Yes, it is casted in the return statement, after I've confirmed that it isn't a None type, because trying to cast a None type to int will return an error, and give the user a 500 server error

augur/routes/pull_request_reports.py
ABrain7710: If someone provides a start date but no end date, the end date defaults to now. This made me think that I should add an error check to make sure the start_date is before the end_date though.

sgoggins (approved):

Issue #1633, opened 2022-02-26
Update CONTRIBUTING.md 


purna135:
1. In CONTRIBUTING.md, the link "[development installation instructions](https://oss-augur.readthedocs.io/en/master/development-guide/installation.html)" is broken.
2. There are a few additional broken connections as well.
   - [Stable documentation (master branch)](https://oss-augur.readthedocs.io/en/master/)
   - [Live Augur demo](http://zephyr.osshealth.io/)
3. Upstream is referring to the forked repo, which should be the original repo that we forked.
4. The contributing guide does not include instructions on how to run tests locally.

PR #1635, opened 2022-02-26
Update CONTRIBUTING.md

purna135:
Address #1633

upstream has been updated to refer to the original repo


mgalexander (commented):
I think the instruction as written is accurate. You don't want to adjust the line changed here because the goal of the process is to have your own fork of Augur that is pointed at the upstream and for your clone of git to know about the upstream. See the doc for `git remote` for more details, but it lets you directly pull from upstream into your workspace w/o having to do stuff on the github website.

purna135:
Thank you very much, @mgalexander. Previously, I was perplexed by the image below : ) 
Thank you so much for your clarification.
Please allow me to close this PR.

image

PR #1641, opened 2022-03-02
Fix the file path for production deployment

mscherer:
@sgoggins , I deployed as we discussed last week, but it seems to fail due to path issues.

### DIFF
--- a/augurface/vue.config.js
+++ b/augurface/vue.config.js
@@ -1,4 +1,4 @@
 // #SPDX-License-Identifier: MIT
 module.exports = {
-    publicPath: '/augurface/'
+    publicPath: '/'
 }
--- a/augurface/public/index.html
+++ b/augurface/public/index.html
@@ -7,7 +7,7 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>logowithoutletters.png">
     <link href="https://fonts.googleapis.com/css?family=Roboto&display=swap" rel="stylesheet"> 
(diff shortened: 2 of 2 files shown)


augurface/vue.config.js
```diff
@@ -1,4 +1,4 @@
 // #SPDX-License-Identifier: MIT
 module.exports = {
```
sgoggins, line 2: @mscherer : This change makes me wonder if you are running `augurface` with the intention of having it be Augur's primary frontend, or if you are using it to load repos (which is what it is for, and its not quite right/perfect yet.).  The main frontend is in the `frontend` directory. 

There is also another project, https://github.com/chaoss/augur_view that we are developing new frontend "stuff" in.

sgoggins (commented):
@mscherer : Just a question to clarify intentions.

sgoggins:
@mscherer : tracking identities .. you are the person who is installing for San Diego that I zoomed with? Sorry. :)

mscherer:
@sgoggins yep, that's me. Forgot my profile was private, so you can't see my orgs nor my name :/

And yes, I finally deployed the proxy we talked about, but index.html point to server.tld/augurface/foo.js, while we serve them directly from the root directory. I can work around using configuration magic, but figured this would be easier.

But if that's supposed to be consumed by more than just our instance, I guess I need to propose another fix.

sgoggins:
@mscherer : It would be better for Augur if we kept the `/augurface` path in the URL, simply because there's another interface served up at `/` .. Is that possible.

sgoggins:
@mscherer : I will merge this and we can see what happens.

PR #1643, opened 2022-03-11
Add Multiprocessing to Facade Commits Processing and Commit Contributor Resolution

IsaacMilarky:
Made the multithreaded option in Facade's config default to 1. By default facade will now be in multithreaded mode. 

Added parallelism to the analysis method in facade06analyze.py. Also added parallelism to the insert_facade_contributor logic in the contributor_interface.

Parallelism done by  splitting commits into mostly equal queues so each process starts with a workload and there is no overhead to pass into queue from the parent.
            
You *Could* put everything in an inter-process queue before running the children processes and then having them just request from the shared resource with simple blocking.
            
*However* that causes problems with the python implementation of ProcessQueue from multiprocessing since if you put large amounts of data in it at once it can cause stalling when you try to .join() from the parent and not to mention the overhead of adding the mutex
            
You could also then try to make the queue small and then have the parent pass commits into it when its not full but that adds lots of overhead that is slightly faster but still pretty bad.
            
So instead each process starts with ~ (1/processes) * len(missing_commits) units of commit data in the queue.

Facade compute time on salt repo without multiprocessing: Completed in 1:57:55
Facade compute time on salt repo with multiprocessing: Completed in 0:58:55

---------------

Also added: 

@IsaacMilarky : @sgoggins added his branch to this one. It includes: 
1. Version update to v0.25.4, as v0.25.3 was released as an Augurface patch earlier this week
2. Schema update to eliminate a rogue `varbit` datatype where it should be `varchar`
3. Updated the Facade version to v1.2.4 in the six places where a version is declared. That will make it more clear for logging and recording the version in the commits and contributors table
4. Markdown documents in the root directory that explain how to deploy augur as a service. The default is to run it with `--disable-housekeeper`, though that is modifiable
5. Update to the `augurface/package-lock.json
6. Inclusion of an augur visual overview in documentation.


sgoggins:
@IsaacMilarky : I added my branch to this one. It includes: 
1. Version update to v0.25.4, as v0.25.3 was released as an Augurface patch earlier this week
2. Schema update to eliminate a rogue `varbit` datatype where it should be `varchar`
3. Updated the Facade version to v1.2.4 in the six places where a version is declared. That will make it more clear for logging and recording the version in the commits and contributors table
4. Markdown documents in the root directory that explain how to deploy augur as a service. The default is to run it with `--disable-housekeeper`, though that is modifiable
5. Update to the `augurface/package-lock.json
6. Inclusion of an augur visual overview in documentation.

sgoggins (dismissed):

sgoggins (approved):

PR #1658, opened 2022-03-22
In rare cases, when running multiple workers, duplicate entries were created for some objects

sgoggins:
Specifically, we found 80 duplicate issues in a corpora of 2.8 million, 93 duplicates in a corpora of 3.9 million pull requests, and 7 repo_labor duplicates in a corpora of 7.8 million 

The workers have duplicate protection logic with a rare failure scenario that is prevented by the addition of the unique keys in this schema modification script.

### DIFF
--- a/docs/source/quick-start.rst
+++ b/docs/source/quick-start.rst
@@ -12,6 +12,7 @@ PostgreSQL Installation
 	sudo apt update
 	sudo apt upgrade
 	sudo apt install software-properties-common
+	sudo apt install python3-dev
 	sudo apt install postgresql postgresql-contrib postgresql-client
 	sudo apt install build-essential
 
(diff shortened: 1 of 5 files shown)


sgoggins:
This script has been run against all existing Augur instances.

ABrain7710:
The foreign key from the  pull_request_reviews table to the pull_requests table still has on delete restrict on it, which I believe needs to be on delete cascade

The foreign key from the  issue_labels table to the issues table still has on delete restrict on it, which I believe needs to be on delete cascade

sgoggins:
> The foreign key from the pull_request_reviews table to the pull_requests table still has on delete restrict on it, which I believe needs to be on delete cascade
> 
> The foreign key from the issue_labels table to the issues table still has on delete restrict on it, which I believe needs to be on delete cascade

I made this change by having two sequential, now identical database scripts. Its a little weird, but since I upgraded all the live databases with the original, this was the best way not to cause an upgrade path mess.

PR #1659, opened 2022-03-22
Add https support to augurface

mscherer:
The patch un-hardcode the use of http, that was found while working on augurface deployment for augur.osci.io.


sgoggins (approved):
Very helpful fix. A good example strategy for other places we need to make this update @Ulincsys / @ABrain7710  / @IsaacMilarky

PR #1695, opened 2022-04-05
Add a workflow for CI

mscherer:
The workflow try to build the docker image on PR, and push
a image on each commit, and stable version when there is a release

Signed-off-by: Michael Scherer <misc@redhat.com>

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -1,0 +1,48 @@
+name: Build Docker images
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  release:
+    types:
+      - published
+
+jobs:
+  build:
+    name: Build image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main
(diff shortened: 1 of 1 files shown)


mscherer:
It seems that the CI fail because the build fail (a issue on npm that I am not sure to be able to debug).

.github/workflows/build_docker.yml
```diff
@@ -0,0 +1,43 @@
+name: Build Docker images
```
sgoggins, line 1: @mscherer : The Docker build in the test is failing. Which may mean something or it may not. 
````
The command '/bin/sh -c npm run build' returned a non-zero code: 1
```

It looks like its being built with React? I know we previously were using NPM, setting the version to the latest of 12.22.x ..

sgoggins: @mscherer : Maybe I should've seen you already noticed that.

sgoggins: OK... I think I see what is happening.

sgoggins (approved):

PR #1704, opened 2022-04-07
Correctly tag the images

mscherer:
Seems I messed up the tagging because I didn't test fully. I would still need to do some test before it should be merged (eg, tag a release on my fork and see how it goes, etc). 

Once that's done, there is also a setting to offer packages that need to be turned on, so people can use the docker image to deploy.


sgoggins (approved):

sgoggins:
@mscherer : Do you know why the docker images are private? Is there a setting to change on the repo, or to the script to make it public?

mscherer:
Yes, I think there is. I did it once, but I think that was the doc I used https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility

PR #1733, opened 2022-04-12
Message Insights Worker Updates

sgoggins:
# Pull Request Template
- Message insight worker updated to use current working libraries. 
- worker_git_integration.py updated to use `timeout` parameter when making a request. Resolves occasional stalling issues.


sgoggins (commented):
LGTM

PR #1762, opened 2022-04-16
Docker.yml:Added quotes to port mapping for docker

Biki-das:
# Pull Request Template

Added quotes to the docker port mapping as the docker docs suggest to use it-
https://docs.docker.com/compose/compose-file/compose-file-v3/#ports

### DIFF
--- a/database-compose.yml
+++ b/database-compose.yml
@@ -8,4 +8,4 @@ services:
       dockerfile: ./util/docker/database/Dockerfile
     restart: unless-stopped
     ports:
-     - 5434:5432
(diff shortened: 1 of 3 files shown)


sgoggins:
@Biki-das : This broke our docker build.

Biki-das:
> @Biki-das : This broke our docker build.

Hmm.. Strange, :-(

Biki-das:
@sgoggins idk if this is breaking the docker build, my last two pr also showed a docker build failure, which were just mere typo fix and documentation changes

PR #1770, opened 2022-04-18
Fix:Removed multiple imports of the object

Biki-das:
Signed-off-by: Biki-das <bikid475@gmail.com>

# Pull Request Template

Some worker files were pulling multiple imports of the same object, this can substantially slow down the files. I have removed the multiple imports for the same

### DIFF
--- a/workers/discourse_analysis_worker/discourse_analysis_worker.py
+++ b/workers/discourse_analysis_worker/discourse_analysis_worker.py
@@ -6,7 +6,6 @@ import pandas as pd
 import sqlalchemy as s
 from workers.worker_base import Worker
 
-import pandas as pd
 from sklearn.model_selection import train_test_split
 from sklearn.feature_extraction.text import CountVectorizer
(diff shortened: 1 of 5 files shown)


workers/discourse_analysis_worker/discourse_analysis_worker.py
```diff
@@ -16,7 +15,6 @@
 from sklearn_crfsuite import scorers
 from sklearn_crfsuite import metrics
 from sklearn.metrics import make_scorer
-from sklearn.model_selection import cross_val_score
```
sgoggins, line 19: In this worker we are importing specific components of the libraries.

sgoggins (changes requested):

workers/discourse_analysis_worker/discourse_analysis_worker.py
Biki-das: ok will reverse the change

PR #1771, opened 2022-04-18
Fix:Removed duplicate dictionary keys

Biki-das:
Signed-off-by: Biki-das <bikid475@gmail.com>

# Pull Request Template

Dictionary expression binds the same key multiple times. This may lead to unintentional behaviour as the key will always map the last value provided to it. It is recommended to have unique keys.


workers/insight_worker/insight_worker.py
```diff
@@ -676,8 +676,10 @@ def update_metrics(self):
                 "cm_defined": True if metric['is_defined'] == 'true' else False,
                 "cm_api_endpoint_repo": metric['endpoint'],
                 "cm_api_endpoint_rg": None,
+                // cm info key is being used twice
```
Biki-das, line 679: @sgoggins i am confused here the cm_info key is being used multiple times to two different values, what changes would be appropriate here

sgoggins (approved):

PR #1813, opened 2022-04-24
Added literal syntax to create the data structure

Biki-das:
# Pull Request Template

Using the literal syntax can give minor performance bumps compared to using function calls to create dict, list and tuple.

This is because here, the name dict must be looked up in the global scope in case it has been rebound. Same goes for the other two types list() and tuple().

### DIFF
--- a/workers/deps_libyear_worker/libyear_utils.py
+++ b/workers/deps_libyear_worker/libyear_utils.py
@@ -33,7 +33,7 @@ def find(name, path):
 def get_parsed_deps(path):
 
     deps_file = None
-    dependency_list = list()
+    dependency_list = []
 
     for f in file_list:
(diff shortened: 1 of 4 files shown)


sgoggins:
@Dhruv-Sachdev1313 : Could you have a look at this? _

Dhruv-Sachdev1313 (commented):
I don't think it would affect the working of the dependency worker, just feels like a more pythonic way of doing stuff! So LGTM.

Biki-das:
> I don't think it would affect the working of the dependency worker, just feels like a more pythonic way of doing stuff! So LGTM.
> 

Yup it's a small minor improvement!

sgoggins (approved):

PR #1855, opened 2022-06-08
Fixing Schema Script and Addressing GH API Timeout. 

sgoggins:
# Pull Request Template

### DIFF
--- a/schema/generate/100-schema_update_102.sql
+++ b/schema/generate/100-schema_update_102.sql
@@ -218,8 +218,8 @@ CREATE MATERIALIZED VIEW "augur_data"."explorer_entry_list"
 AS
 SELECT DISTINCT r.repo_git,
     rg.rg_name
-   FROM (repo r
-     JOIN repo_groups rg ON ((rg.repo_group_id = r.repo_group_id)))
+   FROM (augur_data.repo r
+     JOIN augur_data.repo_groups rg ON ((rg.repo_group_id = r.repo_group_id)))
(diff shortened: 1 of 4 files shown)


IsaacMilarky (approved):

ABrain7710 (changes requested):
The changes in worker_git_integration.py will not fix anything related to handing out new keys to workers. It will simply not make the first key used by a worker random, whereas before the changes the first key used was the key in the config so we don't use up all stored keys first. I would not recommend doing this, because we should use the key defined in the config first.

If there is an issue with handing out new keys to workers, the update_gh_rate_limit method in worker_git_integration.py would need to be looked at.

sgoggins:
@ABrain7710 : The key in the config is no greater or lesser than the other keys. The issue right now is that in practice it is not rotating through the available keys.

ABrain7710:
I understand they have the same number of requests, but we should use the key in the config first, because more people are likely to be using the stored keys at the same time, so we don't want to run those out first, we would rather run the one in the config first since it isn't shared.

sgoggins:
@ABrain7710 : If there's only the one in the config, its going to generate a random number of 1, and only hit that key. Everything still works if there's only the API key in the config. In practice the ones in the table are isolated to the organization running the instance. So all the keys have the same "value".

sgoggins:
@ABrain7710 : Right now, its simply not getting to all the keys before it goes kaput. In practice.

PR #1906, opened 2022-07-06
Schema, Dependency Evaluation, and Machine Learning Updates. 

sgoggins:
**Description**
- Schema, Dependency Evaluation, and Machine Learning Updates.


sgoggins:
Docker build works except for Frontend Docker, which we don't want to deploy as a package right now (because its not working). @mscherer : Thoughts?

PR #1912, opened 2022-07-08
Add user subcommand CLI

Priya730:
Signed-off-by: Priya Srivastava <shivikapriya730@gmail.com>
_This is work in progress_
**Description**
- Progress till now: Add user <username> subcommand with admin flag (more work (access, permissions) on admin role required)
- [x] Next steps that I'm working on: password hashing
- [x] email to login
- [x] Important: changes need to be made in multicommad.py for command to implement

This PR fixes #

**Notes for Reviewers**
Made this PR to keep track of the progress
**Signed commits**
- [x] Yes, I signed my commits.

PR #1968, opened 2022-09-04
augur user add cli commands fixed

Priya730:
Signed-off-by: Priya Srivastava <shivikapriya730@gmail.com>

**Description**
- The changes were required as the current work does not import User model, engine sessions and admin flag does not match with the current schema. Command now also requires firstname and lastname arguments.

image

NOTE: **admin attribute not present in the current schema**
This PR fixes:
1. Imported engine and User., session fixed.
2. Changed echo message

**Notes for Reviewers**
This PR is continues the changes made by #1912

**Signed commits**


Priya730:
@Ulincsys please review this PR

Priya730:
Also, can you tell me how to integrate the admin flag to the current schema

sgoggins:
@Ulincsys : Can you review this?

Ulincsys (changes requested):
I think that the add user command will fail without adding the admin argument in the call to construct the `User` object (since that column has a NOT_NULL constraint). I see that the command line option for this was commented out.

Priya730:
> I think that the add user command will fail without adding the admin argument in the call to construct the `User` object (since that column has a NOT_NULL constraint). I see that the command line option for this was commented out.

Yes I realised that and was doing the same changes right now

Ulincsys (approved):
Looks good!

PR #1982, opened 2022-09-09
Fix for null characters in strings of issue data

ABrain7710:
1. Improve the null character replacement, so it works with the new string that was encountered
2. Split into functionality into 3 different functions. One that removes nulls from a string. One that removes nulls from strings in a dict. One that removes nulls from strings in a list of dicts
3. Add a unit test that tests  the `insert_data` function's ability to handle null data in strings


sgoggins (approved):
LGTM

PR #1999, opened 2022-10-05
Machine Learning Workers Working

sgoggins:
On OSX (and elsewhere) the max python version supported is 3.9. To create a venv on OSX at that python release level, `python3.9 -m venv /path/to/venv`

Check to make sure facade is working as well.


sgoggins (commented):
This does look good to me @ABrain7710  / @IsaacMilarky

PR #2001, opened 2022-10-06
Update Augur New with Development Work

sgoggins:
**Description**
- Hardened `augur-new-dev`
- Added machine learning workers
- Updated Alembic
- Several bug fixes
- Login and REPO Creation API's


augur/api/config.py
```diff
@@ -37,6 +37,8 @@
 
 from augur.application.logs import ROOT_AUGUR_DIRECTORY
 
+loggy_var=ROOT_AUGUR_DIRECTORY+'/logs'
```
ABrain7710, line 40: @IsaacMilarky I'm assuming this was your change. I think we may need a `/` at the end of the logs dir. I'd at least like to check and make sure before this is merged

ABrain7710, line 40: @IsaacMilarky I'm assuming this is your change. I think the logs dir may need a `/` at the end. I'd at least like to check before this is merged.

augur/application/cli/backend.py
```diff
@@ -38,8 +39,15 @@ def cli():
 def start(disable_collection):
     """Start Augur's backend server."""
 
-    raise_open_file_limit(100000)
+    try:
+        raise_open_file_limit(100000)
+    except Exception as e:
```
ABrain7710, line 44: @sgoggins or @IsaacMilarky: Can you catch the specific exception instead of the base Exception class?

augur/api/config.py
sgoggins: @IsaacMilarky / @ABrain7710 : I confirmed there is a trailing slash when its stored in the database. That is working, so ... probably keep it.

sgoggins: Confirmed. Trailing slash is stored as of today.

augur/application/schema/alembic/versions/2022-09-28_12_add_user_repo_table.py
```diff
@@ -0,0 +1,54 @@
+"""Add user repo table
+
+Revision ID: 12
```
ABrain7710, line 3: @ABrain7710 I need to change the revision id on this file

augur/tasks/github/facade_github/tasks.py
```diff
@@ -177,15 +177,19 @@ def process_commit_metadata(contributorQueue,repo_id):
             
             #Executes an upsert with sqlalchemy 
             cntrb_natural_keys = ['cntrb_login']
-            session.insert_data(cntrb,Contributor,cntrb_natural_keys)
+            try:
+                session.insert_data(cntrb,Contributor,cntrb_natural_keys)
+            except Exception as e:
```
ABrain7710, line 182: @IsaacMilarky Can we catch the specific exception here?

augur/tasks/github/facade_github/tasks.py
```diff
@@ -177,15 +177,19 @@ def process_commit_metadata(contributorQueue,repo_id):
             
             #Executes an upsert with sqlalchemy 
             cntrb_natural_keys = ['cntrb_login']
```
ABrain7710, line 179: This cntrb natural keys needs to be `cntrb_id`

augur/tasks/github/issues/tasks.py
```diff
@@ -129,19 +130,20 @@ def process_issues(issues, task_name, repo_id, logger) -> None:
...
         # specified in issue_return_columns is the columns of data we want returned. This data will return in this form; {"issue_url": url, "issue_id": id}
         logger.info(f"{task_name}: Inserting {len(issue_dicts)} issues")
-        issue_natural_keys = ["repo_id", "gh_issue_id"]
+        issue_natural_keys = ["issue_url"]
         issue_return_columns = ["issue_url", "issue_id"]
         issue_string_columns = ["issue_title", "issue_body"]
-        issue_return_data = session.insert_data(issue_dicts, Issue, issue_natural_keys, return_columns=issue_return_columns, string_fields=issue_string_columns)
-
-
+        try:
```
ABrain7710, line 143: @sgoggins I would like to change this to throw an exception when in development but not throw an exception during production. We can use @Ulincsys dev environment variable.  The only other option than that is to throw an exception in production and have people open issues when they get exceptions, but I don't think we want that.

augur/tasks/github/issues/tasks.py
```diff
@@ -129,19 +130,20 @@ def process_issues(issues, task_name, repo_id, logger) -> None:
...
-        session.insert_data(contributors, Contributor, ["cntrb_login"])
+        session.insert_data(contributors, Contributor, ["cntrb_id"])
                             
 
         # insert the issues into the issues table. 
         # issue_urls are gloablly unique across github so we are using it to determine whether an issue we collected is already in the table
         # specified in issue_return_columns is the columns of data we want returned. This data will return in this form; {"issue_url": url, "issue_id": id}
         logger.info(f"{task_name}: Inserting {len(issue_dicts)} issues")
-        issue_natural_keys = ["repo_id", "gh_issue_id"]
+        issue_natural_keys = ["issue_url"]
```
ABrain7710, line 140: @sgoggins / @IsaacMilarky This should be `["repo_id", "gh_issue_id"]` not `["issue_url"]`. If a repo moves, the issue URL will change and we will get duplicate `gh_issue_ids` since the issues id stays the same, but the URL changes.

augur/tasks/github/pull_requests/tasks.py
```diff
@@ -85,14 +85,14 @@ def process_pull_requests(pull_requests, task_name, repo_id, logger):
...
-        session.insert_data(contributors, Contributor, ["cntrb_login"])
+        session.insert_data(contributors, Contributor, ["cntrb_id"])
 
 
         # insert the prs into the pull_requests table. 
         # pr_urls are gloablly unique across github so we are using it to determine whether a pull_request we collected is already in the table
         # specified in pr_return_columns is the columns of data we want returned. This data will return in this form; {"pr_url": url, "pull_request_id": id}
         logger.info(f"{task_name}: Inserting prs of length: {len(pr_dicts)}")
-        pr_natural_keys = ["repo_id", "pr_src_id"]
+        pr_natural_keys = ["pr_url"]
```
ABrain7710, line 95: @sgoggins / @IsaacMilarky This should be ["repo_id", "pr_src_id"] not ["pr_url"]. If a repo moves, the pr URL will change and we will get duplicate pr_src_ids since the pr src id stays the same, but the URL changes.

augur/application/cli/backend.py
sgoggins: @ABrain7710 : I don't know what the specific exception was ... @IsaacMilarky : Do you recall?

augur/application/schema/alembic/versions/2022-09-28_12_add_user_repo_table.py
sgoggins: @ABrain7710 : I fixed this.

augur/tasks/data_analysis/contributor_breadth_worker/runtime.py
```diff
@@ -0,0 +1,24 @@
+#SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
sgoggins: i.e., we don't need the runtime.py?

augur/tasks/github/facade_github/tasks.py
sgoggins: @IsaacMilarky : What @ABrain7710 says here makes sense to me, and I am also wondering if there is a reason we need this to be the natural key. When we first get commit users we don't have the cntrb_login, so I think the cntrb_id is the "unique thing" ... let me know.

augur/tasks/github/issues/tasks.py
sgoggins: @IsaacMilarky : What @ABrain7710 is saying here makes sense to me.

augur/tasks/github/issues/tasks.py
sgoggins: @ABrain7710 : If we do not throw an exception does it simply not update the issue and move ahead? I will trust your judgement on this one. Do what you think should be done.

augur/tasks/github/pull_requests/tasks.py
sgoggins: I agree. I thought we changed this one already. ?

sgoggins (commented):
@ABrain7710 : I made comments. There's some stuff you can go ahead and do. There's other stuff I'm waiting for @IsaacMilarky on.

schema-updates.sql
```diff
@@ -0,0 +1,57 @@
+-- DO NOT IMPLEMENT YET .. trying to figure out why this happened. Running again.
```
ABrain7710, line 1: @sgoggins Is this supposed to be merged in?

ABrain7710 (changes requested):
There are just a few things I saw and commented on. I'll try to fix as many as I can, and then I'll look over everything again

sgoggins:
@ABrain7710 : I think you can do everything we aren't waiting to hear from @IsaacMilarky about.

schema-updates.sql
sgoggins: @ABrain7710 : These are changes I made to correspond with changes in the alembic script .. or to signal necessary changes to the alembic script. I am not sure how to write those yet.

augur/tasks/data_analysis/contributor_breadth_worker/runtime.py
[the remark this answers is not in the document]
ABrain7710: Exactly, I just removed it.

augur/api/config.py
ABrain7710: @IsaacMilarky Actually this is on the old, `config.py` file. I'm not sure why this was changed. I've simply kept this file around for reference.

augur/tasks/github/issues/tasks.py
ABrain7710: @sgoggins Yes, if we catch the exception, it will fail to insert the row, but the process will keep running. So, the issue assignees and labels will not be inserted, and the task will finish. If we do this for production instances and then raise the errors in development so we can fix them, it causes two things. First is a lot less user testing of the code because the errors are being squashed, and they are never notified of it, which means we are not notified. Second, we have to test a lot more because we won't get as many issues from people.

I'm not saying I have a problem with this approach. This is actually the approach I like the most. I'm just stating that this requires us to do a lot more large-scale testing to find the weird bugs.

ABrain7710: @sgoggins This is what I'm proposing.
```
        try:
            issue_return_data = session.insert_data(issue_dicts, Issue, issue_natural_keys, return_columns=issue_return_columns, string_fields=issue_string_columns)
        except IntegrityError as e:
            logger.error(f"Ran into integrity error:{e} \n Offending data: \n{issue_dicts}")

            if development:
                raise e
```

sgoggins:
@IsaacMilarky : Can you review this?

IsaacMilarky (changes requested):
Needs changes to various things including error handling and docker specific configuration.

IsaacMilarky (approved):

IsaacMilarky (approved):

ABrain7710 (approved):

PR #2012, opened 2022-10-20
Miscellaneous Fixes to Augur New

ABrain7710:
**Description**
1. Changed the cli options for `augur config init` to use dashes in the option names rather than underscores
2. Added a logger to the `RandomKeyAuth` class, so it can log when there are no valid keys. Before it was erroring because someone tried to log this statement using `self.error()`
3. Fixes the regex expressions in the repo load controller to allow for `.github` and `.github.io` repos

**This PR fixes**
1. Error when deleting users. The API to delete a user now deletes all the users repos in the `user_repos` table before deleting the user from the `users` table.
2. Fixed the `user/update` API. Now it returns `missing argument` if either `user` or `password` is `None`, rather than only when `user` is `None`. It also checks to make sure an email or username isn't already taken when updating them. It also returns `missing argument` now if only the username and password are returned. Lastly, I removed an unneeded else since the API returned inside of the if clause. 

**Signed commits**

### DIFF
--- a/augur/api/routes/user.py
+++ b/augur/api/routes/user.py
@@ -16,7 +16,7 @@ from augur.tasks.github.util.github_task_session import GithubTaskSession
 from augur.util.repo_load_controller import RepoLoadController
 
 
-from augur.application.db.models import User
+from augur.application.db.models import User, UserRepo
 from augur.application.config import get_development_flag
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


IsaacMilarky (approved):

sgoggins (approved):
LGTM.

PR #2019, opened 2022-10-28
Overhaul Facade and add needed functionality to PR collection. 

IsaacMilarky:
-Completely replace the old facade database lib functions with the newer sqlalchemy ORM logic to match the rest of Augur. 
-Migrate the old pull_request_files_model and pull_request_commits_model to new augur
-Fix and make the graphql pagination logic more robust. 

I haven't seen too many problems in my own testing except for some internet issues out of my control.


sgoggins (approved):
@IsaacMilarky : LGTM. I do suggest that I should probably run a test with `augur-new` as it is, and `augur-new-dev` as it is using the same repos to make sure the commits are aligned and consistent. I'm running `augur-new-dev` under the `isaac-debug` directory on my user on the server right now. I will start `augur-new` against those same repos as well.

ABrain7710 (approved):

PR #2107, opened 2023-01-01
Session typo fix

sgoggins:
**Description**

This PR fixes #

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/github/util/github_paginator.py
+++ b/augur/tasks/github/util/github_paginator.py
@@ -7,7 +7,7 @@ import json
 import asyncio
 import datetime
 import logging
-
+from sqlalchemy.sql import text
 
 from typing import List, Optional, Union, Generator, Tuple
 from urllib.parse import urlencode, urlparse, parse_qs, urlunparse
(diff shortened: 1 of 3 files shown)


augur/tasks/github/util/github_paginator.py
```diff
@@ -30,7 +66,14 @@ def hit_api(key_manager, url: str, logger: logging.Logger, timeout: float = 10,
 
         try:
             response = client.request(
-                method=method, url=url, auth=key_manager, timeout=timeout, follow_redirects=True)
```
sgoggins, line 33: @ABrain7710 : don't we want to follow redirects?

augur/tasks/github/util/github_paginator.py
```diff
@@ -7,16 +7,52 @@
...
+                                        DELETE FROM augur_data.pull_request_message_ref where repo_id={repo_id};
+                                        DELETE FROM augur_operations.user_repos where repo_id={repo_id};
+                                        DELETE FROM augur_data.pull_request_labels where repo_id={repo_id};
+                                        DELETE FROM repo where repo_id={repo_id};
+                                """)
+            session.execute_sql(delete_repo_query)
+            return 
+
+        # update the old url to the new url 
+        query = text("""SELECT * FROM augur_data.repo WHERE repo_git='old_url' FOR UPDATE;
```
sgoggins, line 51: @ABrain7710 : I think we also need to set repo_status='New', repo_path=NULL, and repo_name=NULL here.

sgoggins: Presuming we are deleting everything, which is what it looks like .. then we'd be needing to also reset the repo in this way.

PR #2114, opened 2023-01-11
RabbitMQ Broker Changes and Scaling Improvements 

sgoggins:


IsaacMilarky:
Major Changes:
    - Changed Celery broker from Redis to RabbitMQ. Redis is still used as a result backend and key store for git api keys
    - Change structure of how celery tasks are scheduled so that the number of tasks scheduled aren't directly proportional to the amount of repos to collect
    - Fixed a query in facade to reflect changes in postgres type casting
    - Change in facade logs to not spam the logs too much. Now only logs what is unexpected along with quarterly progress updates
    - Change docs to reflect aforementioned change to RabbitMQ
    - Change docs to reflect improved docker setup and removal of old docker-setup.sh
    - Changed import of machine learning libraries to only happen when the ml phase launches

IsaacMilarky (approved):

PR #2119, opened 2023-01-12
Hardcode bash

mscherer:
**Description**

It seems that for a reason I can't understand (and to be fair, I am not that motivated to go deeply in that rabbit hole), the shebang do not work on podman. Removing /usr/bin/env make it work fine, and since there is almost no reason to see bash move anywhere out of /usr, I think that's safe to just remove env and directly use bash.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docker/backend/entrypoint.sh
+++ b/docker/backend/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #SPDX-License-Identifier: MIT
 set -e
 


sgoggins:
@IsaacMilarky : Can you take a look at this one and let me know if you need an EC2 to experiment.

sgoggins:
@IsaacMilarky : Can you look at this one as well?

IsaacMilarky (approved):

PR #2130, opened 2023-01-19
Updates to RabbitMQ Model, additional tasks

sgoggins:
**Description**
- Updated `rabbitmq-server` code
- Added dependency tasks from prior version of Augur

**Signed commits**


IsaacMilarky:
Significant Changes:
   -Reduced message size greatly as to not overwhelm rabbitmq broker. This is done by:
        - Bundling all tasks to take lists of repo identifiers instead of creating many tasks for every repo
        - Changes to facade to grab commit information inside the task instead of as arguments to the task call
    - Added dependency model from older version of Augur
    - Made Facade not log as much and spam file I/O

PR #2133, opened 2023-01-21
DatabaseEngine class for better handling of db engines

ABrain7710:
**Description**
* Defined a DatabaseEngine class to better handle the creation and clean up of database engines. 
* The class uses a context manager to ensure that the engine is disposed of correctly. 
* It also takes a connection_pool_size so processes that need smaller or larger connection pool can configure that on a per case basis. 

**Notes for Reviewers**
1. The changes to the api and metrics files are to make sure the engine only has one connection, since the gunicorn process will only ever need one connection. 
2. I also defined the short lived database engines in the machine learning workers as a single connection. 

**Signed commits**


sgoggins (commented):
@ABrain7710 : I'm going to merge `dev` with @IsaacMilarky 's changes into your branch first, and test your branch at scale that way, before merging this one.

sgoggins (commented):
@ABrain7710 : After opening a PR I see this one starts with @IsaacMilarky 's changes already in it, so I'm going to run this branch.

augur/api/metrics/contributor.py
```diff
@@ -43,7 +44,7 @@ def contributors(repo_group_id, repo_id=None, period='day', begin_date=None, end
 
     if repo_id:
         contributorsSQL = s.sql.text("""
-           SELECT id::text                         AS user_id,
```
sgoggins, line 46: If I was casting these are text its probably because the frontend relies on it... checking.

ABrain7710:
@sgoggins I made some changes to the changes you made in commit `Strings with size zero throw errors`. The issue wasn't that the strings with size zero. The issue was with the if statements before hand. The if statement that checked if the key `name` exists did not check if the value wasn't None, while the if statement for the email did check for this. So I changed the code to use the dictionary method `get()` and `or` statements to fix the issue and make it more readable

augur/api/metrics/contributor.py
sgoggins: I confirmed that we get an issue if we don't cast the result as text: 
```
2023-01-21 16:34:20 linda augur.api.server[200488] ERROR Exception on /api/unstable/repos/61187/contributors [GET]
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/home/sean/github/rh-augur-new-dev/augur/api/server.py", line 361, in endpoint_function
    data =  self.transform(func, args, kwargs)
  File "/home/sean/github/rh-augur-new-dev/augur/api/server.py", line 300, in transform
    result = data.to_json(orient=orient, date_format='iso', date_unit='ms')
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/pandas/core/generic.py", line 2558, in to_json
    return json.to_json(
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/pandas/io/json/_json.py", line 113, in to_json
    s = writer(
  File "/home/sean/github/virtualenv/k12/lib/python3.8/site-packages/pandas/io/json/_json.py", line 175, in write
    return dumps(
OverflowError: Unsupported UTF-8 sequence length when encoding string
```

sgoggins (changes requested):
Change the query back to do string casting on the UUID in `augur/api/metrics/contributor.py`

Otherwise everything seems to run good!

augur/api/metrics/contributor.py
ABrain7710: Weird, I don't know anything about that. I'm assuming that this branch is just old and the old changes wrote over the dev changes. I can make that change really quick though

sgoggins (approved):

PR #2159, opened 2023-01-28
New method for repo collection

ABrain7710:
**Description**
- Add `collection_status` table that keeps track of the state of a repo's collection. The table consists of the fields repo_id, the time the last collection was complete, the time the last event was collected, the current status, and the task id.
- Defined a task that runs periodically (how often it runs is configurable) and checks to see if there are any repos that need collected. @IsaacMilarky and I agreed that 5-10 minutes is generally ideal to make sure the new repos are getting collected soon after they are added. The task also manages the max number of repos allowed to be collected at once. Currently the `max_repo_count` is hardcoded at 500 but in the future it will be configurable. This was done to make sure that Celery isn't getting overrun with too many jobs, while still making sure there are plenty of repos in the queue ready to be collected. It also will collect repos that haven't been collected in awhile. Currently if a repo hasn't been collected in 30 days it will be collected again (this is currently hardcoded but will be configurable in the future). 

Flow of collection:
1. New repos get added from the cli or the frontend
2. Whenever the periodic collect task runs it will start collecting the new repos (as long as there aren't too many repos already being collected)
3. This loop then continues forever

Note: When calling `augur backend start` the number of repos added to the queue will be the `max_repo_count` or less if there are less repos. Currently if a repo errors its repo will not be collected again until the developer sets the status to `Pending`.

Additional Note: The `dev` branch has already been merged into this branch so if you would like to test it before merging you can just test on `repo-collection-message-scaling`

**Signed commits**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,13 @@
+
 # Augur NEW Release v0.44.0
+
 [![first-timers-only](https://img.shields.io/badge/first--timers--only-friendly-blue.svg?style=flat-square)](https://www.firsttimersonly.com/) We follow the [First Timers Only](https://www.firsttimersonly.com/) philosophy of tagging issues for first timers only, and walking one newcomer through the resolution process weekly. [You can find these issues tagged with "first timers only" on our issues list.](https://github.com/chaoss/augur/labels/first-timers-only).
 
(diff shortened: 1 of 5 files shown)


augur/application/cli/backend.py
```diff
@@ -93,7 +94,19 @@ def start(disable_collection, development, port):
         cpu_worker_process = subprocess.Popen(cpu_worker.split(" "))
         time.sleep(5)
 
-        start_task.si().apply_async()
+        with DatabaseSession(logger) as session:
```
sgoggins, line 97: This could possibly be the source of the connection bloating I am seeing at runtime, including high volumes of rollback transactions.

sgoggins (changes requested):
Some kind of change to the management of database connections exists in this branch, and is causing a high volume of ROLLBACKs in the database, which ultimately consume all available connections. So, this is definitely not ready to merge.

sgoggins:
@ABrain7710 : What are your thoughts on the status of this PR?

PR #2173, opened 2023-02-14
Deprecating Facade Affiliations Processing

sgoggins:
**Description**
- The Facade task, which is used for commit counting in Augur, previously had two functions for "nuking" and "rebuilding" affiliation data inside the commit table. This is a feature no longer required, as it can be replaced with simpler logic, queries, and materialized views if needed. These were also computationally expensive tasks that added limited value. 
- Facade task version bumped to 1.3.0


sgoggins:
@ABrain7710 / @IsaacMilarky : I also just merged @ABrain7710 's database connection patch. I am testing with 8knot big data.

IsaacMilarky (approved):

PR #2177, opened 2023-02-14
Fix Machine Learning Db connection leak

ABrain7710:
**Description**
- Fixed machine learning connection leak, and reduced the number of times we are opening sessions so there is a lower chance for one in the future

**Signed commits**


sgoggins (approved):
LGTM

PR #2182, opened 2023-02-17
Fix adding repos and orgs

ABrain7710:
**Description**
- An error occurs when trying to add repos or orgs on the frontend. It was due to a function using a class that did not have access to the github api keys to validate the repos and orgs that were being added

This PR fixes #
- Adding repos and orgs bug

**Signed commits**


sgoggins (approved):
LGTM

IsaacMilarky (approved):

PR #2184, opened 2023-02-21
Facade collection update

IsaacMilarky:
**Description**
- Remove repo_status from repo table and add it to collection_status table
- Make facade into a separate collection hook so that commit updates can be queried and jobs can be queued based on detected repo updates. 
- Update facade config to remove unneeded options like 'force updates' and 'force analysis'. Now options for updates and analysis just have a simple toggle. 
- Change facade queries to reflect schema change
- Change repo status to only be used to get status of whether repo was cloned or not. If it wasn't cloned it has a status of 'new', if it was cloned it has a status of 'Update.' Other status's made sense when we had a worker-based structure to collection but doesn't make sense for a message queue. This previous function will be done as part of the celery scheduling. 

**Signed commits**


sgoggins:
@IsaacMilarky : Does this make sense for me to test first?
Also, I think some docs about how the jobs are managed in facade would be helpful.

ABrain7710 (changes requested):
I'm a bit confused with some of the changes. I thought we were going to have only the `facade_status` since the `repo_status` is only relating to Facade so it isn't really a `repo_status`. I'm now just confused as to why we need both. Couldn't we just have a `facade_status` that has values of ("Pending", "Update", "Failed Clone", "Success", "Errored")? I thought this is what we discussed the other day.

IsaacMilarky:
> I'm a bit confused with some of the changes. I thought we were going to have only the `facade_status` since the `repo_status` is only relating to Facade so it isn't really a `repo_status`. I'm now just confused as to why we need both. Couldn't we just have a `facade_status` that has values of ("Pending", "Update", "Failed Clone", "Success", "Errored")? I thought this is what we came to the other day.

You're right I'll make those changes. I think I just had forgotten that when I was making the clean_collection_status method since that resets the collection status of interrupted tasks to pending and from that I forgot to remove the repo_status field since that keeps track of the status of the git repo

ABrain7710 (approved):
I made a few small changes, then tested it and it appears to be working as expected. I will note that with these changes it is important to set the `collection_interval` to a smaller value since the secondary data won't be collected until the primary data finishes, and the next `augur_collection_monitor` task runs again.

ABrain7710 (approved):

PR #2186, opened 2023-02-22
Facade repo path changes

IsaacMilarky:
**Description**
- Facade no longer detects collision based on the repo table's repo_path and repo_name fields and instead looks inside the facade directory to detect collision. If collision is detected the repo is set to 'Update' and reset
- Dependency model no longer relies on repo_path and repo_name fields and instead uses repo_git
- Fix detection of repositories being moved to work with the new collection hooks. The prelim phase is still enabled/disabled through the same means but now has specific celery tasks defined for each collection hook. This is necessary because this task has to manage a special condition of collection and needs access to affect the CollectionStatus on the proper fields relating to the corresponding collection hook. 

**Signed commits**


ABrain7710 (changes requested):
If I am understanding these changes correctly, it looks like Facade is still setting the repo_name and repo_path, it just doesn't depend on those fields to be in a certain state anymore. I think that we should be filling the repo_path and repo_name whenever a repo is added, so that we can be certain it existing since any repo that is added will have them set. The method to put this in is in called `insert` and is on the `Repo` class in the `augur_data.py` file.

IsaacMilarky:
> If I am understanding these changes correctly, it looks like Facade is still setting the repo_name and repo_path, it just doesn't depend on those fields to be in a certain state anymore. I think that we should be filling the repo_path and repo_name whenever a repo is added, so that we can be certain it existing since any repo that is added will have them set. The method to put this in is in called `insert` and is on the `Repo` class in the `augur_data.py` file.

Sounds good. I didn't know it was that simple with our orm setup.

ABrain7710 (changes requested):
Do we still need the UPDATE query on line 96 of `facade05repofetch.py` that sets the `repo_name` and `repo_path`? Since we are inserting the `repo_name` and `repo_path` with every new repo. Also I think there should be a revision script that inserts the `repo_name` and `repo_path` for any repos that don't have the `repo_path` or `repo_name`

ABrain7710:
@IsaacMilarky Okay, so does this mean Facade still relies on the `repo_name` and `repo_path` it just doesn't depend on them being NULL? Also it seems if we are worried about the regex not producing the same results as Facade then we should put the way Facade is parsing it into a function, and then use that function in the Repo insert function and in the schema script. Then we would be confident that it is parsing them the way Facade needs them.

IsaacMilarky:
> @IsaacMilarky Okay, so does this mean Facade still relies on the `repo_name` and `repo_path` it just doesn't depend on them being `NULL`? Also it seems if we are worried about the regex not producing the same results as Facade then we should put the way Facade is parsing it into a function, and then use that function in the `Repo.insert()` function and in the schema script. Then we would be confident that it is parsing them the way Facade needs them.

Facade still needs to keep track of the file that it has cloned a repo from previously for how we are doing collection. We can't really do it the same way that facade does it when they are inserted since that relies on the git repo being cloned onto disk. It does python indexing and some regex to get the path and name once the repo has been cloned.

PR #2192, opened 2023-02-25
Db Restructure - Phase 1

ABrain7710:
**Description**
- Define `GithubTaskManifest` which is a utility class for Github Tasks. It aims to eventually replace the `GithubTaskSession` class. 
- I didn't remove the `GithubTaskSession` class in the pr, but rather replaced all the places where the `GithubTaskSession` class is used in Github tasks with the `GithubTaskManifest`. So there are a few facade related tasks that still use it. 
- The goal of this approach is to keep prs small by incrementally phasing out the `GithubTaskSession` rather than doing it all at once
- I also changed some functions signatures so that get passed the key_auth class rather than retrieving it on the session. This makes it easier to tell what a function really needs to run

**Signed commits**


sgoggins (approved):
LGTM

PR #2227, opened 2023-03-09
Changes working towards quantifying repos by their metrics to not overwhelm collection

IsaacMilarky:
**Description**
- Work towards being able to schedule a predefined load of collection by repo sizes rather than raw repo amount
- Add functions to see how large a repo is respective of its pull_requests and issues
- Split start_tasks.py into two to make it less overwhelming
- Split facade collection into two hooks. One to clone/update, one to collect data. This is done so that in the future similar metrics to the pr/issue metrics to see how large a repo is respective of its commits
- Fix issues with logic when handling dependency errors
- Add new status for facade when it is cloning and updating called 'Initializing'
- Change celery settings to work with newly split start_tasks.py
- Fix issue with sqlalchemy where 'one' was used instead of 'first'.

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -431,6 +431,45 @@ def generate_contributor_sequence(logger,repo_git, session):
     return insert_facade_contributors.si(repo_id)
 
 
+def facade_clone_update_phase(repo_git):
+    logger = logging.getLogger(git_repo_initialize_facade_task.__name__)
+    logger.info(f"Generating sequence to update/clone repo {repo_git}")
+
+    with FacadeSession(logger) as session:
+        
+        facade_sequence = []
+
+        #Get the repo_id
+        repo_list = s.sql.text("""SELECT repo_id,repo_group_id,repo_path,repo_name FROM repo 
+        WHERE repo_git=:value""").bindparams(value=repo_git)
+        repos = session.fetchall_data_from_sql_text(repo_list)
+
+        start_date = session.get_setting('start_date')
+
+        repo_ids = [repo['repo_id'] for repo in repos]
+
+        repo_id = repo_ids.pop(0)
+
+        #Get the collectionStatus
+        query = session.query(CollectionStatus).filter(CollectionStatus.repo_id == repo_id)
+
+        status = execute_session_query(query,'one')
+        
+        # Figure out what we need to do
+        limited_run = session.limited_run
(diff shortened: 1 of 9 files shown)


augur/tasks/git/facade_tasks.py
```diff
@@ -457,43 +496,13 @@ def facade_phase(repo_git):
         
         # Figure out what we need to do
         limited_run = session.limited_run
-        delete_marked_repos = session.delete_marked_repos
```
sgoggins, line 460: I think we are getting rid of many of the confusing different settings. yes?

augur/tasks/github/facade_github/tasks.py
```diff
@@ -51,14 +51,14 @@ def process_commit_metadata(session,contributorQueue,repo_id):
...
-            query = session.query(Contributor).filter_by(cntrb_full_name=name)
-            contributors_with_matching_name = execute_session_query(query, 'one')
 
-            login = contributors_with_matching_name.gh_login
+        query = session.query(Contributor).filter_by(cntrb_full_name=name)
+        contributors_with_matching_name = execute_session_query(query, 'first')
 
-        except NoResultFound as e:
-            session.logger.debug(f"Failed local login lookup with error: {e}")
+        if not contributors_with_matching_name:
```
sgoggins, line 58: Will taking this out of a try except cause the collection to fail in the case that a person is not identified?

augur/tasks/start_tasks.py
```diff
@@ -27,161 +27,17 @@
 from augur.tasks.db.refresh_materialized_views import *
 # from augur.tasks.data_analysis import *
```
sgoggins, line 28: I definitely want @ABrain7710 to review this file

sgoggins (dismissed):
I approve as far as I can tell, and some of he changes are significant enough that I want @ABrain7710 to review before merging.

augur/tasks/git/facade_tasks.py
IsaacMilarky: Yes some have been removed but not all. Delete_marked_repos particularly should probably be removed because we never remove repos.

augur/tasks/github/facade_github/tasks.py
ABrain7710: No, because it uses `first` which will return None if it doesn't exist. Where `one` expects a row to exist

ABrain7710 (approved):

PR #2229, opened 2023-03-09
Viz fix

Ulincsys:
**Description**
- Fixes visualizations broken since augur view merger into chaoss/augur

The merger of these two applications made the choice of an external cache directory impossible, so the application must now cache files in a directory relative to the root of the gunicorn server. The `static` directory is used for convenience.

- Yes, I signed my commits.


sgoggins (approved):
Lgtm

ABrain7710 (approved):

PR #2235, opened 2023-03-12
Python module imports detection patch

JamesKunstle:
Uses abstract syntax tree (AST module) of Python file instead of regex to identify python imports.

also detects submodules.

**Signed commits**

### DIFF
--- a/augur/tasks/git/dependency_tasks/dependency_util/python_deps.py
+++ b/augur/tasks/git/dependency_tasks/dependency_util/python_deps.py
@@ -2,17 +2,32 @@ import sys
 import re
 from pathlib import Path
 import codecs
+import ast
+
 
 def get_files(path):
-	#copied from example on https://docs.python.org/3/library/pathlib.html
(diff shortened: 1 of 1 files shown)


augur/tasks/git/dependency_tasks/dependency_util/python_deps.py
```diff
@@ -2,17 +2,32 @@
 import re
 from pathlib import Path
 import codecs
+import ast
```
sgoggins, line 5: I think we need to add this to the setup.py for the task. @ABrain7710 / @IsaacMilarky ?

sgoggins (approved):

augur/tasks/git/dependency_tasks/dependency_util/python_deps.py
IsaacMilarky: ast is builtin

IsaacMilarky (approved):

PR #2236, opened 2023-03-13
Add logic to handle if repo's host returns 404

IsaacMilarky:
**Description**
- Change collection to set repo to 'Error' when repo can't be reached by reason of 404

**Signed commits**

### DIFF
--- a/augur/tasks/github/detect_move/core.py
+++ b/augur/tasks/github/detect_move/core.py
@@ -50,6 +50,11 @@ def ping_github_for_repo_move(augur_db, key_auth, repo, logger,collection_hook='
         logger.info(f"Repo found at url: {url}")
         return
     
+    #Mark as errored if not found
+    if response_from_gh.status_code == 404:
+        logger.error(f"Repo {repo.repo_git} responded 404 when pinged!")
+        raise Exception(f"ERROR: Repo not found at requested host {repo.repo_git}")
+    
     owner, name = extract_owner_and_repo_from_endpoint(key_auth, response_from_gh.headers['location'], logger)
 
     current_repo_dict = repo.__dict__
--- a/augur/tasks/github/detect_move/core.py
+++ b/augur/tasks/github/detect_move/core.py
@@ -5,6 +5,7 @@ from augur.tasks.github.util.github_paginator import hit_api
 from augur.tasks.github.util.util import get_owner_repo
 from augur.tasks.github.util.util import parse_json_response
 import logging
+from datetime import datetime
 from enum import Enum
 from augur.application.db.util import execute_session_query
 
@@ -13,6 +14,20 @@ class CollectionState(Enum):
     PENDING = "Pending"
     ERROR = "Error"
     COLLECTING = "Collecting"
+
+
+def update_repo_with_dict(current_dict,new_dict,logger,db):
+    
+    
+    to_insert = current_dict
+    del to_insert['_sa_instance_state']
+    to_insert.update(new_dict)
+
+    result = db.insert_data(to_insert, Repo, ['repo_id'])
+
+    url = to_insert['repo_git']
+    logger.info(f"Updated repo for {url}\n")
+
 
 
 def extract_owner_and_repo_from_endpoint(key_auth, url, logger):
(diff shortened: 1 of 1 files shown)


augur/tasks/github/detect_move/core.py
```diff
@@ -44,6 +60,23 @@ def ping_github_for_repo_move(augur_db, key_auth, repo, logger,collection_hook='
         logger.warning(f"Could not check if repo moved because the api timed out 10 times. Url: {url}")
         return
 
+    
+    
+    #Mark as errored if not found
+    if response_from_gh.status_code == 404:
+        logger.error(f"Repo {repo.repo_git} responded 404 when pinged!")
+
```
sgoggins, line 68: When we get 404's we should probably do at least one retry before updating the description.

augur/tasks/github/detect_move/core.py
```diff
@@ -44,6 +60,23 @@ def ping_github_for_repo_move(augur_db, key_auth, repo, logger,collection_hook='
...
+        repo_update_dict = {
+        'repo_git': repo.repo_git,
+        'repo_path': None,
+        'repo_name': None,
+        'description': f"During our check for this repo on {datetime.today().strftime('%Y-%m-%d')}, a 404 error was returned. The repository does not appear to have moved. Instead, it appears to be deleted"
```
sgoggins, line 73: What is the current equivalent of setting the status to 'New' for recloning? We should probably do that as well. Maybe its in here and we don't see it?

augur/tasks/github/detect_move/core.py
```diff
@@ -69,11 +99,7 @@ def ping_github_for_repo_move(augur_db, key_auth, repo, logger,collection_hook='
         'description': f"(Originally hosted at {url}) {old_description}"
     }
 
-    current_repo_dict.update(repo_update_dict)
-
-    result = augur_db.insert_data(current_repo_dict, Repo, ['repo_id'])
-
-    logger.info(f"Updated repo for {owner}/{name}\n")
+    update_repo_with_dict(current_repo_dict, repo_update_dict, logger,augur_db)
```
sgoggins, line 102: Maybe this is where a repo gets set to be recloned the next time?

sgoggins (commented):
I made some comments.

augur/tasks/github/detect_move/core.py
IsaacMilarky: detect_move is only used for non-facade tasks. Facade has it's own logic to handle if repos can't be cloned. Facade set's the status to 'Failed Clone' if we can't clone it.

augur/tasks/github/detect_move/core.py
IsaacMilarky: This updates the repo url with the one provided by github but it doesn't affect facade.

augur/tasks/github/detect_move/core.py
IsaacMilarky: I agree I will implement that

augur/tasks/github/detect_move/core.py
sgoggins: What if its deleted after we have already cloned it and populated commits? Should we have a different status for that?

sgoggins (commented):
The cases we are encountering include those where the repo has previously been cloned and counted, then goes away. How are these cases handled?

IsaacMilarky:
> The cases we are encountering include those where the repo has previously been cloned and counted, then goes away. How are these cases handled?

Facade's status is different from the other tasks so this logic won't effect that. Right now if a repo can't be cloned or updated it has a status of "Failed Clone"

augur/tasks/github/detect_move/core.py
IsaacMilarky: I don't think it makes much of a difference. We could make an extra status, something like 'Stale', to be more clear about the repo's status in the database.

sgoggins: I think the thing that's important is making sure we don't get a singular network issue that marks all the repositories some kind of "stale'.

sgoggins (approved):
I approve this, AND I strongly prefer adding some logic to retry so a single 1 hour network failure does not mark all the repos incorrectly stale.

PR #2237, opened 2023-03-13
Improve AugurTaskRoutine and create classes/methods for future weighted collection

IsaacMilarky:
**Description**
- Created subclass to represent the total weight of a set of repo collection tasks that are allowed to be scheduled at once. This class is not used yet but will make implementing weighted collection easier. 
- Edit AugurTaskRoutine class to fully encapsulate starting repo collection on a set of given repositories. Now the class is used by setting the repos to start collection over, the phases to run over those repos, and the collection hook those repos are a part of.
- Create child class of AugurTaskRoutine AugurWeightedTaskRoutine that also encapsulates starting repo collection on a set of given repositores, but takes a weight value where it won't run a set of repos that have a total weight value greater than the given value, i.e. a weight of 10,000 will run a set of repos where the total amount of pull requests and issues over all the repositories being scheduled are hard capped at 10,000 where it keeps a counter that counts down from 10,000 and subtracts each repos weight as they are scheduled stopping when the counter reaches zero. This class is also not used yet but will make implementing weighted collection easier. 
- Create function get_repo_weight_by_commit that returns the amount of commits a repo has weighted by days since last collection. 
- Fix incorrect graphql function call in repo_info task

**Signed commits**


sgoggins (commented):
I need @ABrain7710 to review this one. Its deep in the guts of scheduling and I cannot anticipate all of what the impacts might be.

sgoggins (approved):

PR #2241, opened 2023-03-17
Add exception handle in dependency analysis to use regex if ast can't be used due to version 

IsaacMilarky:
**Description**
Fixed issue with python dependency analysis where versions older than python 3.4 couldn't be parsed. This was because the current version of python's ast (abstract syntax tree) module only supports versions greater than or equal to 3.4. The new method used uses a simple regex to find dependencies if issues are encountered using the ast method. 

**Signed commits**


sgoggins (approved):
LGTM

PR #2251, opened 2023-03-20
Change dependency tasks to bulk transaction

IsaacMilarky:
**Description**
Change dependency related tasks to use bulk-upsert transactions instead of executing a statement every iteration. 

**Signed commits**


sgoggins (approved):
@ABrain7710 : Is this what you meant?

PR #2261, opened 2023-03-22
Dev bugfixes 

IsaacMilarky:
**Description**
- Fix typo similar to previous issue where methods that used to take a session now take a logger and a key_auth object instead but the change was not made correctly universally. 
- Delete unused variable in libyear logic
- Stop libyear logic from trying to parse an empty date.
- Handle issue where regex was encountering empty string. 

**Signed commits**


sgoggins (approved):
LGTM

PR #2264, opened 2023-03-23
Performance updates and misc.

sgoggins:
**Description**
- I think we should check the number of cores and use that information for setting parallel processes in the next release. My heuristic was, on a 64 CPU server, I set it to 50. This is working well. I think ~14 processes running in parallel **less** that those on the hardware is optimal. We have to check for "CPU", NOT "Cores" because the core checking does not work on OSX. So, we can go with ~14 less than the CPUS available for safety, or a default of 14. Whichever is greater.

We could also simply check the OS, and if its a linux flavor, evaluate cores.


sgoggins:
There is also a patch for the commits table here, setting the affiliation data to `NULL` instead of a string "NULL". The updates on the string "NULL" are very expensive.

augur/tasks/start_tasks.py
```diff
@@ -372,7 +372,7 @@ def augur_collection_monitor():
 
         if facade_phase.__name__ in enabled_phase_names:
             #Schedule facade collection before clone/updates as that is a higher priority
-            start_facade_collection(session, max_repo=30, days=30)
```
IsaacMilarky, line 375: I'm slightly worried that this might crowd out the other task processes since facade shares a worker queue with the core task collection. Now that we are switching to having 50 processes available for the core worker we could do something like splitting it 50/50 by having start_tasks schedule 25 repos each for facade and core collection.

IsaacMilarky (changes requested):
This mostly looks good although I would change the repo scheduling to be more balanced now that we have a lot more processes to use and don't have to have as many messages in the queue.

ABrain7710:
I agree with @IsaacMilarky comment. Maybe like 40 processes for core and 15 for secondary? This would about triple the number of processes each worker

augur/tasks/start_tasks.py
sgoggins: @IsaacMilarky : OK. This makes a lot of sense. I did not understand that this was coming out of the core allocation. I will adjust.

IsaacMilarky: Shouldn't we make the primary collection start 30 repos and then have facade collection start 30 repos to evenly distribute the available processes among the repos being scheduled since we have 60 processes? I would prefer to not put more messages on the queue than we need to and in this pr it is scheduling 85 repos at once for 60 processes. Especially since we aren't using eventlet anymore. Is there a reason we are scheduling so many repos at once?  We will still have to keep the messages for these repos in the queue bc of all the jobs we are doing for each repo but under this configuration we have like ~25 repos just sitting in the rabbitmq queue doing nothing and not even have a reason to be in the queue because none of those would have tasks being processed by the worker processes at that point.

IsaacMilarky (changes requested):
I would set max_repo to 30 for start_primary_collection and I would set max_repo to 30 for start_facade_collection. Otherwise LGTM :)

augur/tasks/github/facade_github/tasks.py
```diff
@@ -190,8 +190,9 @@ def link_commits_to_contributor(session,contributorQueue):
                 UPDATE commits 
                 SET cmt_ght_author_id=:cntrb_id
                 WHERE 
-                cmt_author_raw_email=:cntrb_email
-                OR cmt_author_email=:cntrb_email
+                (cmt_author_raw_email=:cntrb_email
+                OR cmt_author_email=:cntrb_email)
+                AND cmt_ght_author_id is NULL
```
sgoggins, line 195: @IsaacMilarky : Without that `AND cmt_ght_author_id is NULL` we are updating a *lot* more rows than we need to *every* time. I think this helps explain why the commits table started to bloat. I checked this with the following SQL `select * from commits where cmt_ght_author_id is NULL limit 1000` and 
```sql
SELECT
	* 
FROM
	commits 
WHERE
  cmt_ght_author_id IS NULL AND
	( cmt_author_raw_email = 'dev@shardingsphere.apache.org' OR cmt_author_email = 'dev@shardingsphere.apache.org' ) 
	LIMIT 10
```

We will be skipping a significant number of costly updates.

sgoggins: Put another way: Every time postgresql does an update on a row, it copies the row to new physical space that's slowly recovered with autovacuum. I strongly suspect that by updating every row, even if it was already populated, was creating a ton of dead rows on the commits table very quickly. My working theory is that this will fix this problem. I am going to test that theory around March 29.

augur/application/cli/backend.py
```diff
@@ -107,8 +107,8 @@ def start(disable_collection, development, port):
             os.remove("celerybeat-schedule.db")
 
         scheduling_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=1 -n scheduling:{uuid.uuid4().hex}@%h -Q scheduling"
-        core_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=14 -n core:{uuid.uuid4().hex}@%h"
-        secondary_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=5 -n secondary:{uuid.uuid4().hex}@%h -Q secondary"
+        core_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=70 -n core:{uuid.uuid4().hex}@%h"
```
IsaacMilarky, line 110: We are scheduling 80 repos for 70 processes (60 on primary 20 on facade). I don't think we should do this. I think we _really_ should schedule a number of repos less than or equal to the number of processes we have. Otherwise we put things in rabbitmq when we don't have to. I think we shouldn't use rabbitmq if we don't have to bc large queue sizes have been an issue in the past.

augur/tasks/github/facade_github/tasks.py
IsaacMilarky: That makes sense to me. This looks much better.

augur/application/cli/backend.py
IsaacMilarky: We should not schedule more repos than processes because it provides no benefit and can only cause issues, as we already have the information for that pending repo in the db. If we schedule more repos than processes we will have a repo in the rabbitmq queue doing nothing because it has no processes. Repos that are running still _have_ to use rabbitmq because there are multiple jobs per repo and it has to store the messages for these jobs in rabbitmq. And repos that don't have available processes should not have messages in rabbitmq

augur/tasks/start_tasks.py
```diff
@@ -365,14 +365,14 @@ def augur_collection_monitor():
         enabled_phase_names = get_enabled_phase_names_from_config(session.logger, session)
 
         if primary_repo_collect_phase.__name__ in enabled_phase_names:
```
IsaacMilarky, line 367: In the future we should get the amount of repos to schedule from the amount of processes we have available and then divide them among the collection hooks. This way one repo gets one process. For now we can just hard code it I think.

IsaacMilarky (changes requested):
I still don't think we should schedule more repos than processes we have available.

sgoggins:
> 

I went with 30 for primary and 20 for secondary @ABrain7710 because secondary takes so much longer and core is always waiting for it to finish.

PR #2285, opened 2023-04-04
Move github contributors query outside of facade collection hook

IsaacMilarky:
**Description**
- Move the grab_committers task out of the facade collection hook. This is done because it was slowing down the rest of facade collection and because the task doesn't rely on commit data. The facade collection hook should really be reserved for tasks that directly relate to facade commit data.
- Improve grab_committers task to use GithubTaskManifest instead of GithubTaskSession which is in the process of being deprecated. 
- Improve grab_committers task to use new retrieve_dict_from_endpoint method instead of request_dict_from_endpoint method which is in the process of being deprecated.
- Create retrieve_dict_from_endpoint method that is more uniform with the rest of how augur now paginates the github REST api. This method is better because it checks the validity of returned data the same way as everything else instead of the older way that request_dict_from_endpoint does. The new method is modeled after the retrieve_data method used in pagination but for the purpose of single data retrieval instead which is used throughout contributor resolution. 

**Signed commits**

### DIFF
--- a/augur/tasks/github/contributors/tasks.py
+++ b/augur/tasks/github/contributors/tasks.py
@@ -7,6 +7,7 @@ from augur.application.db.data_parse import *
 from augur.tasks.github.util.github_paginator import GithubPaginator, hit_api
 from augur.tasks.github.util.github_task_session import GithubTaskManifest
 from augur.tasks.util.worker_util import wait_child_tasks
+from augur.tasks.github.facade_github.tasks import *
 from augur.application.db.models import PullRequest, Message, PullRequestReview, PullRequestLabel, PullRequestReviewer, PullRequestEvent, PullRequestMeta, PullRequestAssignee, PullRequestReviewMessageRef, Issue, IssueEvent, IssueLabel, IssueAssignee, PullRequestMessageRef, IssueMessageRef, Contributor, Repo
 from augur.application.db.util import execute_session_query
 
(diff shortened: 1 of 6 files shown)


augur/tasks/github/contributors/tasks.py
```diff
@@ -104,5 +106,19 @@ def retrieve_dict_data(url: str, key_auth, logger):
...
+    from augur.tasks.init.celery_app import engine
+
+    logger = logging.getLogger(grab_comitters.__name__)
+    with DatabaseSession(logger,engine) as session:
+
+        repo = session.query(Repo).filter(Repo.repo_git == repo_git).one()
+        repo_id = repo.repo_id
+
+    try:
+        grab_committer_list(GithubTaskManifest(logger), repo_id,platform)
```
ABrain7710, line 121: Need to use context manager with the `GithubTaskManfiest` to ensure everything is closed

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -406,18 +407,18 @@ def get_login_with_commit_hash(session, commit_data, repo_id):
 
 
 
-def create_endpoint_from_repo_id(session, repo_id):
+def create_endpoint_from_repo_id(manifest, repo_id):
```
ABrain7710, line 410: This function should be changed, so it takes `augur_db`, `logger`, and `repo_id` as arguments. The manifest class is not made to be passed around, it is made to be a utility class that makes it easier to create all the needed objects for a github task and that all those objects are closed properly.

augur/tasks/github/util/github_paginator.py
```diff
@@ -550,4 +551,68 @@ def get_url_page_number(url: str) -> int:
     except KeyError:
         return 1
 
-    return page_number
\ No newline at end of file
+    return page_number
+
+
+def retrieve_dict_from_endpoint(manifest, url, timeout_wait=10) -> Tuple[Optional[dict], GithubApiResult]:
```
ABrain7710, line 557: Similar to above. This should take a logger, key_auth, url, and timeout_wait

ABrain7710 (commented):
Overall looks good. Just some small things need to be fixed to keep consistency across augur with the `GithubTaskManifest`

ABrain7710:
@sgoggins Can this be reverted? I think I accidentally left a comment rather than requesting changes. I would like those small things to be done before it is merged.

PR #2291, opened 2023-04-05
Move query of contributors from the github api to the core collection hook from the facade hook 

IsaacMilarky:
**Description**
This pr is a version of https://github.com/chaoss/augur/pull/2285 with the changes requested from @ABrain7710 

**Signed commits**


sgoggins (commented):
We do end up with some unresolved contributors on commits sometimes. This seems kind of impossible if we are querying the commit hash for the gh_login. Does this fix this, or at least not make it worse?

sgoggins (approved):

PR #2294, opened 2023-04-06
Remove stack trace print from invalid command output

ABrain7710:
**Description**
- Changed the invalid command exception, so that it no longer prints out the entire stack trace when a user enters an invalid command

**Signed commits**

### DIFF
--- a/augur/application/cli/_multicommand.py
+++ b/augur/application/cli/_multicommand.py
@@ -28,9 +28,7 @@ class AugurMultiCommand(click.MultiCommand):
             module = importlib.import_module('.' + name, 'augur.application.cli')
             return module.cli
         except ModuleNotFoundError:
-            print(f"Error")
-            print(traceback.format_exc())
-  
+            pass
 
 @click.command(cls=AugurMultiCommand, context_settings=CONTEXT_SETTINGS)
 @click.pass_context


augur/application/cli/_multicommand.py
```diff
@@ -28,9 +28,7 @@ def get_command(self, ctx, name):
             module = importlib.import_module('.' + name, 'augur.application.cli')
             return module.cli
         except ModuleNotFoundError:
-            print(f"Error")
```
IsaacMilarky, line 31: We should probably print an 'invalid command' message to stderr when this happens

IsaacMilarky (changes requested):

augur/application/cli/_multicommand.py
ABrain7710: The `click` library already prints this by default.

Here is an example when I put in `augur backen start`

Usage: augur [OPTIONS] COMMAND [ARGS]...
Try 'augur --help' for help.

Error: No such command 'backen'.

IsaacMilarky: I gotcha that makes sense.

IsaacMilarky (approved):

PR #2306, opened 2023-04-07
Update contributor logic to be more testable and uniform

IsaacMilarky:
**Description**
Update contributor resolution logic to be cleaner and more testable. Also now uses less deprecated functions overall. GithubTaskSession is no longer used now in favor of a GithubTaskManifest that passes its attributes.


sgoggins (approved):
LGTM. 

let's make sure the API call to the gh_login for the commit is used. We should have nearly 100% resolution that way.

IsaacMilarky:
Yep that API call is used in contributor resolution :)

PR #2314, opened 2023-04-12
Remove the use of shell for mkdir -p

mscherer:
See https://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python

**Description**

This PR fixes a problem I bumped into while deploying (eg, the container do not work without /bin in PATH). I have a workaround for my setup, but it is better to fix the root cause.

**Notes for Reviewers**

**Signed commits**


augur/tasks/git/util/facade_worker/facade_worker/facade05repofetch.py
```diff
@@ -114,21 +115,7 @@ def git_repo_initialize(session, repo_git):
...
-        if return_code != 0:
-            print("COULD NOT CREATE REPO DIRECTORY")
-
-            update_repo_log(session, row.repo_id,'Failed (mkdir)')
-            session.update_status(f"Failed (mkdir {repo_path})")
-            session.log_activity('Error',f"Could not create repo directory: {repo_path}" )
-
-            raise Exception("Could not create git repo's prerequisite directories. "
-                " Do you have write access?")
+        pathlib.Path(repo_path).mkdir(parents=True, exist_ok=True)
```
IsaacMilarky, line 118: This is a good change however it should still catch any exceptions Path.mkdir raises and respond like how it did before when we didn't get a return code. It should update its internal logs when it is unable to create the directory.

IsaacMilarky: Just put the mkdir line in a try/except block and put the body of the if statement on line 123 in the exception handle. Otherwise looks good.

IsaacMilarky (changes requested):
Just missing some exception handling.

IsaacMilarky (approved):
Looks good to me now.

sgoggins (commented):
@IsaacMilarky / @ABrain7710 : Could you go ahead and review this?

ABrain7710 (approved):

PR #2317, opened 2023-04-14
Add pytest fixture that creates and takes down empty augur database

ABrain7710:
**Description**
- Add pytest fixture that creates a temporary augur database that looks just like a brand new augur database that was initialized with alembic. Then after the test runs the database gets torn down

**Signed commits**


sgoggins (approved):
Fantastic to have this done!

PR #2318, opened 2023-04-14
Fix seed data so repo path and repo name are correct

ABrain7710:
**Description**
- The repo path and repo name are not correct in the seed data. I figured out that it was because the regex in the 7th migration was not capturing the repo path and repo name correctly, so I changed it to use the standard function that gets the repo owner and repo name. Then I constructed the repo path with the repo owner

**Signed commits**


sgoggins (approved):
Thanks for fixing this!

PR #2333, opened 2023-04-20
Avoid import * from .init

mscherer:
'import *' shouldn't be used, as that may import
variable and modules that overrides existing symbol, and have side effect. It also make harder to refactor the code.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/api/view/utils.py
+++ b/augur/api/view/utils.py
@@ -1,16 +1,13 @@
 from pathlib import Path
 from concurrent.futures import ThreadPoolExecutor
 from flask import render_template, flash, url_for, Flask
-from .init import *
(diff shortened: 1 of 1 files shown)


Ulincsys:
This will cause an `ImportError`, because `logger` cannot be imported before it is created. It must be imported from `init` after `init_logging()` is called.

Ulincsys (changes requested):
It is fine to remove `import *`, but the logger must be initialized before it can be imported.

Ulincsys:
Closing as this has unresolved breaking changes. Please feel free to reopen if a resolution is reached.

PR #2339, opened 2023-04-20
fix constraint issues with secondary and facade collection hooks

IsaacMilarky:
**Description**
Fix constraints on the collection_status table to reflect edge cases of collection.

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_operations.py
+++ b/augur/application/db/models/augur_operations.py
@@ -997,44 +997,38 @@ class CollectionStatus(Base):
     __tablename__ = "collection_status"
     __table_args__ = (
         CheckConstraint(
-            "(core_data_last_collected IS NOT NULL AND core_status = 'Success') OR "
-            "(core_data_last_collected IS NULL AND core_status = 'Pending') OR "
-            "(core_status = 'Error') OR "
-            "(core_status = 'Collecting')",
+            "NOT (core_data_last_collected IS NULL AND core_status = 'Success') AND "
+            "NOT (core_data_last_collected IS NOT NULL AND core_status = 'Pending')",
             name='core_data_last_collected_check'
         ),
         CheckConstraint(
-            "(core_task_id IS NULL AND core_status IN ('Pending', 'Success', 'Error')) OR "
-            "(core_task_id IS NOT NULL AND core_status = 'Collecting')",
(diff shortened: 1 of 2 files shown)


sgoggins (dismissed):
LGTM. I think @ABrain7710 should also check it so he knows what's changed.

ABrain7710 (changes requested):
After thinking about the core and secondary dependency check. I think it should just test the negative case rather than specifying all the different positive cases. So I think it should just be `NOT(core_status = 'Pending' AND secondary_status='Collecting')`, so we ensure that the secondary status isn't collecting when the core status is pending. I think the rest of cases are valid

augur/application/db/models/augur_operations.py
```diff
@@ -997,44 +997,38 @@ class CollectionStatus(Base):
...
+            "NOT (facade_task_id IS NULL AND facade_status IN ('Pending', 'Success', 'Error', 'Failed Clone')) AND "
+            "NOT (facade_task_id IS NOT NULL AND facade_status IN ('Collecting','Initializing'))",
             name='facade_task_id_check'
         ),
         CheckConstraint(
-            "(core_status = 'Success') OR "
-            "(core_status IN ('Pending', 'Collecting', 'Error') AND secondary_status = 'Pending')",
+            "NOT (core_status IN ('Success','Collecting')) AND "
+            "NOT (core_status IN ('Pending', 'Collecting', 'Error') AND secondary_status = 'Pending')",
             name='core_secondary_dependency_check'
```
ABrain7710, line 1032: I'm getting really confused on this check. Wouldn't this be the one case we want to restrict?
`NOT(core_status = 'Pending' AND secondary_status='Collecting')`

ABrain7710:
@IsaacMilarky I think it would be good if we listed out all the cases we are wanting to avoid. Then write a boolean expression that matches each case. Then all we would have to do is wrap each case in NOT and place an OR in between each. I'm having a lot of trouble going through and wrapping my head around these conversions and I think there are some errors. So I think it would be best if we used this simple structure of 
Not(invalid case) OR NOT(invalid_case) OR NOT(invalid_case) within each check

ABrain7710 (approved):
LGTM

PR #2346, opened 2023-04-21
Make sure weight is updated after collection is completed for each collection hook 

### LINKED ISSUE #2342: Collection Weight doesn't penalize already collected repos until 1 day later
**Description:**
- The collection weight isn't penalizing repos that have already been collected until a day after there first collection (when the update collection weight task is run). This will mean that really small repos will be collected multiple times in the same day util eventually the update collection weight task is completed

**How to reproduce:**
1. Run augur with a small number of repos
2. Watch the core_weight and see that it doesn't increase after a repo has been collected

**Expected behavior:**
A repo's weight should be updated right after it is finished collecting

IsaacMilarky:
**Description**
Append Augur's collection hooks to update the weight after collection has finished. Core, facade, and secondary collection hooks now update the weight of the repo after they complete and populate the last_collected timestamp. 

This PR Fixes #2342

**Signed commits**

### DIFF
--- a/augur/tasks/start_tasks.py
+++ b/augur/tasks/start_tasks.py
@@ -311,6 +311,11 @@ def start_facade_collection(session,max_repo):
 
     facade_enabled_phases.append(facade_task_success_util_gen)
 
+    def facade_task_update_weight_util_gen(repo_git):
+        return git_update_commit_count_weight.si(repo_git)
+
+    facade_enabled_phases.append(facade_task_update_weight_util_gen)
+
     active_repo_count = len(session.query(CollectionStatus).filter(CollectionStatus.facade_status == CollectionState.COLLECTING.value).all())
 
     #cutoff_date = datetime.datetime.now() - datetime.timedelta(days=days)
--- a/augur/tasks/start_tasks.py
+++ b/augur/tasks/start_tasks.py
@@ -87,7 +87,7 @@ def primary_repo_collect_phase(repo_git):
 
     repo_task_group = group(
         repo_info_task,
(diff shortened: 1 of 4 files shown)


augur/tasks/start_tasks.py
```diff
@@ -87,7 +87,7 @@ def primary_repo_collect_phase(repo_git):
 
     repo_task_group = group(
         repo_info_task,
-        chain(primary_repo_jobs | core_task_update_weight_util.s(repo_git=repo_git),secondary_repo_jobs,process_contributors.si()),
+        chain(primary_repo_jobs | issue_pr_task_update_weight_util.s(repo_git=repo_git),secondary_repo_jobs,process_contributors.si()),
         #facade_phase(logger,repo_git),
```
ABrain7710, line 91: Wouldn't the core data last collected still be NULL when the `issue_pr_task_update_weight_util` runs, since core collection isn't done yet. I'm confused as to why updating the secondary and facade weights are done in similar ways, but core is completely different. Could we just do it the same way for core?

IsaacMilarky: It is done in the same way for core. That chord is to update the issue and pr raw count from the primary tasks. The method at the end of the hook then updates it following the collection without updating the issue and pr raw count.

IsaacMilarky: core_task_success_util updates the core weight and secondary weight in the same way as secondary_task_success_util. The call to issue_pr_task_update_weight_util is just to update the raw count

ABrain7710: Oh, I see now. That makes sense

ABrain7710 (approved):

PR #2351, opened 2023-04-24
Fix syntax errors in backend.py

ABrain7710:
**Description**
- The `facade_worker_process` variable was only defined if running collection, otherwise it resulted in this error
`UnboundLocalError: local variable 'facade_worker_process' referenced before assignment`.
- Also there was this typo `facade_worker_process,terminate()` which should be`facade_worker_process.terminate()`

**Signed commits**


sgoggins (approved):
LGTM

PR #2352, opened 2023-04-24
Add a delay of a day before collecting repositories again

### LINKED ISSUE #2349: Secondary tasks never run when repo count is small
**Description:**
When running on the default repos, I noticed the secondary tasks weren't getting run. This is happening because the core collection is finishing really quickly, updating the weight to a large number, and then the next time around there are no new repos to collect so the scheduler is picking all the old core repos to recollect since there is nothing else to collect. In turn the core repos are always collecting whenever the `start_secondary_collection` function gets run.

**How to reproduce:**
1. Create a new database with the default repos
2. Run collection while watching the collection status table
3. See how the core status while go from collecting to success to collecting to success all while the secondary status stays pending

**Expected behavior:**
I think we should only consider repos for recollection when they are a day or more old. This would leave the core status as success for 1 whole day after finishing, so the secondary collection has a chance to start. This is also useful because it means a small instance collecting 30 repos won't recollect each of those repos multiple times a day since there is nothing else to collect

IsaacMilarky:
**Description**
Add a restriction to repository collection when collecting old repositories; disallow repositories that have already been collected from being collected again until a day has passed since it was last collected 

This PR fixes #2349

**Signed commits**

### DIFF
--- a/augur/tasks/start_tasks.py
+++ b/augur/tasks/start_tasks.py
@@ -159,7 +159,7 @@ def non_repo_domain_tasks():
         The below functions define augur's collection hooks.
         Each collection hook schedules tasks for a number of repos
     """
-def start_primary_collection(session,max_repo):
+def start_primary_collection(session,max_repo, days_until_collect_again = 1):
 
     #Get list of enabled phases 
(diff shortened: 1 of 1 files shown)


ABrain7710 (approved):
LGTM

PR #2357, opened 2023-04-25
Make facade scheduling dynamic 

ABrain7710:
**Description**
- Currently facade initialization tasks are not being scheduled often enough to ensure there are always repos available to be collected by the facade collection tasks. To solve this, @IsaacMilarky and I came up with a solution that limits the number of cloning tasks and collection tasks we start based on the number of each task we start rather than just saying a specific number of repos for each can run. This is best explained with examples.
For example. 
Say there are 500 repos ready to be cloned. This is what the augur collection monitor would start each minute, assuming that all the clones finished in the minute, and the collection tasks also finished. If they didn't, then less tasks would be scheduled
1. 108 cloned 0 collected
2. 108 cloned 21 collected
3. 108 cloned 21 collected
4. 108 cloned 21 collected
5. 68 cloned 24 collected
6. 0 cloned 30 collected
6. 0 cloned 30 collected

As you can see, this process prioritizes cloning repos while still giving some room in the queue to start collecting some of the repos. Then eventually once all of the repos are cloned the whole queue is used for collection. Essentially the process starts up to 108 cloning jobs and up to 30 collection jobs depending on how many clones are running.

Note: All 30 of the collection jobs should finish each minute if the repos have less than 600 commits. If we wanted to keep the pipe full the entire minute for repos with an average of 100 commits, then we would need to schedule around 75-100 repos. I just started with 30 repos so that we were at least collecting for the whole minute when the repos are around 600 commits. Also these numbers are taken from some of the chaoss repos commit counts and how long they took to run through facade collection, repos with larger or smaller commits could yield slightly different results. The goal I'm shooting for is minimizing downtime in the queue since we are scheduling the smallest repos first

**Signed commits**

### DIFF
--- a/augur/tasks/start_tasks.py
+++ b/augur/tasks/start_tasks.py
@@ -273,6 +273,46 @@ def start_secondary_collection(session,max_repo, days_until_collect_again = 1):
         )
 
 
+
+def start_facade_collections(session, pipe_size, clone_percentage=0.6):
+
+    # for facade collection we are maintaining a pipe size,
+    # this is the number of messages that can be added to the queue at a time for facade
+    # each clone task is counted for 1, each collection task is counted for 12 since those 
+    # are the number of messages they add to the queue
+    
+    # also the clone percentage determines how much of the pipe is dedicated to clone tasks
+    # after cloning is done, the rest of the pipe is filled with collection tasks
+
+    cloning_section_size = pipe_size * clone_percentage
+    cloning_count = len(session.query(CollectionStatus).filter(CollectionStatus.facade_status == CollectionState.INITIALIZING.value).all())
+
+    # remove the number of repos that are already being cloned from cloning_section_size
+    cloning_section_size = cloning_section_size - cloning_count
+
+    # start cloning tasks until the cloning_section is full. We passing cloning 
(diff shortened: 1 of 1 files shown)


IsaacMilarky (changes requested):
I think that these methods should share a pipe size value. As it stands the pipe is split down the middle 50/50 where half of it can have facade collection and half can have clone/update.
This is because start_facade_collection only considers other facade_collection jobs when it checks how many jobs are running in the pipe and start_facade_clone_update only considers other clone/update jobs when it checks how many jobs are running in the pipe.
Effectively we have two separate checks where 15 spots are reserved for collection and 15 are reserved for clone/update
 
What I think we want is being able to completely fill the pipe with facade collection jobs if we have enough cloned repositories to do so. If we have a pipe size of n we want to first attempt to schedule n repos for start_facade_collection. Then, if we ended up starting m repos instead where m < n we should then call start_facade_clone_update with a max_repo value of n - m.

For example, say we have a pipe size of 30. We would first try to schedule 30 repos in start_facade_collection. Turns out we could only schedule 20 repos. At that point we would call start_facade_clone_update with a max_repo of 10.

PR #2359, opened 2023-04-25
Handle edge case facade NULL error

IsaacMilarky:
**Description**
Add an update query to facade to handle the case when an existing repo path is found but the repo's record shows a path of null. This way the path is always updated to the path the repo has been cloned to. 

**Signed commits**


ABrain7710 (approved):
LGTM

PR #2362, opened 2023-04-27
Simplify facade collection process

ABrain7710:
**Description**
- Simplify facade collection scheduling by running the cloning task for all the repos that need cloning. Then scheduling the cloning task 5 minutes after all of them are finished to see if there are any new ones

Note: The facade clone task only gets one repo at a time to initialize so more than one of the task can run at the same time if a clone task is taking more than 5 minutes

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -367,7 +367,7 @@ def clone_repos():
             setattr(repoStatus,"facade_status", CollectionState.UPDATE.value)
             session.commit()
 
-        clone_repos.si().apply_async(countdown=60)
+        clone_repos.si().apply_async(countdown=60*5)
 
 
 
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -342,12 +342,10 @@ def clone_repos():
     # get first repo to process
 
     with FacadeSession(logger) as session:
-        try:
-            repo_git = get_collection_status_repo_git_from_filter(session, is_pending, 1)[0]
-        except IndexError:
-            return
 
-        while repo_git:
+        # process up to 1000 repos at a time
+        repo_git_identifiers = get_collection_status_repo_git_from_filter(session, is_pending, 999999)
+        for repo_git in repo_git_identifiers:
         
             # set repo to intializing
             repo = session.query(Repo).filter(Repo.repo_git == repo_git).one()
@@ -369,11 +367,7 @@ def clone_repos():
             setattr(repoStatus,"facade_status", CollectionState.UPDATE.value)
             session.commit()
 
-            # get next repo
-            try:
-                repo_git = get_collection_status_repo_git_from_filter(session, is_pending, 1)[0]
-            except IndexError:
-                return
+        clone_repos.si().apply_async(countdown=60)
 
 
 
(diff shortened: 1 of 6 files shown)


IsaacMilarky:
Wouldn't we need to also add a check to when we are scheduling the clone phase to take into account the collecting repos on the clone side of the pipe? I think this general solution would work but I think this current setup allows more clones then what is specified in the function call.

IsaacMilarky (changes requested):
I think this would work if the cloning_section_size was also subtracted by how many collecting repos were taking up space in the cloning section. This way we wouldn't schedule cloning jobs over the limit defined by the pipe size and clone percentage.

augur/tasks/init/celery_app.py
```diff
@@ -216,6 +215,10 @@ def setup_periodic_tasks(sender, **kwargs):
         logger.info(f"Scheduling update of collection weights on midnight on even numbered days.")
         sender.add_periodic_task(crontab(0, 0,day_of_month='2-30/2'),augur_collection_update_weights.s())
 
+        cloning_interval = 5*60
```
IsaacMilarky, line 218: I think adding the cloning as a periodic task will cause problems with cloning multiple things at once and with having multiple instances of this task active at once. The solution to this, in my opinion, is to make it so that the task schedules itself at the end of its loop when it runs out of repos to clone on a delay. That way, we would never run more than once instance of the task at once but we would still check for repos to clone every five minutes.

augur/tasks/git/facade_tasks.py
```diff
@@ -331,12 +333,50 @@ def git_repo_cleanup_facade_task(repo_git):
...
+            repoStatus = repo.collection_status[0]
+            setattr(repoStatus,"facade_status", CollectionState.UPDATE.value)
+            session.commit()
+
+            # get next repo
+            try:
+                repo_git = get_collection_status_repo_git_from_filter(session, is_pending, 1)[0]
+            except IndexError:
+                return
+
```
IsaacMilarky, line 377: I think this task should re-schedule itself for five minutes in the future and exit in the event that there aren't enough repositories to clone. This way we will always check to clone repos at least every five minutes and we avoid having multiple instances of this task running at once

IsaacMilarky (changes requested):
Mostly good except I don't like the approach of scheduling the clone task as part of a periodic celery task schedule. I think a better way of doing it would be to have the task re-schedule itself for five minutes in the future in the event of running out of repositories to clone.

augur/tasks/git/facade_tasks.py
ABrain7710: Multiple instances of the same task at once isn't an issue, and it is actually desirable behavior, because if we have enough repos to initialize that a clone task is taking 5 minutes then it would be good to have another task to help finish them faster

IsaacMilarky: I just worry about the undefined upper bounds of how many tasks we are scheduling. I think it does make sense to have more than one task working to clone repos at a time but I would prefer if the amount of tasks cloning were predefined or at least had an upper limit of some sort.

ABrain7710: Okay yeah that makes sense. I wasn't thinking about the case where the collection tasks are using all of the queue and the clone task has to wait until they are done to be processed

IsaacMilarky (approved):

PR #2365, opened 2023-04-28
Fix bug in repo clone task

ABrain7710:
**Description**
- Currently if one of the repos fails to clone the clone task errors, and the clone repos task is never scheduled again
- I added a custom exception that it raised when a clone fails, and if this error is raised, the clone repos task catches it and continues to the next repo
- I also added retry logic to the clone repos task. So if any other unpredictable errors such as network errors occur, it will retry in 5 minutes. Also these errors are logged to flower in the reties tab so we will still know if there are frequent errors happening in the task, that need to be patched

**Signed commits**


IsaacMilarky (approved):
I like this approach. In the future that custom exception could use a doc-string though

PR #2369, opened 2023-04-28
Collection schedule by user

IsaacMilarky:
**Description**
Add a mechanism to enforce a standard of fairness between collection instances with multiple users. This is done in two parts:

Part A (Handle repos that haven't been collected):

1. Split up all users that have new repos into 4 randomly assigned groups 
2. For each group:
     - query a set of new repos that belong to this group of users sorted by lowest weight first
     - Start repos up to the limit, break if the max_repo limit is reached

Part B (Handle repos that have been collected before):
1. Split up all users into 4 randomly assigned groups 
2. For each group:
     - query a set of old repos that belong to this group of users sorted by lowest weight first
     - Start repos up to the limit, break if the max_repo limit is reached

This mechanism is used for all collection hooks: core, secondary, and facade. 

@ABrain7710 and I developed this general algorithm over a meeting this week

**Signed commits**


augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+    split_user_list = split_list_into_chunks([row[0] for row in user_list], 4)
+
+    session.logger.info(f"User_list: {split_user_list}")
+
+    #Iterate through each fourth of the users fetched
+    for quarter_list in split_user_list:
+        if limit <= 0:
+            return
+
+        #Query a set of valid repositories sorted by weight, also making sure that the repos are new
```
sgoggins, line 479: So is this *only* going to run new repos?

ABrain7710: It will always run new repos first if they exist, and then if we didn't hit the max repo count when scheduling new repos we schedule more old repos until we hit the max repo count or if there aren't enough old repos (repos collected more than a day ago) we just schedule as many as we can

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+"""
+    Generalized function for starting a phase of tasks for a given collection hook with options to add restrictive conditions
+"""
+def start_repos_by_user(session, max_repo,phase_list, days_until_collect_again = 1, hook="core",new_status=CollectionState.PENDING.value,additional_conditions=None):
+
+    #getattr(CollectionStatus,f"{hook}_status" ) represents the status of the given hook
+    #Get the count of repos that are currently running this collection hook
+    desired_status = f"{hook}_status"
+    active_repo_count = len(session.query(CollectionStatus).filter(getattr(CollectionStatus,desired_status ) == CollectionState.COLLECTING.value).all())
+
```
ABrain7710, line 446: I think changing `desired_status` to `status_column` would be more clear.

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+            #Create conditions for this query of valid repos to make sure they are old enough and valid
+            collected_before = getattr(CollectionStatus,f"{hook}_data_last_collected" ) != None
+            make_sure_valid_repo = CollectionStatus.repo_id.in_(list(valid_repo_git_ids))
+            old_enough = getattr(CollectionStatus,f"{hook}_data_last_collected" ) <= cutoff_date
+            cond_list = [not_erroed, not_collecting,collected_before,old_enough,make_sure_valid_repo]
+
+
+            #Order by the relevant weight for the collection hook
+            order = getattr(CollectionStatus,f"{hook}_weight" )
+
```
ABrain7710, line 571: I think a `weight` would be more descriptive variable name, because `sort=order` doesn't make much sense to me

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+            JOIN augur_operations.user_repos ON augur_operations.user_groups.group_id = augur_operations.user_repos.group_id
+            JOIN augur_data.repo ON augur_operations.user_repos.repo_id = augur_data.repo.repo_id
+            JOIN augur_operations.collection_status ON augur_operations.user_repos.repo_id = augur_operations.collection_status.repo_id
+            WHERE user_id IN :list_of_user_ids AND {desired_status}='{str(new_status)}'
+            ORDER BY {hook}_weight
+            LIMIT :limit_num
+        """).bindparams(list_of_user_ids=tuple(quarter_list),limit_num=(limit*2))
+
+        #Get a list of valid repo ids, limit set to 2 times the usual
+        valid_repos = session.execute_sql(repo_query).fetchall()
```
ABrain7710, line 493: Why is the limit 2 times the usual amount here?

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+        valid_repo_git_ids = [repo[0] for repo in valid_repos]
+
```
ABrain7710, line 495: I think `valid_repo_ids` would be a more descriptive, `valid_repo_git_ids` isn't super clear to me, because the ids are of the repo not the `repo_git`

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+        collection_size = start_block_of_repos(
+            session.logger, session,
+            condition,
+            limit, phase_list, repos_type="new", hook=hook,sort=order
+        )
+        
+        #Update limit with amount of repos started
+        limit -= collection_size
+
+    #Now start old repos if there is space to do so.
```
ABrain7710, line 522: I believe there should still be a check here to see if the the limit is less than or equal to 0. To cover a case like this: The loop is in its last iteration and the limit is 5. Then 5 repos are started so the limit is 0. The loop then exits, so the limit check in the for loop is never reached to return. So it would continue on to start new repos even though the limit is 0

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+    random.shuffle(user_list)
+
+    #Extract the user id from the randomized list and split into four chunks
+    split_user_list = split_list_into_chunks([row[0] for row in user_list], 4)
+
+    for quarter_list in split_user_list:
+
+        #Break out if limit has been reached
+        if limit > 0:
+
```
ABrain7710, line 541: Returning if the limit is less than of equal to 0 just like you do above in the new repos scheduling is much easier to read

augur/tasks/util/collection_util.py
```diff
@@ -425,3 +433,155 @@ def start_block_of_repos(logger,session,condition,limit,phases,repos_type,hook="
...
+	            repo.repo_id
+                FROM augur_operations.user_groups 
+                JOIN augur_operations.user_repos ON augur_operations.user_groups.group_id = augur_operations.user_repos.group_id
+                JOIN augur_data.repo ON augur_operations.user_repos.repo_id = augur_data.repo.repo_id
+                JOIN augur_operations.collection_status ON augur_operations.user_repos.repo_id = augur_operations.collection_status.repo_id
+                WHERE user_id IN :list_of_user_ids AND {desired_status}='Success'
+                ORDER BY {hook}_weight
+                LIMIT :limit_num
+            """).bindparams(list_of_user_ids=tuple(quarter_list),limit_num=(limit*2))
+
```
ABrain7710, line 554: I think these queries should be factored out into a `get_user_list_repo_ids` function or something similar to that. Since they are doing the same thing they just have different statues.

ABrain7710 (changes requested):

augur/tasks/util/collection_util.py
IsaacMilarky: That makes sense

augur/tasks/util/collection_util.py
IsaacMilarky: Because I thought that it made sense to query more than the desired limit before applying the additional conditions of the second orm query that actually gets the repo_git. Looking back I think it makes more sense to scrap the second orm query altogether and just pass in a string object that concatenates all the conditions and binds them to the query.

augur/tasks/util/collection_util.py
IsaacMilarky: I agree. I'm nixing the redundant orm query

augur/tasks/util/collection_util.py
IsaacMilarky: It would return before it would start any repos. It would query the users table unnecessarily though

ABrain7710: Yeah, I didn't mean that it would start any repos, but I think we should add this check because it removes an unnecessary query like you mentioned and makes the code easier to read since it is easy to see how the code fulfills this comment "Now start old repos if there is space to do so."

IsaacMilarky: That makes sense

ABrain7710 (dismissed):

PR #2371, opened 2023-04-29
Fix scheduling of old repos

ABrain7710:
**Description**
- The secondary and facade collections were being recollected based on the `core_data_last_collected` date. Instead of the `secondary_data_last_collected` and `facade_data_last_collected`

**Signed commits**


sgoggins (approved):
LGTM

PR #2373, opened 2023-04-29
Fix clean collection status

ABrain7710:
**Description**
- The clean collection status was setting all facade statuses to `Update` if they were collecting. So I changed it so that it sets the status to `Update` if the repo hasn't been collected yet, and `Success` if the repo has been collected before. The goal of this is to reset it to the status it had before it started collecting.So if a repo is collecting and doens't have a collection date then its status was update beforehand, and if it is collecting and has a collection date it must be recollecting meaning that it had a status of 'Success' before. 

To fix this I updated the clean collection status to achieve this behavior, and I also added a check constraint to ensure that a repo is never set to update when it has a facade collection date 

**Signed commits**


sgoggins (approved):
LGTM.

PR #2394, opened 2023-05-03
make sure status of repo is updated when the cloning fails

IsaacMilarky:
**Description**
Fix issue where repos were not getting marked as failed when cloning them via git fails. 

**Signed commits**


sgoggins:
@ABrain7710 : I know it wasn't always working. I don't know if this patch solves that or not.

PR #2396, opened 2023-05-03
Change format of facade directories 

IsaacMilarky:
**Description**
- Change the repo_path that facade collection used to no longer be determined by the repo's group id, now a repo_id is used which is unique to that repo.
- Create an alembic script to automatically migrate instances to the new change
- Fix issues associated with the aforementioned change due to changes in formatting

**Signed commits**


sgoggins (dismissed):
This LGTM .. with the assumption that something in here is deleting the old clone, and making a new one ... or moving the old one. Either way, we can't afford to have the old clones and the new clones both on disk. It would be too costly for augur users, and not something they expect.

IsaacMilarky:
> This LGTM .. with the assumption that something in here is deleting the old clone, and making a new one ... or moving the old one. Either way, we can't afford to have the old clones and the new clones both on disk. It would be too costly for augur users, and not something they expect. 

The old clones are deleted and then cloned again via an alembic script.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -132,7 +132,9 @@ def create_endpoint_from_commit_sha(logger,db,commit_sha, repo_id):
 
     # Else put into a more readable local var
     #session.logger.info(f"Result: {result}")
-    repo_path = result.repo_path.split("/")[1] + "/" + result.repo_name
+
+    split_git = result.repo_git.split('/')
+    repo_path = split_git[-2] + "/" + result.repo_name
```
ABrain7710, line 138: I think this is parsing out the org / owner name so the variable name should be org or owner for clarity.

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
```diff
@@ -0,0 +1,73 @@
...
+
+    conn.execute(text(f"""
+
+    UPDATE augur_operations.collection_status
+    SET facade_status='Pending', facade_task_id=NULL, facade_weight=NULL,commit_sum=NULL,facade_data_last_collected=NULL;
+
+    UPDATE repo
+    SET repo_path=NULL,repo_name=NULL;
+    """))
+
```
ABrain7710, line 38: After thinking on this, I think it would be better if we just wrote a function in this alembic script that creates the new path with the repo id and path identifier. Then used a mv command to move the clone to the new directory. We should be able to do this because we already have the location of the facade directory, and we can tell which directories are old because they are directories that are just an integer (repo_group_id) rather than the new format which is an integer-string (repo_id-repo_path). I think this would make more sense to existing users, and all the facade repos wouldn't have to go back through the process of moving through statues (this isn't a huge deal it just clogs up facade for new repos until of these reset ones make it through which isn't ideal)

ABrain7710 (changes requested):

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
IsaacMilarky: That makes sense

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
IsaacMilarky: I didn't want to do this because we want to calculate the weight for old repos anyways and its not that much more to ask to add the cloning overhead to that weight calculation since it happens in the clone_repos method. Also depending on the configuration that people have on their instances moving each repo could take a long time which is unexpected for an alembic script.

sgoggins: @ABrain7710 : So, what do you think?

ABrain7710: @IsaacMilarky / @sgoggins What Isaac mentioned makes sense. If we are taking this approach with old repos for facade I think we should also update the core and secondary tasks so we ensure they compute the issue_pr_sum. To make sure we don't reset every repo on padres though we should only update core and secondary tasks to pending if the issue_pr_sum is NULL.

ABrain7710 (changes requested):
The only thing I see that needs added is a query that updates the core and secondary statues to pending if the issue_pr_sum is NULL

IsaacMilarky:
> The only thing I see that needs added is a query that updates the core and secondary statues to pending if the issue_pr_sum is NULL

That makes sense. I have made the change

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
```diff
@@ -0,0 +1,77 @@
...
+    UPDATE augur_operations.collection_status
+    SET facade_status='Pending', facade_task_id=NULL, facade_weight=NULL,commit_sum=NULL,facade_data_last_collected=NULL;
+
+    UPDATE repo
+    SET repo_path=NULL,repo_name=NULL;
+
+    UPDATE augur_operations.collection_status
+    SET core_status='Pending', secondary_status='Pending', core_data_last_collected=NULL,secondary_data_last_collected=NULL
+    WHERE issue_pr_sum IS NULL;
+    """))
```
ABrain7710, line 41: Just to be safe I think we should also set the task ids to NULL

ABrain7710 (changes requested):

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
IsaacMilarky: Made the change

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def total_facade_reset():
+    conn = op.get_bind()
+
+    conn.execute(text(f"""
+
+    UPDATE augur_operations.collection_status
+    SET facade_status='Pending', facade_task_id=NULL, facade_weight=NULL,commit_sum=NULL,facade_data_last_collected=NULL;
+
```
ABrain7710, line 34: I think we should only run this on repos that don't have a commit_count or have a repos that are stored in the old structure. The reason being that I would like this script to reset all of facade on a huge instane if someone accidentally downgrades the database

ABrain7710 (changes requested):
I think we should try to restrict these operations rather than just running them on every repo, so that if someone runs a downgrade/updgrade in the future it won't just reset everything for no reason

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
IsaacMilarky: If they wanted to downgrade the schema it would be proper to also reset the facade repos again. We also can't check based on the commit_count because the revision before this one has commit_count without the directory change. I think it's fine how it is in my opinion.

IsaacMilarky: I don't think it matters if someone accidentally clears their facade repos while downgrading the schema since it's so fast to clone.

ABrain7710: Reply to first part: How would it be proper? I don't see why that would be the desired behavior. Usually in my experience downgrades are to bring the database back to the prior state, but resetting the repos doesn't do anything to reset to the old state. 

Reply to second part: I think we can do it based on commit_count we would run it if the facade directory was wrong or the commit_count doesn't exist since they are both valid times to do it.

I do think it matters, because although it is fast to clone, each repo still has to go through the collecting phase and if there are 22,000 repos that go reset and there collections finished instantly (since the commits exist) it would take around 6 hours for all the repos to get back through to success and this is not even considering the time to add all the tasks to the queue, and the time to clone.

ABrain7710 (approved):

augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
IsaacMilarky: > I do think it matters, because although it is fast to clone, each repo still has to go through the collecting phase and if there are 22,000 repos that go reset and there collections finished instantly (since the commits exist) it would take around 6 hours for all the repos to get back through to success and this is not even considering the time to add all the tasks to the queue, and the time to clone.

The only issue with this is that it sets all the repos to new but otherwise it doesn't really change the underlying behavior of facade. I do agree that people could see that as a problem though since people want to prioritize the newly added repos.

ABrain7710: I approved because it does solve the problem. I'm just a bit apprehensive about doing this without conditions because it may not do what people would expect it to do when it is run multiple times

IsaacMilarky: > I approved because it does solve the problem. I'm just a bit apprehensive about doing this without conditions because it may not do what people would expect it to do when it is run multiple times

I see the concern but I think any issue this alembic script might cause would be more of an inconvenience caused by not knowing facade repos are all being deleted. If we want to we can add a confirm prompt so that people are absolutely aware that their facade directories are being deleted. If we want to we can also mark the old repos or something so that they aren't re-collected as new.

PR #2400, opened 2023-05-09
Migration patch

IsaacMilarky:
**Description**
Correct use of pathlib to use mkdir instead of touch 

**Signed commits**


augur/application/schema/alembic/versions/18_schedule_any_old_facade_repositories_to.py
```diff
@@ -55,7 +55,7 @@ def total_facade_reset():
 
         shutil.rmtree(path)
         #Create path
-        path.touch()
+        path.mkdir()
```
sgoggins, line 58: Yes, that will work.

sgoggins (approved):

sgoggins (approved):

sgoggins (approved):

PR #2404, opened 2023-05-10
No valid Github API keys fix

ABrain7710:
**Description**
- I came across these issues when I saw the frontend flash that a repo was added, but it actually wasn't. The issue was that I had no valid Github API keys. To fix this issue, I made the Github API key handler raise a `NoValidKeysError` when there are no valid github api keys. This improves the case when all the Github API keys are expired or invalid, because it now raises an exception rather than just silently not working. 
- Now in the add repo and add org code it catches NoValidKeysError and returns False so the frontend shows that the repos weren't added
- I also came across an issue in the view code, where it wasn't unpacking a tuple before checking if the repo or org was added. This issue is why we were always see it flash that the repos were added even it they really weren't

**Notes for Reviewers**
There are two imports of the `NoValidKeysError` in `augur_operations.py` that are each inside of a function, this is because adding the import at the module level, creates a circular dependency (it has proved difficult to not get circular dependencies in the model files so we have been doing imports in the functions most of the time)

**Signed commits**

### DIFF
--- a/augur/api/view/api.py
+++ b/augur/api/view/api.py
@@ -40,14 +40,14 @@ def av_add_user_repo():
 
         # matches https://github.com/{org}/ or htts://github.com/{org}
         if Repo.parse_github_org_url(url):
-            added = current_user.add_org(group, url)
+            added = current_user.add_org(group, url)[0]
             if added:
                 added_orgs += 1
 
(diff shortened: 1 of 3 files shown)


IsaacMilarky (approved):
LGTM

PR #2405, opened 2023-05-10
Handle GitHub API 204 response

ABrain7710:
**Description**
- This endpoint, `https://api.github.com/repos/{owner}/{repo}/contributors?state=all`, which is used in `grab_commiters` returns a `204 No Content` status and an empty string when there are not contributors. This was causing the response to be processed as a string, and a warning was logged because it was unable to be processed. Now if a `204 No Content` status is returned by the Github API then all an empty list is returned with a status of `Success`

**Signed commits**


IsaacMilarky (approved):

sgoggins (approved):
Could we get an empty data return as a GitHub error and then get data with a retry in this case?  If so should we do at least one retry?

PR #2407, opened 2023-05-11
Scale celery memory use to system's available memory

IsaacMilarky:
**Description**
- Reduce/Increase the amount of processes celery is allowed to use depending on the amount of available memory on the system. Data is gathered with psutil
- Hard cap the maximum amount of processes for each worker process to the previous default values

**Notes for Reviewers**
These changes should help to preserve some memory on systems with smaller amounts of it without hampering collection on larger memory instances. 

**Signed commits**

### DIFF
--- a/augur/application/cli/backend.py
+++ b/augur/application/cli/backend.py
@@ -145,7 +145,8 @@ def start_celery_worker_processes():
     #Calculate process scaling based on how much memory is available on the system in bytes.
     #Each celery process takes ~500MB or 500 * 1024^2 bytes
 
-    available_memory_in_bytes = psutil.virtual_memory().available
+    #Cap memory usage to 30% of total virtual memory
+    available_memory_in_bytes = psutil.virtual_memory().total * .3
     available_memory_in_megabytes = available_memory_in_bytes / (1024 ** 2)
     max_process_estimate = available_memory_in_megabytes // 500
 
--- a/augur/application/cli/backend.py
+++ b/augur/application/cli/backend.py
@@ -156,6 +156,7 @@ def start_celery_worker_processes():
 
     #2 processes are always reserved as a baseline.
     scheduling_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=2 -n scheduling:{uuid.uuid4().hex}@%h -Q scheduling"
+    max_process_estimate -= 2
 
     core_num_processes = determine_worker_processes(.6, 45)
     #60% of estimate, Maximum value of 45
@@ -165,7 +166,7 @@ def start_celery_worker_processes():
     #20% of estimate, Maximum value of 25
     secondary_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency={secondary_num_processes} -n secondary:{uuid.uuid4().hex}@%h -Q secondary"
 
-    facade_num_processes = determine_worker_processes(.15, 20)
+    facade_num_processes = determine_worker_processes(.2, 20)
     #15% of estimate, Maximum value of 20
     facade_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency={facade_num_processes} -n facade:{uuid.uuid4().hex}@%h -Q facade"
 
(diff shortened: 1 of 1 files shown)


augur/application/cli/backend.py
sgoggins, line 1: I think this is good. Does celery reserve only a percentage of available memory. We wouldn’t want it to take it all.

sgoggins (commented):
Let me know what the max percent of available memory celery will reserve. I’m guessing ~20% makes sense. We need to assume database and rabbitmq need mememory tooo. I just am not sure what this does total percentage wise.

sgoggins (commented):
Let me know what the max percent of available memory celery will reserve. I’m guessing ~20% makes sense. We need to assume database and rabbitmq need mememory tooo. I just am not sure what this does total percentage wise.

augur/application/cli/backend.py
IsaacMilarky: It for sure won't take up all of the memory. I don't know the exact proportion given the custom celery settings we have.

sgoggins: Can we add some checks to ensure there’s some kind of cap? Is the underlying issue that celery isn’t grabbing enough memory?

IsaacMilarky: I just added a flat 30% cap of the total memory to keep it simple.

augur/application/cli/backend.py
```diff
@@ -162,6 +140,50 @@ def start(disable_collection, development, port):
...
+    available_memory_in_megabytes = available_memory_in_bytes / (1024 ** 2)
+    max_process_estimate = available_memory_in_megabytes // 500
+
+    #Get a subset of the maximum procesess available using a ratio, not exceeding a maximum value
+    def determine_worker_processes(ratio,maximum):
+        return min(round(max_process_estimate * ratio),maximum)
+
+    #2 processes are always reserved as a baseline.
+    scheduling_worker = f"celery -A augur.tasks.init.celery_app.celery_app worker -l info --concurrency=2 -n scheduling:{uuid.uuid4().hex}@%h -Q scheduling"
+
```
ABrain7710, line 159: I think 2 needs to be subtracted from `max_process_estimate` otherwise on small systems more than 30 percent of virtual memory will be used.

ABrain7710 (changes requested):
I really like this. I just think we should subtract 2 from the `max_process_estimate` to make it more accurate

augur/application/cli/backend.py
IsaacMilarky: That makes sense. I made the change.

augur/application/cli/backend.py
```diff
@@ -162,6 +140,51 @@ def start(disable_collection, development, port):
...
+    #Cap memory usage to 30% of total virtual memory
+    available_memory_in_bytes = psutil.virtual_memory().total * .3
+    available_memory_in_megabytes = available_memory_in_bytes / (1024 ** 2)
+    max_process_estimate = available_memory_in_megabytes // 500
+
+    #Get a subset of the maximum procesess available using a ratio, not exceeding a maximum value
+    def determine_worker_processes(ratio,maximum):
+        return min(round(max_process_estimate * ratio),maximum)
+
```
ABrain7710, line 156: We need to either add a min of 1 to this, or return an error if that the system does not have enough memory to run Augur because a system with 8 GB of RAM with Augur configured to use 30 percent of virtual memory will currently run 2 scheduling processes, 1 core process, 0 secondary processes, and 0 facade processes. Note: Around 8.2 GB with a 30 percent usage is what is needed to for 2 scheduling processes and 1 of all the others. At 20 percent usage 12.3 GB is needed

ABrain7710 (changes requested):

augur/application/cli/backend.py
IsaacMilarky: I think we should bump it up to at least 40% and then add a minimum process of 1.

IsaacMilarky: We already have a hard limit on the maximum level of processes each worker can have for larger instances. I think for now a default cap of 20-30% would be fine. I think it is a good idea to put the percentage in the config though.

ABrain7710: Yeah I didn't notice that before, this sounds good to me

ABrain7710 (approved):

PR #2409, opened 2023-05-15
Add config options for memory usage and frequency of refresh materialized views

IsaacMilarky:
**Description**

- Add default options to the config so users can set caps on celery worker memory usage as well as how often materialized views are refreshed. 

    - Default values are 7 days to refresh the materialized views and 0.25 (25%) for the virtual memory cap
- Add alembic script to handle updating existing configs to have the new options

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -66,7 +66,9 @@ default_config = {
                 "log_level": "INFO",
             },
             "Celery": {
-                "concurrency": 12
+                "concurrency": 12,
+                "worker_process_vmem_cap": 0.25,
+                "refresh_materialized_views_interval_in_days": 7
             },
             "Redis": {
                 "cache_group": 0, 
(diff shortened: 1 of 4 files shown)


augur/application/config.py
```diff
@@ -66,7 +66,9 @@ def get_development_flag():
                 "log_level": "INFO",
             },
             "Celery": {
-                "concurrency": 12
+                "concurrency": 12,
+                "worker_process_vmem_cap": 0.25,
+                "refresh_materialized_views_interval_in_days": 7
             },
```
ABrain7710, line 72: I don't think we still need the `concurrency` config option, since we have `worker_process_vmem_cap` now

IsaacMilarky: You're right I didn't notice that

sgoggins (approved):

PR #2410, opened 2023-05-17
Increase security of password hashing

ABrain7710:
**Description**
- Changed the hashing algorithm to sha512 so it is less vulnerable to brute force attacks
- Increased the salt length to 32 (it actually was using salt before it was only 8 long though) to add more variability to the hashes
- Defined a standard method to get a password hash, so all the augur code uses the same security parameters

**Signed commits**


sgoggins (approved):

Ulincsys (approved):
LGTM

PR #2411, opened 2023-05-17
Implement asynchronous processing for adding orgs and repos

ABrain7710:
**Description**
- Defined a task that takes a list of repo and org urls that returns a list of invalid urls, added orgs, and added repos
- This now happens in the background because gunicorn just adds a task to the celery task queue
- Once the task completes all the repos show up in the user's group
- Updated backend.py to start only the frontend worker when collection is disabled

**Signed commits**

### DIFF
--- a/augur/api/view/api.py
+++ b/augur/api/view/api.py
@@ -1,6 +1,7 @@
 from flask import Flask, render_template, render_template_string, request, abort, jsonify, redirect, url_for, session, flash
 from flask_login import current_user, login_required
 from augur.application.db.models import Repo
+from augur.tasks.frontend import add_org_repo_list
 from .utils import *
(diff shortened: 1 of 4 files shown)


augur/api/view/api.py
```diff
@@ -1,6 +1,7 @@
 from flask import Flask, render_template, render_template_string, request, abort, jsonify, redirect, url_for, session, flash
 from flask_login import current_user, login_required
 from augur.application.db.models import Repo
+from augur.tasks.frontend import add_org_repo_list
```
sgoggins, line 4: There is no augur.tasks.frontend file or directory @IsaacMilarky / @ABrain7710  .... so boot is failing.

PR #2424, opened 2023-05-23
Protect against sql injection

ABrain7710:
**Description**
- Make queries paramaterized so they are secure against sql injection

**Signed commits**


sgoggins (approved):
I trust this is ok and tested.

IsaacMilarky (approved):
Tested at this morning's meeting

PR #2455, opened 2023-06-28
Send mail

Ulincsys:
**Unrelated changes in this branch**

- Modify config installation script to source the `.env` file before asking for API key information
  - This allows the admin to avoid needing to enter their credentials for every reinstallation

**Description**

- Add user email verification page
  - Give logged in users the ability to update email for verification
  - Add associated backend endpoint for verifying One Time Passcodes
- Add user password reset page
  - Password reset requires email to be verified
- Add service CLI
  - Only service currently available is email
  - Only email service provider currently available is SendGrid
- Add email service file
  - Service configuration allows for multiple configurable providers to be accessible through a common interface

### TODO

No abuse mechanism detection currently exists. We **must** limit the number and frequency of which a user can initiate an email request from the server. Thoughts?

**Signed commits**


sgoggins:
@Ulincsys : This is still labeled "in progress". Is it ready to be merged? It should probably have dev merged into it and then get tested at this point.

Ulincsys:
We need abuse prevention mechanisms integrated into this before it can go live. The mail service integration works, but as it stands now a user can just force Augur to send unlimited mail requests.

PR #2462, opened 2023-07-06
Remove unused legacy code

Ulincsys:
Signed-off-by: Ulincsys <ulincsys@gmail.com>

**Description**
- Remove unused and extraneous legacy code from frontend
- Resolve overlooked dependency on a specific external server

**Signed commits**


augur/api/view/utils.py
```diff
@@ -209,67 +126,6 @@ def toCacheFilepath(endpoint, append = True):
...
-            cache_files_requested.remove(filename)
-        return data
-    except Exception as err:
-        logger.error("An exception occurred while fulfilling a json request")
-        logger.error(err)
-        return False, str(err)
-
-""" ----------------------------------------------------------------
-"""
-def requestPNG(endpoint):
```
sgoggins, line 258: Don't the Visualization endpoints rely on this method?

sgoggins (commented):
I understand what this PR is aiming to accomplish. What I am not certain about is whether or not we really intend to remove all the things that are removed, and if they will possibly break other things.

augur/api/view/server/User.py
```diff
@@ -1,276 +0,0 @@
-from flask_login import UserMixin
```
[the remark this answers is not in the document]
Ulincsys: User.py was the view-side wrapper for the User endpoints on the backend. Since the frontend and the backend now exist in the same app, this wrapper is no longer used.

augur/api/view/utils.py
Ulincsys: No, the visualization endpoints do not rely on any methods in the fontend code.

If you are talking about the report page which uses the visualization endpoints, those use a different function for downloading and caching the images.

sgoggins (approved):

PR #2468, opened 2023-07-10
[WIP] [API] Paginate SQL queries

Seltyk:
**Description**

When finished, this PR will support and enforce pagination on SQL queries. At time of writing, only the `/api/unstable/repos` endpoint features this change.

Two URL parameters are accepted for pages: `page` contains the page number, defaulting to 1 if not provided, and `count` contains the number of elements (repos) per page, defaulting to 20 if not provided. Postgres provides `LIMIT` and `OFFSET` options for its `SELECT` queries which is used here to generate paginated responses.

- A naive defense against malicious URLs is present in the form of casting `request.args.get("foo")` from `str` to `int`, but I do not trust this fully enough for a production setting. The input of more experienced web developers would be greatly appreciated for this.
- It is not obvious to me how these changes could be factored out of the `get_all_repos` function and applied to other endpoints, such as `get_all_repo_groups` and `get_repos_in_repo_group`. Input is more than welcome.
- Being that this project belongs not to me but to CHAOSS, changes that their developers wish to see in the API design are important for me to consider, and are of course welcome in this thread.

[The `augur` branch of Mystic Evoker](https://opensource.ieee.org/rit/mystic-group/mystic-evoker/-/tree/augur?ref_type=heads) already contains support for these changes.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/api/routes/util.py
+++ b/augur/api/routes/util.py
@@ -16,6 +16,12 @@ from augur.application.config import AugurConfig
 
 logger = AugurLogger("augur").get_logger()
 
+# Magic number configuration for pagination
+PAGE_DEFAULT = 1
+PAGE_MINMAX = (1, 4000)
+COUNT_DEFAULT = 20
+COUNT_MINMAX = (1, 50)
+
 def get_args(args: MultiDict, defaults: Dict[str, int]) -> Dict[str, int]:
     """
     Read from the provided `request.args` object, based on defaults.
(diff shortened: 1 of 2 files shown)


sgoggins (approved):
LGTM

sgoggins:
@Seltyk : Let me know when you think this is ready.

Seltyk:
It *could* be called done now, but the `get_args` type hints contain a really nasty footgun (i.e. forcing every argument to be an `int`) which limits future scalability. Were we using a statically-typed language, having `get_args` accept a type argument would solve this issue, but since this is a Python project, some other solution probably needs to be found instead (unless Python has some weird implementation of generic types I can leverage). The other big problem is that input sanitization begins and ends at casting the `str` parameters to `int`s. Better than nothing, but it feels like a bad idea.

This whole PR was made under the assumption that the code written for it would only be used for the PR itself. That was a **terrible** assumption to make and I'm very disappointed in myself.

tl;dr -- this can be merged now, because it already works, or I could take another crack at cleaning up my bad code. The obvious problem, of course, is that I don't know how soon I'll be able to do that. Maybe tomorrow, maybe September.

PR #2471, opened 2023-07-18
Ml patch mel

sgoggins:
**Description**
Adding machine learning worker updates to the dependency update branch.


sgoggins (commented):
Checked.

PR #2472, opened 2023-07-20
[docs] Clarify OAuth flow documentation

Seltyk:
**Description**
- add note about using AUGUR_DEV envvar to allow localhost clients (i.e. to disable SSL enforcement)
- remove Augur View and docs terminology disparity
- fix "acquired" typo

**Notes for Reviewers**

In my opinion, the SSL issue should be met with a backend change. Assuming Augur is running with HTTPS support, enforcing SSL *into* Augur is a good idea, but enforcing it *out of* Augur appears to be why localhost clients don't work. The redirect back to the user's client application at http://127.0.0.1 fails SSL enforcement silently, responding with (non-existent) temporary authorization code "undefined". Should I make a GitHub issue for this?

**Signed commits**

### DIFF
--- a/docs/source/login.rst
+++ b/docs/source/login.rst
@@ -67,7 +67,7 @@ The Client Application must make the following request in order to facilitate th
     URL: https://augur.example.com/api/unstable/user/session/generate
     arguments:
         code: [the temporary authorization code]
-        grand_type: "code"
+        grant_type: "code"
     headers:
         Authorization: Client [your client secret]
 
@@ -101,7 +101,7 @@ The application may also attempt automatic reauthorization using the previously 
     URL: https://augur.example.com/api/unstable/user/session/refresh
     arguments:
         refresh_token: [the previously provided refresh token]
-        grand_type: "refresh_token"
+        grant_type: "refresh_token"
     headers:
         Authorization: Client [your client secret]
 
--- a/docs/source/login.rst
+++ b/docs/source/login.rst
@@ -6,6 +6,10 @@ Augur implements the Oauth 2.0 specification, and each Augur instance is capable
 Prerequisites
 --------------
 
+If your Augur instance is running behind Nginx or Apache, make sure this parameter (or its Apache equivalent) is set in your ``sites-enabled`` configuration::
+
+    proxy_set_header X-Forwarded-Proto $scheme;
+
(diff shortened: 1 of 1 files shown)


Ulincsys:
The doc updates look good to me. I'm not sure I understand what the SSL issue is you've mentioned? I'm able to authenticate with a client locally from an instance of augur served behind HTTPS.

Seltyk:
The Jinja template `authorization.j2` gives Flask the redirect target to handle temporary auth code generation to then be handed back to Jinja. From what I can tell, if that target lacks SSL (because, say, the redirect is to `http://localho.st`), [the endpoint rejects immediately](https://github.com/chaoss/augur/blob/dev/augur/api/routes/user.py#L74) and Jinja is left with a `{"status": "SSL required"}` object when [it expects an object with a `code` field](https://github.com/chaoss/augur/blob/dev/augur/templates/authorization.j2#L39). No `code` field means `data.code` is `undefined`.

Ulincsys:
The application's redirect URL is not considered as part of the request to generate a temporary auth code. As long as the user agent makes the request to the backend with https as the scheme, then the requirement for SSL should be met (assuming that the certs have been setup on the host).

If the instance of Augur you're querying is hosted behind a server like Nginx or Apache, you'll additionally need to make sure you have the following parameter set in your `sites-enabled` configuration under the location which targets the backend:
```
proxy_set_header        X-Forwarded-Proto           $scheme;
```

Seltyk:
Further testing reveals I'm actually getting a 426 from `/user/authorize` when I click the "authorize" button, which is not resolved by the nginx config you mentioned. `AUGUR_DEV=1` does, however, which strongly implies there is some SSL shenanigans happening with my frontend. I will tweak my PR when the solution is found.

Seltyk:
4 days of confusion down to one little typo. Fixed in the latest commit.

Ulincsys:
If you have some time on Wednesday or Thursday @Seltyk I think it might help to meet via Zoom to resolve the SSL issue.

Seltyk:
I'm available until 16:00 EDT today and until 18:00 EDT tomorrow. GitHub says you're in CDT, so that should be easy to work out.

Ulincsys:
Let's plan for 3pm CDT Thursday. My work email should be visible on my profile page, if you'd like to reach out there we can arrange the meeting link.

Ulincsys (approved):

PR #2474, opened 2023-08-01
[auth] Fix session token refresh

Seltyk:
**Description**
Refreshing session tokens first loads the authenticated client application from a SQL query, then verifies it against the user's browser session, and finally refreshes the token. The verification step uses a simple inequality check between two `ClientApplication` objects. Because this class does not have a definition for (in)equality (i.e. the `__eq__` and `__ne__` methods are not implemented), that inequality only checks the objects' pointers/addresses*, which is guaranteed to fail. Thankfully, the `ClientApplication` class has an `id` field which matches the primary key in the associated database row. Comparing the underlying hex string values should make correct equality checks. This PR's main change is a simple implementation of `__eq__` for `ClientApplication` as described. Because `__ne__` is not implemented, inequality defers to this method and inverts the output, so both `==` and `!=` are well-defined.

*I presume, based on zero evidence

This request also makes some nitpick changes, listed here:

- ghostbytes in both modified files is removed, including on lines matching the regex `/^\s+$/`
- unused imports and variables are removed (resolving a handful of [Ruff](https://beta.ruff.rs/) lints)
- a module-level import is moved to the top of the file (resolving another lint)
- `value == False` and `value == True` are replaced (resolving the last lint)
- docstring typo "dictionairy" is resolved
- token refresh failures return code 400 instead of 200
- "coding" replaced with "encoding"

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/api/routes/user.py
+++ b/augur/api/routes/user.py
@@ -127,10 +127,10 @@ def refresh_session(application):
 
         refresh_token = session.query(RefreshToken).filter(RefreshToken.id == refresh_token_str).first()
         if not refresh_token:
-            return jsonify({"status": "Invalid refresh token"})
+            return jsonify({"status": "Invalid refresh token"}), 400
 
         if refresh_token.user_session.application != application:
-            return jsonify({"status": "Invalid application"})
+            return jsonify({"status": "Invalid application"}), 400
 
         user_session = refresh_token.user_session
         user = user_session.user
@@ -424,7 +424,7 @@ def toggle_user_group_favorite():
     Returns
     -------
     dict
-        A dictionairy with key of 'status' that indicates the success or failure of the operation
+        A dictionary with key of 'status' that indicates the success or failure of the operation
     """
     group_name = request.args.get("group_name")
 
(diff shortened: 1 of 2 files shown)


Ulincsys:
The equality method of `ClientApplication` should check that the `other` object is of the same class type as `self` before performing the ID comparison, and return `False` if it is not. Currently, if one were to try comparing it against an object without the `id` attribute, it would throw an exception.

Seltyk:
> `ClientApplication` is a subclass of the SQLAlchemy `declarative_base` class, which provides the functionality observed

At a glance, the `declarative_base` call chain never defines equality for its objects, and the docs don't mention equality or the `__eq__` method. That said, if it does define this equality, I'm getting `{"status": "Invalid application"}` when in theory I shouldn't.

> The conditional in this instance returns a callable for use in `filter()`, it does not resolve to a boolean.

There were two `== Bool` instances. One is `checkPassword`, which is the result of a function that returns a bool, so this one should be no issue. The other, which I think is what you mentioned, is `UserGroup.favorited`. I had to stare at it for a few minutes here, but I see what you're getting at now. I'll file a bug report with Ruff and revert that change.

Ulincsys (approved):

PR #2476, opened 2023-08-07
[api] Add endpoint to get repo by ID

Seltyk:
Documentation also written for this, though I'm not 100% sure "object" is a valid type. It definitely doesn't return an array.

**Description**
- Adds a utility endpoint to query a single repository. Contains much the same information as the /repos endpoint, but instead of returning an array of repositories, only one is returned. If the SQL read fails or no repository with the given ID is found, returns a 400.

I am told this functionality used to exist by using the special repo group ID 9999, and code using this feature [is still in Augur](https://github.com/chaoss/augur/blob/main/augur/tasks/data_analysis/insight_worker/tasks.py#L62), but that endpoint has been a 404 for as long as I have known Augur.

As usual, I also cleared linter warnings along the way (module-level import not at top of file and unused import) and removed ghostbytes.

**Notes for Reviewers**

The schema type is marked as `array` not because this endpoint actually returns an array -- it doesn't -- but because `object` wouldn't render the schema. For the first time, I had to document an error in the docs in the docs instead of fixing the error.

**Signed commits**

### DIFF
--- a/augur/api/routes/util.py
+++ b/augur/api/routes/util.py
@@ -1,20 +1,17 @@
 #SPDX-License-Identifier: MIT
+from augur.api.routes import AUGUR_API_VERSION
+from ..server import app, engine
 import base64
 import sqlalchemy as s
 import pandas as pd
 import json
 from flask import Response
-import logging
 
 from augur.application.db.session import DatabaseSession
 from augur.application.logs import AugurLogger
 from augur.application.config import AugurConfig
 
 logger = AugurLogger("augur").get_logger()
-
-from augur.api.routes import AUGUR_API_VERSION
-from ..server import app, engine
-
 
 @app.route('/{}/repo-groups'.format(AUGUR_API_VERSION))
 def get_all_repo_groups(): #TODO: make this name automatic - wrapper?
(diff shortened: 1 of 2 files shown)


Seltyk:
Further docs testing shows that if I set `type: array` the response schema looks fine but is described as an array. If I set `type: object` it doesn't show at all. The OpenAPI spec says that the schema object is a superset of JSON Schema, which defines both "array" and "object" as instance types. Interestingly, Augur's `spec.yml` doesn't appear to fit the OpenAPI documentation, but if that were the case, surely the docs wouldn't build correctly?

Seltyk:
Another thought: this would make `/repos/:id` return more data per repo than `/repos`. This isn't dangerous, but it does mean inconsistency, which is not pleasant for frontends to parse. One of them ought to change.

sgoggins:
@Seltyk : What kind of environment did you test this against? I'm trying to decide if I should merge into dev, or create a different branch and take it for a spin soon one way or another.

Seltyk:
You might want to take it for a spin first, but the only *functional* change I made was adding an endpoint, so in theory it shouldn't have any issues going straight to dev

sgoggins (approved):
LGTM

PR #2477, opened 2023-08-08
Update sqlalchemy to 2.0.17 and adjust Augur accordingly 

IsaacMilarky:
**Description**
- Remove redundant orm relationship definitions in augur_data.py and augur_operations.py
- Update method of converting records to python dicts 
- Update method of inserting and updating data using .begin() method instead of .connect() method as raw sql statements are no longer autocommit by default in sqlalchemy 2.x
-Update alembic scripts to work with sqlalchemy 2.x

**Notes for Reviewers**
- Certain tests will have to be updated with new sqlalchemy autocommit rules

**Signed commits**


sgoggins (dismissed):
I'm going to test this against a copy of dev, just to make sure the API's are all solid.

sgoggins:
@IsaacMilarky : Should I merge this into a copy of dev and test yet?

IsaacMilarky:
Not yet I need to upgrade the pandas version and test it.

sgoggins:
> Not yet I need to upgrade the pandas version and test it.

Don't be surprised if updating the pandas version requires you to then upgrade other libraries. I have tried "doing them all" and it causes a lot to break ... mostly because of the SQLAlchemy fix issues. 


You *might* consider trying the setup.py (and setup.py for all the data_analysis tasks) from this branch: 

https://github.com/chaoss/augur/tree/dev-sqlalchemy-lib-updates

I kept all the specified versions in comments, but basically took all the versions out ... DB stuff is all that seems to *not* work ... It would sort of be the "extreme" approach ... I could also merge your branch into it and back out the updates you made to setup.py ... 

In fact, why don't I do that just in case it works. It won't hurt anything because the branch is dead without the SQLAlchemy updates.

PR #2495, opened 2023-08-24
Split ENTRYPOINT script in 2

mscherer:
ENTRYPOINT is made to be used when the container is running one specific executable (think bash, or python) and is hard to override. However, when using the container, it is sometime needed to execute specific subcommand, such as augur db upgrade-db-version

**Signed commits**

### DIFF
--- a/docker/backend/init.sh
+++ b/docker/backend/init.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#SPDX-License-Identifier: MIT
+set -e
+
+if [[ "$AUGUR_DB_SCHEMA_BUILD" == "1" ]]; then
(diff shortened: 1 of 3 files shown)


docker/backend/init.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
sgoggins, line 1: @ABrain7710 : do we support this way of loading anymore, or does it need to be depricated?

sgoggins (commented):
@mscherer : I have one question for one of our other maintainers I asked in context. Waiting for an answer.

ABrain7710 (approved):

PR #2497, opened 2023-08-25
Value worker migrate to task

IsaacMilarky:
**Description**
Migrate the old value worker into the new version of augur as a task called process_scc_value_metrics. This task does the same thing the old value worker did by using https://github.com/boyter/scc to gather data about files in a repo by language. 

Also, alter the unique constraint on the repo_labor table to no longer be initially deferrable. This was done in order to have postgres ON CONFLICT logic work with the constraint which it didn't before altering it. 

This new task is run as part of the facade tasks because it depends on the repository being cloned for it to work.

This new task also automatically clones and builds https://github.com/boyter/scc in the user's home directory in order to make it able to be used by augur in the same way that we do with ossf scorecard.  

**Signed commits**


sgoggins (approved):
testing on a small db

PR #2504, opened 2023-08-28
Updated key rotation to be more random

sgoggins:
**Description**
- Use of API keys was not being rotated, and so jobs were getting stuck.
- This PR fixes the glitch.
- This PR now also fixes an old issue with how the commit_count was retrieved by the repo_info task. Previously we selected a `main` branch. The new code gets the `default branch`. I have verified that where commits were getting stored in the metadata as NULL in ~25% of cases, and there are 5-10% of repos in any augur collection where `main` is not the default branch. Using the default branch always get commit_count data correctly. NO more NULL values. No more counting the "not default branch".
- This PR now also fixes the refreshability of materialized views by adding unique keys on all but one, and enabling concurrent refreshes on all but one. It also removes 3 libyear related materialized views that need to be refactored.

### DIFF
--- a/augur/tasks/github/util/github_api_key_handler.py
+++ b/augur/tasks/github/util/github_api_key_handler.py
@@ -135,7 +135,13 @@ class GithubApiKeyHandler():
 
 
         # shuffling the keys so not all processes get the same keys in the same order
-        valid_keys = random.shuffle(valid_keys)
+        try: 
+            self.logger.info(f'valid keys before shuffle: {valid_keys}')
+            valid_keys = random.shuffle(valid_keys)
+            self.logger.info(f'valid keys AFTER shuffle: {valid_keys}')
+        except Exception as e: 
+            self.logger.info(f'{e}')
+            pass 
 
         return valid_keys
 
--- a/augur/tasks/github/util/github_api_key_handler.py
+++ b/augur/tasks/github/util/github_api_key_handler.py
@@ -72,10 +72,10 @@ class GithubApiKeyHandler():
 
         select = WorkerOauth.access_token
         # randomizing the order at db time
-        select.order_by(func.random())
+        #select.order_by(func.random())
         where = [WorkerOauth.access_token != self.config_key, WorkerOauth.platform == 'github']
 
-        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).all()]
+        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).order_by(func.random()).all()]
 
 
     def get_api_keys(self) -> List[str]:
--- a/augur/tasks/github/util/github_api_key_handler.py
+++ b/augur/tasks/github/util/github_api_key_handler.py
@@ -76,7 +76,8 @@ class GithubApiKeyHandler():
         #select.order_by(func.random())
         where = [WorkerOauth.access_token != self.config_key, WorkerOauth.platform == 'github']
 
-        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).order_by(func.random()).all()]
+        #return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).order_by(func.random()).all()]
+        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).all()]
 
 
     def get_api_keys(self) -> List[str]:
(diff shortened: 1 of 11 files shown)


IsaacMilarky (approved):
Mostly LGTM. Randomizing the order of key use in both GithubRandomKeyAuth's and GithubApiKeyHandler's constructor might be a bit redundant but not really a problem.

augur/tasks/util/redis_list.py
```diff
@@ -168,8 +168,10 @@ def pop(self, index: int = None):
         """
 
         if index is None:
-
-            redis.rpop(self.redis_list_key)
+            # This will get a random index from the list and remove it, 
+            # decreasing the likelihood of everyone using the same key all the time
+            #redis.rpop(self.redis_list_key)
+            redis.spop(self.redis_list_key)
```
ABrain7710, line 175: This won't do anything because this isn't how we are getting the api keys from redis.

augur/tasks/github/util/github_random_key_auth.py
```diff
@@ -16,6 +17,7 @@ def __init__(self, session: DatabaseSession, logger):
     
         # gets the github api keys from the database via the GithubApiKeyHandler
         github_api_keys = GithubApiKeyHandler(session).keys
+        github_api_keys = random.sample(github_api_keys, len(github_api_keys))
```
ABrain7710, line 21: I believe the goal here is to randomize the order. If so random.shuffle should be used since Random.sample is used to get a random subset of a list. This will randomly order the list but it is less clear what the goal is

augur/tasks/github/util/github_api_key_handler.py
```diff
@@ -130,6 +133,18 @@ def get_api_keys(self) -> List[str]:
...
+
+        # shuffling the keys so not all processes get the same keys in the same order
+        valid_now = valid_keys
+        try: 
+            self.logger.debug(f'valid keys before shuffle: {valid_keys}')
+            valid_keys = random.sample(valid_keys, len(valid_keys))
+            self.logger.debug(f'valid keys AFTER shuffle: {valid_keys}')
+        except Exception as e: 
+            self.logger.debug(f'{e}')
+            valid_keys = valid_now
```
ABrain7710, line 145: This only gets executed the first time the GithubApiKeyHandler is created after that the keys will be stored in Redis so this function will return in the `if redis_keys` block.

augur/tasks/github/util/github_api_key_handler.py
```diff
@@ -71,9 +72,11 @@ def get_api_keys_from_database(self) -> List[str]:
         from augur.application.db.models import WorkerOauth
 
         select = WorkerOauth.access_token
+        # randomizing the order at db time
+        #select.order_by(func.random())
         where = [WorkerOauth.access_token != self.config_key, WorkerOauth.platform == 'github']
 
-        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).all()]
+        return [key_tuple[0] for key_tuple in self.session.query(select).filter(*where).order_by(func.random()).all()]
```
ABrain7710, line 79: I don't think this is needed as this is just going to randomize the order of the keys that are cached in Redis the first time the GithubApiKeyHandler is created. So this will just make the keys in Redis stored in a different order

ABrain7710 (changes requested):
I'm unsure there is an issue with Github API keys being assigned randomly. I say this because the current implementation is as simple as it gets. We store a list of keys and assign a random key to every request we make. So unless the random.choice function isn't working, I don't think this is the issue.

sgoggins:
@IsaacMilarky 
@ABrain7710 

I was surprised by how much more effective introducing randomization in all three places than in any one of them extended or eliminated periods where the API Rate Limit was exceeded. 

I tried each of them one at a time, and the triple shuffle has API key stops less than 1/2 the time as an single or combination shuffle. 

NOTE: This PR now also includes a fix to the repo_info task. Previously we were pulling the `main` branch commits (actually, it was worse than that: it was `master`), so we were missing about 25% of commit counts, and getting counts off of sometimes "not the default branch". Now its accurate, and on the default branch commit count, so the alignment with our facade commit counting will be perfect. This makes it a lot easier to accurately verify commit collection.

sgoggins:
> I'm unsure there is an issue with Github API keys being assigned randomly. I say this because the current implementation is as simple as it gets. We store a list of keys and assign a random key to every request we make. So unless the random.choice function isn't working, I don't think this is the issue.

I noted this in my specific comments: but I went through several iterations trying to get all the data fully collected, and the three different shuffles occur at different points in execution and are causing key reuse at such a low rate that with a sufficient number of keys we hardly ever run out on a pretty large database.

ABrain7710:
@sgoggins I made an additional observation on the key rotation issue in the discord here: https://discord.com/channels/839539671671504907/915622840798158859/1146192017420996738

ABrain7710 (approved):

PR #2522, opened 2023-09-05
Make Collection Process Easier to Understand

IsaacMilarky:
**Description**
- Change messy web of function calls to start collection to the construction of a CollectionRequest object that you pass in to the AugurTaskRoutine. This new object also has a method to automatically get the desired repos according to the passed in attributes. 

- Put sql calls into their own function for better readability and organization 
- Change old much more general functions to simpler logic that derives more instead of counting on request specific information to be handled by the calling program.

-Update dei endpoint to run repo with new logic

**Signed commits**


augur/tasks/util/collection_util.py
```diff
@@ -36,6 +36,168 @@ class CollectionState(Enum):
...
+            AND augur_operations.collection_status.ml_data_last_collected IS NOT NULL
+            AND ml_status!='{str(CollectionState.COLLECTING.value)}'
+            AND ml_data_last_collected <= NOW() - INTERVAL '{days_until_collect_again} DAYS'
+        """
+    
+    return condition_concat_string
+
+class CollectionRequest:
+    def __init__(self,name,phases,max_repo = 10,new_status = CollectionState.PENDING.value,additional_conditions = None,days_until_collect_again = 1):
+        self.name = name
```
ABrain7710, line 123: I don't think new_status should be a parameter that the constructor takes, since it is derivable from the name. And we aren't currently every passing it into the constructor we are either keeping the default parameter value or deriving it from the fact that it is a facade collection request.

I don't think additional_conditions should be a parameter that the constructor takes either because we aren't using the value

augur/tasks/util/collection_util.py
```diff
@@ -36,6 +36,168 @@ class CollectionState(Enum):
...
+        if name == "facade":
+            self.new_status = CollectionState.UPDATE.value
+
+
+    #Get repo urls based on passed in info.
+    def get_valid_repos(self,session):
+        #getattr(CollectionStatus,f"{hook}_status" ) represents the status of the given hook
+        #Get the count of repos that are currently running this collection hook
+        #status_column = f"{hook}_status"
+        active_repo_count = len(session.query(CollectionStatus).filter(getattr(CollectionStatus,f"{self.name}_status" ) == CollectionState.COLLECTING.value).all())
```
ABrain7710, line 143: I think this should be extracted out into a function called get_active_repo_count(status). So the get_valid_repos method is easer to read without having to worry about how this sql query works

augur/tasks/util/collection_util.py
```diff
@@ -36,6 +36,168 @@ class CollectionState(Enum):
...
+        #Now start old repos if there is space to do so.
+        if limit <= 0:
+            return
+
+        #Get a list of all users.
+        query = s.sql.text("""
+            SELECT  
+            user_id
+            FROM augur_operations.users
+        """)
```
ABrain7710, line 178: Same thing with this as I mentioned above. This could be refactored out into get_users()

augur/tasks/util/collection_util.py
```diff
@@ -477,29 +651,15 @@ def start_repos_from_given_group_of_users(session,limit,users,condition_string,p
...
+    #    valid_repo_git_list,
+    #    phases, repos_type=repos_type, hook=hook
+    #)
 
-    #The maximum amount of repos to schedule is affected by the existing repos running tasks
-    limit = max_repo-active_repo_count
+    return valid_repo_git_list
 
+def split_random_users_list(session,status_col, status_new):
     #Split all users that have new repos into four lists and randomize order
```
ABrain7710, line 663: I don't see where this function is splitting the user list into 4 sections

augur/tasks/util/collection_util.py
```diff
@@ -36,6 +36,168 @@ class CollectionState(Enum):
...
+        active_repo_count = len(session.query(CollectionStatus).filter(getattr(CollectionStatus,f"{self.name}_status" ) == CollectionState.COLLECTING.value).all())
+
+        #Will always disallow errored repos and repos that are already collecting
+
+        #The maximum amount of repos to schedule is affected by the existing repos running tasks
+        limit = self.max_repo-active_repo_count
+
+
+        user_list = split_random_users_list(session,f"{self.name}_status",self.new_status)
+
```
ABrain7710, line 152: I mentioned this below as well, but it appears the splitting of the list into 4 chunks isn't happening in this method call

ABrain7710 (changes requested):
There are a few small things I think would help readability

augur/tasks/util/collection_util.py
IsaacMilarky: Agreed

augur/tasks/util/collection_util.py
IsaacMilarky: I think this would work well as a method for CollectionRequest

augur/tasks/util/collection_util.py
IsaacMilarky: Oh you're totally right this will be fixed

sgoggins (commented):
@IsaacMilarky : Have we tested this branch?

IsaacMilarky:
> @IsaacMilarky : Have we tested this branch?

Yes

ABrain7710 (approved):

PR #2546, opened 2023-10-10
Use Bulk Inserts in Facade Collection Instead of Single Inserts

IsaacMilarky:
**Description**
- Change trim_commits phase of facade to trim commits in bulk instead of one at a time. Before, commits were deleted one at a time.
- Change analyze_commits_in_parallel phase of facade to insert commits in bulk instead of one at a time. Before, more database transactions were used than desirable.
- Implement a stagger insert that inserts a block of facade records when the amount of objects in memory exceeds 10000. 

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -23,7 +23,7 @@ from celery.signals import after_setup_logger
 from datetime import timedelta
 import sqlalchemy as s
 
-from sqlalchemy import or_, and_, update
+from sqlalchemy import or_, and_, update, insert
 
 from augur.tasks.git.util.facade_worker.facade_worker.utilitymethods import update_repo_log, trim_commit, store_working_author, trim_author
 from augur.tasks.git.util.facade_worker.facade_worker.utilitymethods import get_absolute_repo_path, get_parent_commits_set, get_existing_commits_set
@@ -264,7 +264,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
         logger.info(f"Got to analysis!")
         absoulte_path = get_absolute_repo_path(session.repo_base_directory, repo.repo_id, repo.repo_path,repo.repo_name)
         repo_loc = (f"{absoulte_path}/.git")
-        
+
         pendingCommitRecordsToInsert = []
 
         for count, commitTuple in enumerate(queue):
@@ -284,27 +284,37 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
                 if len(pendingCommitRecordsToInsert) > 10000:
                     pendingCommitRecordsToInsert.append(commitRecord)
                 else:
-                    session.insert_data(pendingCommitRecordsToInsert, Commit,["repo_id","cmt_commit_hash","cmt_filename", "cmt_committer_date"],on_conflict_update=False)
+
+                    session.excute(
+                        insert(Commit),
+                        pendingCommitRecordsToInsert,
+                    )
+                    session.commit()
+                    #session.insert_data(pendingCommitRecordsToInsert, Commit,["repo_id","cmt_commit_hash","cmt_filename", "cmt_committer_date"],on_conflict_update=False)
                     pendingCommitRecordsToInsert = []
 
-    try:
-        session.insert_data(pendingCommitRecordsToInsert, Commit,["repo_id","cmt_commit_hash","cmt_filename", "cmt_committer_date"],on_conflict_update=False)
-    except DataError as e:
-        session.logger.error(f"Ran into bad data when trying to insert commit with values: \n {commit_record} \n Error: {e}")
+        try:
+            session.excute(
+                    insert(Commit),
+                    pendingCommitRecordsToInsert,
+                )
+            session.commit()
+        except DataError as e:
+            session.logger.error(f"Ran into bad data when trying to insert commit with values: \n {commit_record} \n Error: {e}")
 
-        #Check for improper utc timezone offset
-        #UTC timezone offset should be betwen -14:00 and +14:00
+            #Check for improper utc timezone offset
+            #UTC timezone offset should be betwen -14:00 and +14:00
+
+            for commit_record in pendingCommitRecordsToInsert:
+                if "time zone displacement" in f"{e}":
+                    commit_record['author_timestamp'] = placeholder_date
+                    commit_record['committer_timestamp'] = placeholder_date
+                else:
+                    raise e
+        except Exception as e:
         
-        for commit_record in pendingCommitRecordsToInsert:
-            if "time zone displacement" in f"{e}":
-                commit_record['author_timestamp'] = placeholder_date
-                commit_record['committer_timestamp'] = placeholder_date
-            else:
-                raise e
-    except Exception as e:
-
-        session.logger.error(f"Ran into issue when trying to insert commit with values: \n {commit_record} \n Error: {e}")
-        raise e
+            session.logger.error(f"Ran into issue when trying to insert commit with values: \n {commit_record} \n Error: {e}")
+            raise e
 
 
     # Remove the working commit.
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -281,7 +281,11 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
             #logger.info(f"Got to analysis!")
             commitRecord = analyze_commit(session, repo_id, repo_loc, commitTuple)
             if commitRecord:
-                pendingCommitRecordsToInsert.append(commitRecord)
+                if len(pendingCommitRecordsToInsert) > 10000:
+                    pendingCommitRecordsToInsert.append(commitRecord)
+                else:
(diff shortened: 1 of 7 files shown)


augur/tasks/git/util/facade_worker/facade_worker/analyzecommit.py
```diff
@@ -126,68 +126,28 @@ def store_commit(repos_id,commit,filename,
 		#session.logger.info(f"Timestamp: {author_timestamp}")
 		commit_record = {
 			'repo_id' : repos_id,
-			'commit' : str(commit),
```
sgoggins, line 129: Am I right that this is making everything map directly, allowing us to remove mapping code below?

augur/tasks/git/facade_tasks.py
```diff
@@ -125,16 +125,9 @@ def update_analysis_log(repos_id,status):
 
         # If there's a commit still there, the previous run was interrupted and
         # the commit data may be incomplete. It should be trimmed, just in case.
-        for commit in working_commits:
```
sgoggins, line 128: What is happening here appears to be simply a refactoring of how working commits are trimmed. Basically putting it into a method. Is that right, @IsaacMilarky ?

augur/tasks/git/facade_tasks.py
```diff
@@ -194,8 +187,8 @@ def update_analysis_log(repos_id,status):
 
 
 
-        for commit in trimmed_commits:
```
sgoggins, line 197: Here down is the factored out trim_commits method, correct, @IsaacMilarky ... trying to make sure I understand what's happening.

sgoggins (commented):
@IsaacMilarky : Have we tested this with the 7 repo starting set to ensure that this branch generates commit data identical to the current method, and also clears out the working_commits table?

augur/tasks/git/facade_tasks.py
```diff
@@ -282,13 +279,54 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+        except DataError as e:
+            session.logger.error(f"Ran into bad data when trying to insert commit with values: \n {commit_record} \n Error: {e}")
+
+            #Check for improper utc timezone offset
+            #UTC timezone offset should be betwen -14:00 and +14:00
+
+            for commit_record in pendingCommitRecordsToInsert:
+                if "time zone displacement" in f"{e}":
+                    commit_record['author_timestamp'] = placeholder_date
+                    commit_record['committer_timestamp'] = placeholder_date
```
ABrain7710, line 312: Is it possible to detect an invalid timezone offset before doing the database insert? If so we should detect it in the analyze_commit method and have it update the timestamps to be the placeholder_date.

augur/tasks/git/facade_tasks.py
```diff
@@ -282,13 +279,54 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+                    pendingCommitRecordsToInsert = []
+
+        try:
+            session.execute(
+                    insert(Commit),
+                    pendingCommitRecordsToInsert,
+                )
+            session.commit()
+        except DataError as e:
+            session.logger.error(f"Ran into bad data when trying to insert commit with values: \n {commit_record} \n Error: {e}")
```
ABrain7710, line 304: commit_record is not defined in this scope since it is outside the for loop

augur/tasks/git/facade_tasks.py
```diff
@@ -282,13 +279,54 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+            for commit_record in pendingCommitRecordsToInsert:
+                if "time zone displacement" in f"{e}":
+                    commit_record['author_timestamp'] = placeholder_date
+                    commit_record['committer_timestamp'] = placeholder_date
+                else:
+                    raise e
+        except Exception as e:
         
+            session.logger.error(f"Ran into issue when trying to insert commit with values: \n {commit_record} \n Error: {e}")
+            raise e
```
ABrain7710, line 318: commit_record is not defined in this scope since it is outside the for loop

augur/tasks/git/facade_tasks.py
```diff
@@ -282,13 +279,54 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+            #logger.debug(commitRecord)
+            if commitRecord:
+                if len(pendingCommitRecordsToInsert) < 10000:
+                    pendingCommitRecordsToInsert.append(commitRecord)
+                else:
+
+                    session.execute(
+                        insert(Commit),
+                        pendingCommitRecordsToInsert,
+                    )
```
ABrain7710, line 292: This insert needs to have the same error handling as below

augur/tasks/git/facade_tasks.py
IsaacMilarky: Yes exactly this also uses a bulk operation instead of sending a query for every commit to trim.

augur/tasks/git/util/facade_worker/facade_worker/analyzecommit.py
IsaacMilarky: Yes mapping is not needed since I just use the table field names in this branch.

IsaacMilarky:
> @IsaacMilarky : Have we tested this with the 7 repo starting set to ensure that this branch generates commit data identical to the current method, and also clears out the working_commits table?

I tested it with a single repo, I did not test it with all 7 starting repos. I will check the 7 starting repos after merging dev

augur/tasks/git/util/facade_worker/facade_worker/analyzecommit.py
```diff
@@ -356,19 +318,10 @@ def store_commit(repos_id,commit,filename,
 					whitespaceCheck.append(line[1:].strip())
 
 	# Store the last stats from the git log
-	store_commit(repo_id,commit,filename,
+	recordToInsert = generate_commit_record(repo_id,commit,filename,
 		author_name,author_email,author_date,author_timestamp,
 		committer_name,committer_email,committer_date,committer_timestamp,
 		added,removed,whitespace)
```
ABrain7710, line 324: This appears to only take the last file encountered in the git log since the recordToInsert variable is being overwritten for each file found. I could be misunderstanding the parsing, but recordToInsert is be overwritten by the last call (the one outside the for loop) to generate_commit_record. So all the calls to generate_commit_record inside the for loop won't matter

ABrain7710 (changes requested):
I found few issues. In addition, I think we should have a fall back method when a bulk insert fails that simply inserts all 10000 commits one at a time so that one bad row doesn't cause us to miss 10000 commits. Alternatively anytime an error occurs we could have it try to insert half the data each time and then and then recursively insert half until the bad data row is the only thing left (this would be more efficient but more complex)

augur/tasks/git/facade_tasks.py
IsaacMilarky: It should be possible from what I remember. You're right this change should be made.

augur/tasks/git/util/facade_worker/facade_worker/analyzecommit.py
IsaacMilarky: Oh I don't know how I didn't notice that. I will make the method return a list next time with all of the files included.

augur/tasks/git/facade_tasks.py
```diff
@@ -282,13 +275,29 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+
+        
+
 
-            analyze_commit(session, repo_id, repo_loc, commitTuple)
+    # Remove the working commit.
+    remove_commit = s.sql.text("""DELETE FROM working_commits 
+    	WHERE repos_id = :repo_id AND working_commit IN :hashes
+    	""").bindparams(repo_id=repo_id,hashes=tuple(queue))
+    session.execute_sql(remove_commit)
```
ABrain7710, line 296: This query to remove the working commits appears to be at the wrong indentation level, which might be causing silent issues

augur/tasks/git/util/facade_worker/facade_worker/utilitymethods.py
```diff
@@ -205,3 +215,55 @@ def update_facade_scheduling_fields(session, repo_git, weight, commit_count):
 
 	session.execute(update_query)
 	session.commit()
+
+def facade_bulk_insert_commits(session,records):
```
ABrain7710, line 219: I think this method might be cleaner if it took in the list of records and then tried to insert them. And if an exception occurs and the insert was on more than one row then we log the multiple record error message, and split in half then try again. But if an exception occurs and the number or records is 1 then we check if it a DataError exception and do all the logic and error logging for a single record insert. Put simply it moves the if len(records) == 1 into the exception block to determine what to log or how to react

ABrain7710 (changes requested):
Overall looks good. There is one small indentation issue and some thoughts on a way to make a method potentially more clear

augur/tasks/git/facade_tasks.py
IsaacMilarky: Yeah this should be indented will fix

augur/tasks/git/util/facade_worker/facade_worker/utilitymethods.py
IsaacMilarky: I implemented this change. I think it makes it slightly more readable

ABrain7710 (approved):

PR #2549, opened 2023-10-16
Updates from graphql branch and debugging branch

sgoggins:
**Description**
- GraphQL API
- Minor bug fixes


sgoggins (commented):
LGTM

augur/application/cli/backend.py
```diff
@@ -184,7 +184,7 @@ def determine_worker_processes(ratio,maximum):
         sleep_time += 6
 
         #15% of estimate, Maximum value of 20
-        facade_num_processes = determine_worker_processes(.2, 20)
+        facade_num_processes = determine_worker_processes(.15, 20)
```
ABrain7710, line 187: I believe this should be 20% so that the sum of the percentages is 100%. 60% core + 20% secondary + 20% facade = 100%. Whereas 15% would cause the sum to 95%

PR #2561, opened 2023-10-26
Repo load message patch

sgoggins:
**Description**
Added descriptive text to explain when users adding new repositories or organizations to their profile should expect to see their data.
image


Ulincsys (approved):
🛫

Issue #2403, opened 2023-05-10
Augur View Registration Form Behavior


ABrain7710:
**Description:**
If user enters a user name and password before pressing the register button, it jumps the user to the email field rather than jumping them to the next field in the list (confirm password). 

Note: It works properly if the user doesn't enter a user name or password and presses register, and also if the user only enters a username then presses register

**How to reproduce:**
1. Enter user name
2. Enter password
3. Press register
4. See that the cursor jumped to the email field

sgoggins:
@Ulincsys

sgoggins:
@Ulincsys : Is this fixed?

kaxada:
@sgoggins assign me this issue. I was in contact with @Ulincsys and he proposed I fix

Ulincsys:
Closed by #2564

PR #2564, opened 2023-10-30
fixed focus field when 'register' is clicked

kaxada:
This PR fixes https://github.com/chaoss/augur/issues/2403


kaxada:
@sgoggins and @Ulincsys I think this solution suffices. WDYT?

Ulincsys (approved):
Looks good!

Ulincsys:
It should be merged into dev though, we typically don't merge directly into main

PR #2581, opened 2023-11-04
Fix orm warnings

ABrain7710:
**Description**
- Fixed orm warnings by adding the correct back_populates on each relationship

NOTE: I thought about adding back_populates to all the relationships so we have them in the future, but I decided to keep this pr small and only add the ones that were showing warnings. I'll likely add the rest in the future

**Signed commits**


sgoggins (approved):
LGTM!  Thanks, @ABrain7710 !!

PR #2590, opened 2023-11-29
Implement ability to add gitlab repos from frontend

ABrain7710:
**Description**
- Adds the ability to add Gitlab repos from the frontend. This includes changes to insert the record into the database, and changes to ensure that the Gitlab repos are ignored. The collection status record is also created, but currently the issue_pr_sum, commit_count, and weights are incorrect since that logic still needs to be implemented with the gitlab api

**Signed commits**


sgoggins (approved):
LGTM. Trying.

PR #2666, opened 2024-01-17
Use Reviewdog to Annotate PRs with Pylint and Spellcheck 

IsaacMilarky:
**Description**
- Add GitHub workflow to run reviewdog and annotate with relevant linters and checkers 
- When a PR is created into either dev or main, pylint checks are ran and annotated on that PR for review
- The only files considered are the ones that contain changes to not be overwhelming

**Notes for Reviewers**
I decided on Pylint because it's not that strict of a linter. We should consider how strict we are willing to be on linting and checking in general in the future as reviewdog supports many tools. Some good linting alternatives are listed in the readme for the [reviewdog repo](https://github.com/reviewdog/reviewdog)

We can also automatically force new PRs to follow PEP8 but that might be overkill. 

**Signed commits**

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,0 +1,19 @@
+name: "run-linting-checks"
+on:
+  pull_request:
+    branches: [main, dev]
+
+jobs:
+  run-pylint:
+    name: runner / pylint
+    permissions: write-all
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


IsaacMilarky:
Example of Pylint check in use on this PR: https://github.com/DSACMS/metrics/pull/74

sgoggins:
I am currently fixing the docs build error that exists across the board following readthedocs.io changes last week.

sgoggins (approved):

PR #2678, opened 2024-02-08
Allow 404 Repos to Automatically be Re-collected 

IsaacMilarky:
**Description**
- Add a new `CollectionState` enum value called 'Standby' that pauses collection on the repo until midnight. This lets us automatically retry repositories that have hit a 404 without crowding out other tasks.
- Fix some import issues where `CollectionState` was being defined multiple times. I have moved this class to its own file in an attempt to make sure that we only define enums once. 
- Add periodic task to set repos to pending daily

**Signed commits**

### DIFF
--- a/augur/application/cli/backend.py
+++ b/augur/application/cli/backend.py
@@ -19,7 +19,8 @@ from urllib.parse import urlparse
 from datetime import datetime
 
 from augur import instance_id
-from augur.tasks.start_tasks import augur_collection_monitor, CollectionState, create_collection_status_records
+from augur.tasks.util.collection_state import CollectionState
+from augur.tasks.start_tasks import augur_collection_monitor, create_collection_status_records
 from augur.tasks.git.facade_tasks import clone_repos
 from augur.tasks.data_analysis.contributor_breadth_worker.contributor_breadth_worker import contributor_breadth_model
 from augur.tasks.init.redis_connection import redis_connection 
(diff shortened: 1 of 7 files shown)


ABrain7710 (changes requested):
The logic in `ping_github_for_repo_move` doesn't appear to be quite right. It appears to not be setting the repos that return a 404 (deleted repos) to a status of IGNORE. It throwing an excpetion for 404, succeeding if it the status wasn't a 301, and setting the repos that are 301 (moved) to IGNORE. We need to be setting 404s to IGNORE and erroring repos that are 301 so they can be retried

sgoggins:
@IsaacMilarky : Is the question @ABrain7710 raised addressed? 

**The logic in ping_github_for_repo_move doesn't appear to be quite right. It appears to not be setting the repos that return a 404 (deleted repos) to a status of IGNORE. It throwing an excpetion for 404, succeeding if it the status wasn't a 301, and setting the repos that are 301 (moved) to IGNORE. We need to be setting 404s to IGNORE and erroring repos that are 301 so they can be retried**

sgoggins (approved):
LGTM, waiting to see if @ABrain7710 's comment is addressed before merging.

ABrain7710:
@sgoggins No the requested changes have not been addressed. I messaged Isaac and he agreed the logic is off

ABrain7710 (changes requested):
I believe we need to be updating all the collection statuses to IGNORE so that we don't collect facade or ml on deleted repos

ABrain7710 (approved):

PR #2687, opened 2024-02-18
Fix typo in query string

ABrain7710:
**Description**
- `secondary_status` is spelled `secondary_staus` in a query string which is causing a syntax error

**Signed commits**


sgoggins (approved):
Looks awesome to me!!

PR #2714, opened 2024-02-21
Remove Unneeded 'git commit' Call in Github Action Script Because it Returns Non-Zero Bash Code

IsaacMilarky:
**Description**
The `git commit` call in the script was causing problems with the script because GitHub actions counts any non-zero bash code as a failure condition. The `git commit` call returns a non-zero bash code because there is nothing to commit. 

**Notes for Reviewers**
Hopefully this is the last one. 

**Signed commits**


sgoggins (approved):
LGTM

PR #2719, opened 2024-02-26
Patch DB CLI to Include GitLab Logic

IsaacMilarky:
**Description**
Allow GitLab repositories to be added to augur via the old cli interface that uses csv files.  Also patches a bug in the CLI that used deprecated syntax to add a repo group. 

**Signed commits**

### DIFF
--- a/augur/application/cli/db.py
+++ b/augur/application/cli/db.py
@@ -126,9 +126,10 @@ def add_repo_groups(filename):
                 if int(row[0]) not in repo_group_IDs:
                     repo_group_IDs.append(int(row[0]))
                     connection.execute(
-                        insert_repo_group_sql,
-                        repo_group_id=int(row[0]),
-                        repo_group_name=row[1],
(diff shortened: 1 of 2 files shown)


augur/application/cli/db.py
```diff
@@ -66,7 +66,6 @@ def add_repos(filename):
                 
                 print(
                     f"Inserting repo with Git URL `{repo_data['url']}` into repo group {repo_data['repo_group_id']}")
-                controller.add_cli_repo(repo_data)
```
ABrain7710, line 69: It appears that controller.add_cli_repo has been removed

ABrain7710 (changes requested):

augur/application/cli/db.py
IsaacMilarky: Oh that was a typo when I was removing a test print statement before making the PR. My bad

ABrain7710 (approved):

PR #2729, opened 2024-03-09
Config usage improvements

ABrain7710:
**Description**
- Start reducing usage of AugurConfig class and start using the database library methods so that the database access logic is centralized
- Defined get_value and get_section in the database lib.py file
- Define temporary_database_engine context method for use in cases where a process needs a short-lived engine like in the celery app setup where a database engine cannot be defined when the worker processes are forked from the parent

**Signed commits**

### DIFF
--- a/augur/api/gunicorn_conf.py
+++ b/augur/api/gunicorn_conf.py
@@ -4,46 +4,42 @@ import logging
 from pathlib import Path
 from glob import glob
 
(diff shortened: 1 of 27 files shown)


sgoggins (approved):
Lets try it!

IsaacMilarky (approved):

PR #2732, opened 2024-03-11
Change server.py back to using AugurConfig

ABrain7710:


IsaacMilarky (approved):
The DCO is missing but good otherwise

PR #2735, opened 2024-03-13
Allow exception visibility in CLI

Ulincsys:
**Description**
This change makes the CLI pass-through *all* exceptions when they occur while loading a module.

Previously, import exceptions were silenced during import of a CLI command module. Although most exceptions would have been displayed, by catching `ModuleNotFoundError`, any import errors in sub-modules would be silenced during Click imports.

The fix is to check first if the module exists using a file check, and if it does not, return None to indicate as such.

**Signed commits**

### DIFF
--- a/augur/application/cli/_multicommand.py
+++ b/augur/application/cli/_multicommand.py
@@ -7,6 +7,8 @@ import os
 import click
 import importlib
 import traceback
+
+from pathlib import Path
 # import augur.application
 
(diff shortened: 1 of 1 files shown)


IsaacMilarky (dismissed):
LGTM this is needed

sgoggins (dismissed):
Thank you!

sgoggins (approved):

PR #2759, opened 2024-04-10
Add Check to SQL Update Statements to Avoid constraints on `date_last_collected`

IsaacMilarky:
**Description**
- There was a bug where tasks that had succeeded in the past but then subsequently fail later would not be reset correctly; running into a constraint on 'date_last_collected'
- In response I have added checks to the `date_last_collected` field in order to prevent this issue
- Specifically we now check for null values in the date field and update to either `SUCCESS` or `PENDING` depending on if the field is null or not. 
- The function now should reset the task to a state where it can be run again. 

**Signed commits**


sgoggins (approved):
LGTM

PR #2777, opened 2024-04-27
Rollback transaction when query throws exception

ABrain7710:
**Description**
- Rollback changes when a query using the session throws an exception so that the session can still be used

This PR fixes #
- Issue where frontend doesn't work because the session fails on every query

Note: I don't care if this is merged into main or dev, I just noticed that dev is the same as main and this is a small fix so I opened it into main

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -878,7 +878,13 @@ class Repo(Base):
     @staticmethod
     def get_by_id(session, repo_id):
 
-        return session.query(Repo).filter(Repo.repo_id == repo_id).first()
+        try:
+            return session.query(Repo).filter(Repo.repo_id == repo_id).first()
+        except Exception as e:
+            session.rollback()
+            raise e
+
+        
 
     @staticmethod
     def get_by_repo_git(session, repo_git):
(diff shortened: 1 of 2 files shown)


augur/application/db/models/augur_data.py
```diff
@@ -878,7 +878,13 @@ class Repo(Base):
     @staticmethod
     def get_by_id(session, repo_id):
 
-        return session.query(Repo).filter(Repo.repo_id == repo_id).first()
+        try:
+            return session.query(Repo).filter(Repo.repo_id == repo_id).first()
+        except Exception as e:
```
Ulincsys, line 883: Is there a specific exception we can catch? What happens if some other exception occurs which doesn't necessitate a rollback?

ABrain7710: There wasn't a specific exception that I could find (if you can find one other than rolling back when we get the PendingRollbackException that would be awesome). I don't think calling rollback when nothing happened like during an operational exception should do anything.

Ulincsys (approved):
lgtm

sgoggins (approved):

PR #2782, opened 2024-05-02
handle rate limiting and more repo load messages

GaryPWhite:
**Description**
- more messages when loading repos
- patched edge cases where sometimes no message would occur
- rate limiting will now wait to continue loading repositories (before it would fail and continue to the next repo, making large load jobs very difficult)
- allowed for [use of a proxy](https://www.python-httpx.org/environment_variables/#http_proxy-https_proxy-all_proxy) when checking internet access (`http_proxy`/`HTTP_PROXY` is standard in many libraries, including the one in use here)

Future enhancements could include skipping validation on github until checking that a given repo is not in the database -- that's a problem for another day.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/cli/__init__.py
+++ b/augur/application/cli/__init__.py
@@ -3,9 +3,9 @@ import click
 from functools import update_wrapper
 import os
 import sys
-import socket
 import re
 import json
+import httpx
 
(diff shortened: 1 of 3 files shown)


sgoggins (approved):
LGTM.

PR #2791, opened 2024-05-14
Add default arg values for rabbitmq dockerfile

Ulincsys:
**Description**
- Provide default values for the RabbitMQ dockerfile args so it will build by default

It is still recommended to pass values for production builds via `--build-arg key=value`

**Signed commits**


sgoggins (approved):
LGTM

PR #2823, opened 2024-06-17
Improve github paginator

ABrain7710:
**Description**
- Created a more generic github data access class that can be used to paginate over a resource or get a single resource if the github endpoint returns a dict. It also handles retries with a library so that we don't have to build our own logic. It uses the `response.raise_for_status()` httpx method to ensure that an exception is thrown if a failure status code is returned (this allows us to assume that the response is successful if no exceptions occur Also the `make_request()` method throws a rate limit exception so that the caller can determine how to handle the rate limit, but the higher level methods like paginate resource and get resource handle the rate limit like the github api docs tell us to. 

I did this work to support the core recollection work, since I need a reliable method to get the page count of a resource and the old github paginator was not reliable enough. 

Currently I only added it to the pull request task, but over time I will start to add it to other tasks as we become more confident in it

**Signed commits**


sgoggins (approved):
Ready to test!

PR #2825, opened 2024-06-20
fix missing import error

Ulincsys:
**Description**
- Temporary patch for required import to determine the frontend config location

**Signed commits**


sgoggins (approved):
LGTM!

PR #2826, opened 2024-06-21
Explicitly identify iconized buttons using role attribute

Ulincsys:
**Description**
- Makes the favorite/star buttons use a pointer cursor
- Allows screen-reader users to identify buttons that were previously not visible to screen-reader software

**Signed commits**


sgoggins (approved):
LGTM!

PR #2831, opened 2024-06-24
Add check to see if git credentials are writable

IsaacMilarky:
**Description**
Add check to config.sh to see if we can write git credentials and exit if we can't 

**Signed commits**


sgoggins (approved):
LGTM

PR #2838, opened 2024-06-29
Use key manager in github data access

ABrain7710:
**Description**
- The new github data access had a key manager passed to it but it was not using it to make the request, therefore the requests were being made without a github api key. This simply passes the key manager so it is provided an api key
- 
**Signed commits**


sgoggins (approved):
Good fix!

PR #2839, opened 2024-06-29
Remove empty files and move redundant folders to simplify project structure

ABrain7710:
**Description**
- Removed empty core.py files. This left several folders with just a tasks.py for example the issues folder only had a tasks.py. So in the issues example the file was at `augur/tasks/github/issues/tasks.py` and now it is at `augur/tasks/github/issues.py`.

**Note**
Should be merged after: #2840 

**Signed commits**


sgoggins (approved):
Are you sure? :)

PR #2842, opened 2024-07-01
Proper Gunicorn bringup detection

Ulincsys:
**Description**
- Add `/api` route to get the location of the backend API
- Check that Gunicorn has actually started during `backend start` invocation
  - Poll the process state every 0.5 seconds
  - Make GET request to `/api` to check that service has started
  - Repeat until the state of the service can be determined

**Future Work**
- Implement timeout in case Gunicorn takes too long to respond or gets stuck during startup
  - Currently this waits until either the process dies or the server responds

**Signed commits**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ augur_export_env.sh
 !docker.config.json
 config.yml
 reports.yml
+*.pid
 
 node_modules/
 .idea/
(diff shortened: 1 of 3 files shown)


sgoggins (approved):
Looks worth testing after this coming release since we are already testing.

ABrain7710 (approved):

PR #2860, opened 2024-07-03
Add metadata to json decode errors

ABrain7710:
**Description**
- Currently we are logging the output when we get the json code errors in `parse_json_from_subprocess_call ` but it is impossible to find these logs in the huge log files so I added the exception so that the relevant data will show up in flower

**Signed commits**


IsaacMilarky (approved):
LGTM

PR #2861, opened 2024-07-04
Refactor github graphql 

ABrain7710:
**Description**
- Create graphql equivalent to the new github data access so we can be sure that the proper exception will be thrown when an error occurs. 
- This change is in response to these `TimeoutError: No data received from endpoint.` errors in the current logic. These occur when we don't get data (None is returned) due to us returning None in places rather than throwing exceptions. The new code will help us understand why the error is occurring because it will throw an exception for any issue rather than returning None.

**Signed commits**


sgoggins (approved):
Looks good!

PR #2879, opened 2024-07-25
Add Commit Message Table and Add Commit Messages to Facade Collection

IsaacMilarky:
**Description**
- Previously Facade did not care about the commit message 
- Now, we have a new table that has a unique record for each repo and commit hash with the hash's corresponding commit message
- We bulk insert the commit message data

**Notes for Reviewers**
Should be tested on a range of repositories before being put into production. For now this is a draft PR until we can be more confidant in it. 

**Signed commits**


sgoggins (approved):
LGTM

PR #2943, opened 2024-10-27
Fix Libyear Parsing Function to Support More than One Dependency Manifest per Project

IsaacMilarky:
**Description**
- Use `.get` instead of dictionary indices to avoid missing info in poetry.lock
- Make each dependency function extend the dep list instead of returning once we find the first manifest file
- Parse the path into a short string that can be more easily compared
- Parse through all found dependency manifest files and put them all in one big dictionary

**Notes for Reviewers**
I have tested locally on my repos but @sgoggins you might want to test it a bit on other repos.

**Signed commits**


sgoggins (approved):
LGTM.

PR #2947, opened 2024-11-06
Improve dockerization

Mbaoma:
**Description**
- I updated the Dockerfile to use multistage build to reduce image build time when the ```docker-compose up --build``` command is run.

This PR fixes #
My pull request reduced the duration of the **Run the Build** job in the pipeline from **6 minutes and 15 seconds** to **2 minutes and 8 seconds**.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/scripts/docker/install-go.sh
+++ b/scripts/docker/install-go.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
-# install Go
-installGo() (
-    cd "$(mktemp -d)"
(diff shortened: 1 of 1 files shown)


Mbaoma:
@GregSutcliffe Thank you for catching that! Yes, that accounts for the speedup, as those scripts are typically resource-intensive. However, upon review, I ran the installation scripts in the `runtime` stage, to ensure the dependencies are installed before the application starts.

Issue #2952, opened 2024-11-15
Docker-compose: rustup: not found


sebastian-luna-valero:
**Description:**

Deploying `augur` using docker:

https://oss-augur.readthedocs.io/en/main/docker/getting-started.html

Does not work.

**How to reproduce:**

1. clone repo

```bash
git clone https://github.com/chaoss/augur.git
cd augur/
```

2. create env file

```bash
vi .env
AUGUR_GITHUB_API_KEY=<pat>
AUGUR_GITHUB_USERNAME=<user>
```

3. run it

```bash
sudo docker-compose up --detach
```

**Expected behavior:**

All docker services up and running.

**Screenshots**
N/A

**Log files**

Using the `main` branch

```bash
[+] Running 2/2
 ! rabbitmq Warning                                                                                                                                                                                           1.7s 
 ! augur Warning                                                                                                                                                                                              1.7s 
[+] Building 190.9s (27/46)                                                                                                                                                                         docker:default
 => [rabbitmq internal] load build definition from Dockerfile                                                                                                                                                 0.1s
 => => transferring dockerfile: 721B                                                                                                                                                                          0.0s
 => [rabbitmq internal] load metadata for docker.io/library/rabbitmq:3.12-management-alpine                                                                                                                   1.8s
 => [rabbitmq internal] load .dockerignore                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                               0.0s
 => [rabbitmq  1/10] FROM docker.io/library/rabbitmq:3.12-management-alpine@sha256:0b44fbcc3a4bf22d00090f1353127577dbe1fcb109c41669733a9d7ecf6c3a78                                                           5.0s
 => => resolve docker.io/library/rabbitmq:3.12-management-alpine@sha256:0b44fbcc3a4bf22d00090f1353127577dbe1fcb109c41669733a9d7ecf6c3a78                                                                      0.0s
 => => sha256:b4c22cf1a563f645ce107213da67c68b395c2c0ea004c5dd93190919d6902905 8.91kB / 8.91kB                                                                                                                0.0s
 => => sha256:43c4264eed91be63b206e17d93e75256a6097070ce643c5e8f0379998b44f170 3.62MB / 3.62MB                                                                                                                0.3s
 => => sha256:cf96d81494f45a6afe66a5dc908f2977f2f5f9ce5cb2c1c08f6c37db088adb2d 39.93MB / 39.93MB                                                                                                              0.8s
[...]

PR #2957, opened 2024-12-05
Update Go version in Docker, and re-enable SCC/Scorecard

GregSutcliffe:
**Description**
This commit fixes a couple of things.

Firstly, the `install-go` script uses an ancient version of Go that conflicts with the newer version needed for SCC, so that is updated. That script also wrapped the install in a function for no apparent reason, so I simplified it.

Second, I re-enabled the SCC & Scorecard installs. These have been commented out for some time, apparently because of the Go issues. Rather than uncomment it, I copied the relevant parts from `workers.sh` because the container doesn't need checks about whether directories exist - these are guaranteed not to exist in a fresh container.

**Notes for Reviewers**

I believe this will result in working SCC and Scorecard calculations in the container now

**Signed commits**
- [x ] Yes, I signed my commits.

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -100,9 +100,10 @@ RUN set -x \
     && /opt/venv/bin/pip install wheel \
     && /opt/venv/bin/pip install .
 
+RUN ./scripts/docker/install-go.sh
+ENV PATH="${PATH}:/usr/local/go/bin"
 RUN ./scripts/docker/install-workers-deps.sh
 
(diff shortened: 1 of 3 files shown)


sgoggins (approved):
Awesome! LGTM!

GregSutcliffe:
Thanks @sgoggins - for the record I don't think this affects #2952, as I've not touched the rust build part at all. I suspect that was a transient failure to download rust - I have a plan to move the image to use build layers (see #2947 for some of that) which will help there, but ENOTIME :)

PR #2970, opened 2025-01-15
Only collect messages on data that was updated

ABrain7710:
**Description**
- Speed up message collection by only collecting data that was updated


sgoggins (commented):
lgtm

PR #2996, opened 2025-02-13
Fix key orchestrator missing in Docker build

### LINKED ISSUE #2994: Error running release 0.81.0 in Docker
The latest release seems to build but does not run when built for Docker. The image crashes with this error:

```
  File "/opt/venv/lib/python3.9/site-packages/augur/tasks/github/util/github_data_access.py", line 6, in <module>
    from keyman.KeyClient import KeyClient
ModuleNotFoundError: No module named 'keyman'
```

I can see a `keyman` dir in the source code that seems new, but it isn't copied into the image - there are bunch of copies (https://github.com/chaoss/augur/blob/main/docker/backend/Dockerfile#L76-L82) but `keyman` is outside of `/augur`.

Not sure if it needs to be pip-installed or copied in, or both.

Ulincsys:
**Description**
- Copy keyman directory in Dockerfile
- Update graphical dockerfile to reflect the current build process

This PR fixes #2994 

**Signed commits**

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -80,6 +80,7 @@ COPY ./augur/ augur/
 COPY ./metadata.py .
 COPY ./setup.py .
 COPY ./scripts/ scripts/
+COPY ./keyman/ keyman/
 
 # Add rust and cargo to PATH
 ENV PATH="/usr/bin/:/root/.cargo/bin:/usr/local/bin:${PATH}"
--- a/docker/backend/graphical
+++ b/docker/backend/graphical
@@ -1,22 +1,75 @@
-#SPDX-License-Identifier: MIT
-FROM python:3.9-slim-bullseye
+# SPDX-License-Identifier: MIT
+FROM python:3.11-slim-bullseye
 
 LABEL maintainer="outdoors@acm.org"
-LABEL version="0.51.1"
+LABEL version="0.76.6"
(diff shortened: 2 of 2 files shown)


sgoggins:
@Ulincsys : Is there a reason this is still in draft mode? Do we want @GregSutcliffe to test this branch on his docker config?

Ulincsys:
I'm not currently capable of running the docker build, ideally we would need **somebody** to test it before we push it to the main branch.

sgoggins:
OK. That makes sense. We will see if @GregSutcliffe has time and is comfortable doing that ... otherwise I will set out to test it in the morning.

GregSutcliffe:
I already tried just doing a COPY in my environment yesterday, and it didn't fix the issue I'm afraid.

Ulincsys:
Hi @GregSutcliffe 

I was able to (eventually) get Docker working on my end.

I found that the version of Python we're using to run the Docker build does not seem able to import from a module which does not contain an `__init__.py`. We tested this on a few different versions of Python on bare-metal, and none of them complained about it! After adding that, I was able to run collection via Docker compose from a clean environment, and verified that the key orchestrator was running properly.
image

Please let us know if you still have issues with the build, but otherwise I think this is probably good to patch into main.

sgoggins (approved):

PR #3000, opened 2025-02-17
Strip whitespace from keys while loading

### LINKED ISSUE #2998: New key management hard-crashes on a bad key
Following on from #2994 discussions...

The new key management is v0.8.1 is handy, but it hard crashes if it encounters a bad key:

```
beat: Starting...
2025-02-14 14:18:13 da9ef9a76ffe augur[7] INFO Retrieved 16 github api keys for use
WARNING: The key 'redacted' is not a valid key. Hint: If valid in past it may have expired
WARNING: The key 'redacted' is not a valid key. Hint: If valid in past it may have expired
WARNING: The key 'redacted' is not a valid key. Hint: If valid in past it may have expired
Protocol Error: <class 'httpx.ProtocolError'>

augur backend start command setup failed
                 You are not connected to the internet.
                 Please connect to the internet to run Augur
                 Consider setting http_proxy variables for limited access installations.
```

Given a long-lived instance is pretty much *guaranteed* to hit a bad or expired key during it's lifetime, this should be handled and reported to the user, rather than causing a crash.

Ulincsys:
**Description**
- Fixes a crash that occurs when using HTTPX with API keys that have trailing whitespace

This PR fixes #2998 

**Signed commits**

### DIFF
--- a/augur/tasks/github/util/github_api_key_handler.py
+++ b/augur/tasks/github/util/github_api_key_handler.py
@@ -107,6 +107,8 @@ class GithubApiKeyHandler():
 
         if len(keys) == 0:
             return []
+        
+        keys = [key.strip() for key in keys]
 
         valid_keys = []
         with httpx.Client() as client:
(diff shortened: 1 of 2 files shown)


sgoggins (approved):
LGTM.

PR #3002, opened 2025-02-17
Evaluate new things for key rotation

sgoggins:
**Description**
- key rotation improvements

### DIFF
--- a/augur/application/cli/__init__.py
+++ b/augur/application/cli/__init__.py
@@ -12,7 +12,7 @@ from augur.application.db import get_engine, dispose_database_engine
 from sqlalchemy.exc import OperationalError 
 
 
-def test_connection(function_internet_connection):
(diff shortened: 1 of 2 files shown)


augur/application/cli/__init__.py
```diff
@@ -34,7 +34,39 @@ def new_func(ctx, *args, **kwargs):
...
         
-    return update_wrapper(new_func, function_internet_connection)
+    return update_wrapper(new_func, function_internet_connection) """
+
+import socket
+import sys
+import functools
+import click
+
+def has_internet(host="8.8.8.8", port=53, timeout=3):
```
sgoggins, line 44: This is the important part, which changes what we check from a site URL that may have some volatility from time to time to a DNS Nameserver that has the highest uptime requirements of anything on the internet.

PR #3029, opened 2025-03-01
Hotfix since date issue in collect events task

ABrain7710:
**Description**
Context: `since` is set to None when it is a full collection. The query for `thorough` events collection was always adding `since` to the query which caused it to fail

Solution: Only include `since` in the query if it is set

**Signed commits**


sgoggins (approved):
Thank you! THis should fix the odd bug.

PR #3046, opened 2025-03-11
Events fix : Seems to work

sgoggins:
**Description**
- This is a fix for how events are managed for insertion. Specifically, the key structure is changed so that uniqueness is determined via an integer key instead of an encoded string. There is no clear reason why, in the case of one Augur instance, the string based key stopped getting correctly evaluated. 
- Numeric/Integer keys are more performant and robust, so this change implements a change to the unique key.


Ulincsys (approved):
Fantastic!

Ulincsys (approved):
Even better

Issue #3053, opened 2025-03-12
Facade Error: process_scc_value_metrics: FileNotFoundError: [Errno 2] No such file or directory: '/root/scc'


cdolfi:
On our instance we have 1,000+ of these since core got unblocked: 

Traceback: 

> Traceback (most recent call last):
  File "/opt/venv/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/venv/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
  File "/augur/augur/tasks/git/scc_value_tasks/tasks.py", line 13, in process_scc_value_metrics
    value_model(logger,repo_git,)
  File "/augur/augur/tasks/git/scc_value_tasks/core.py", line 25, in value_model
    required_output = parse_json_from_subprocess_call(logger,['./scc', '-f','json','--by-file', path], cwd=path_to_scc)
  File "/augur/augur/tasks/util/worker_util.py", line 132, in parse_json_from_subprocess_call
    p = subprocess.run(subprocess_arr,cwd=cwd,capture_output=True, text=True, timeout=None)
  File "/usr/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1823, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/root/scc'

MoralCode:
uuuh i hope that wasnt my fault 😬 

(looks like it probably was lol)

https://github.com/chaoss/augur/blob/65e08c9351a56a0ff55eeac6e29e27c2fd31b7e0/docker/backend/Dockerfile#L123

I think i wasnt being careful about paying attention to the WORKDIR change earlier in the file. The destination  path of this line and probbaly the subsequent L124 should probably be in /root, not relative to the PWD)

Alternatively, moving these lines to before the `WORKDIR /augur` call may also work


this is just my quick and dirty analysis though. Let me know if nobody else has time to fix this and i can give it a shot

cdolfi:
If you could try to fix it that would be great! I think youd have the most context

[...]

Issue #3054, opened 2025-03-12
Facade Error: insert_facade_contributors: TypeError('sequence item 1: expected a bytes-like object, NoneType found')


cdolfi:
Since core got unblocked we have 1,000+ of this error: 

Exception: 

> TypeError('sequence item 1: expected a bytes-like object, NoneType found')

Traceback (most recent call last): 
[...]

PR #3055, opened 2025-03-14
adjust path that scc gets copied to

MoralCode:
**Description**
This is a *really* low effort shot in the (almost) dark solution to #3053 

Having some testing would be appreciated, especially if this does anything other than fixing the issue or at least moving the error to some thing similar to a `file not found` for `scorecard` (rather than `scc`) 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -120,7 +120,7 @@ RUN set -x \
 
 RUN ./scripts/docker/install-workers-deps.sh
 
-COPY --from=golang "/root/scc" "./scc"
+COPY --from=golang "/root/scc" "/root/scc"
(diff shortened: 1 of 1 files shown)


MoralCode:
ok, so this fix solves the narrow issue that cali reported.

I have yet to confirm whether the `scorecard` binary has a similar issue.

@cdolfi do you see any issues relating to scorecard processes in your augur instance? When does that run in relation to when fascade tries to run the `scc` binary?

cdolfi:
Yes I do, the error is in #3054

cdolfi:
Is this ready to go from draft to a full PR? I'm very pro small PRs that each tackle something

sgoggins (approved):
LGTM

PR #3061, opened 2025-03-16
Remove Vagrantfile

MoralCode:
Vagrant is [no longer FOSS](https://github.com/hashicorp/vagrant/blob/main/LICENSE) and I have a suspicion that this has not been used as a development environment for a while since there is a lot more docker related infrastructure in the repos.

Unsure if there are any viable forks to move to

Especially with all the new GSOC contributors who are just starting out, I guess I'm just casually seeing if there are ways to clean up the repository so that the onboarding is cleaner


sgoggins (approved):
Thank you!

PR #3069, opened 2025-03-18
Remove key from pool when it is no longer valid

ABrain7710:
**Description**
- Removes github api key from the pool when the github api returns a 401 reponse 

**Signed commits**


Ulincsys (approved):
Looks good!

PR #3071, opened 2025-03-19
Leave room on github api key rate limit

ABrain7710:
**Description**
- Stop using github api keys when they have 50 requests left

**Signed commits**

### DIFF
--- a/augur/tasks/github/util/github_data_access.py
+++ b/augur/tasks/github/util/github_data_access.py
@@ -5,7 +5,7 @@ from tenacity import retry, stop_after_attempt, wait_fixed, retry_if_exception, 
 from urllib.parse import urlparse, parse_qs, urlencode
 from keyman.KeyClient import KeyClient
 
-GITHUB_RATELIMIT_REMAINING_CAP = 100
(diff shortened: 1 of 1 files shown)


sgoggins (approved):
LGTM

MoralCode:
@ABrain7710 i think this introduces a bug

the line `self.logger.warning(f"X-RateLimit-Remaining was not an integer. Value: {response.headers["X-RateLimit-Remaining"]}")` is causing python to throw `f-string: unmatched '['`.

While it doesnt make sense on its face because the matching  `]` is clearly there, what this error is actually complaining about is the quotes. You used double quotes `"` for both the enclosing f-string AND the dictionary access when reading the header, causing python to think the f-string ends at `f"X-RateLimit-Remaining was not an integer. Value: {response.headers["` and causing the issue


Corrected code that uses single quotes to disambiguate:
`self.logger.warning(f"X-RateLimit-Remaining was not an integer. Value: {response.headers['X-RateLimit-Remaining']}")`

another valid fix would be to move `response.headers['X-RateLimit-Remaining']` into a variable and refer to that variable in the f-string

PR #3076, opened 2025-03-19
copy scorecard binary to correct place too

MoralCode:
Another experimental patch for scorecard fix, similar to #3055 

**Signed commits**

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -121,7 +121,7 @@ RUN set -x \
 RUN ./scripts/docker/install-workers-deps.sh
 
 COPY --from=golang "/root/scc" "/root/scc"
-COPY --from=golang "/root/scorecard/scorecard" "./scorecard"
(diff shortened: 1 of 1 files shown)


docker/backend/Dockerfile
```diff
@@ -121,7 +121,7 @@ RUN set -x \
 RUN ./scripts/docker/install-workers-deps.sh
 
 COPY --from=golang "/root/scc" "/root/scc"
-COPY --from=golang "/root/scorecard/scorecard" "./scorecard"
+COPY --from=golang "/root/scorecard/scorecard" "/root/scorecard"
```
sgoggins, line 124: @MoralCode : I've ignored this because its in DRAFT mode. Should we do it?

MoralCode: oh, uuh if its been causing issues then sure - or maybe at least for consistency?

@cdolfi have you seen errors relating to this?

PR #3093, opened 2025-03-31
Rework CI container builds

### LINKED ISSUE #3092: Ensure containers are built in CI and gate merges to main and dev
From the call today, I agreed to:

- All docker containers should be built during PRs targeted at `main` and `dev`
- The builds should gate merges
- Preserve the current push to ghcr on merge/release

@sgoggins Feel free to correct my summarization.

JohnStrunk:
**Description**
- Move from custom scripting to re-usable actions
- Add workflow_dispatch so build can be triggered manually
- Add build caching via actions cache

This PR fixes #3092

**Notes for Reviewers**

**Signed commits**


JohnStrunk:
I've left this as a draft pending feedback on what the proper handling of `latest` and `devel-latest` image tags should be.
- The original workflow would only update `latest` during a release, matching it to the most recently released version.
- `devel-latest` would get updated any time there was a push to either `main` or `dev`, and I don't think that's an intended behavior.

What would you like the behavior to be?

sgoggins:
@JohnStrunk : My first thought, which I acknowledge is perhaps uselessly vague, is to have the behavior be consistent with what other projects would use. I think matching user expectations is important. 

Left to my own devices I think it makes sense to update devel-latest when changes are pushed to dev.

JohnStrunk:
ok. I think I understand the project branching model as (somewhat loosely):
- New development goes into `dev`
- At some point `dev` is merged to `main`
- Releases are tagged from `main`

With the above, I think having `dev` map to `devel-latest` is fine, and only updating `latest` w/ a release is ok.

JohnStrunk:
Updated and ready for review.
This is targeted at main, but I *think* it will also need to be cherry-picked to `dev` in order for it to work on that branch as intended.
I'm also happy to rebase this on dev and CP the other way, too.

sgoggins (approved):
Looks great to me! THANK YOU!

PR #3100, opened 2025-04-01
Fix failing CI when not tagging a docker image

JohnStrunk:
**Description**
This is a follow-up to #3093 that attempts to fix the current failing CI due to trying to push a non-tagged docker image.

The following is taken from #3010
```
ERROR: tag is needed when pushing to registry
```

The issue is that we are trying to push the image for any non-PR triggered build, but that doesn't match the logic for when we actually tag an image. Hopefully the logic here will key off of whether we have actually applied a tag with the actual intent to push.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -71,7 +71,8 @@ jobs:
           file: ./docker/${{ matrix.image }}/Dockerfile
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64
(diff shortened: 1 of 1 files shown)


JohnStrunk:
@sgoggins I noticed my PR from yesterday was causing some failed checks. This should fix it.

sgoggins:
@JohnStrunk : Just to confirm my understanding here: We will push new images when a new release is tagged now?

PR #3120, opened 2025-04-08
[infrastructure] Add .gitattributes to normalize line endings and prevent Windows Docker build issues.

platynom:
Description
- This pull request introduces a `.gitattributes` file at the root of the repository to help normalize line endings across different operating systems. This is especially useful for Windows users, where improper line endings can cause Docker build scripts (like `workers.sh`) to fail with "file not found" errors, even when the file exists.

The file contains:
# Normalize line endings for all text files
* text=auto

# Force shell scripts to use LF (important for Docker builds)
*.sh text eol=lf

Notes for Reviewers
This change does not impact application logic or features.

It is strictly an environment improvement for cross-platform compatibility.

Helpful for preventing subtle Docker build issues caused by CRLF/LF mismatch on Windows systems.

Signed commits
 Yes, I signed my commits..

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Normalize line endings for all text files
+* text=auto
+
+# Force shell scripts to use LF (important for Docker builds)
+*.sh text eol=lf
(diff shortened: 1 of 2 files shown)


platynom:
Just wanted to clarify a small error in my GSoC proposal — I mistakenly titled the project as "Conversion Rate Metrics", which was not my intended topic. While reviewing some PRs during the submission process, I believe I accidentally copy-pasted the wrong title.

My actual project is "Strengthening OSS Communities: Metrics & Infrastructure for Augur", focusing specifically on implementing the Upstream Code Dependencies metric. The proposal content reflects this correctly — only the title was mislabeled.

This PR and all my ongoing work are aligned with the upstream dependencies metric under the correct scope. Apologies for any confusion, and thank you for your understanding and guidance!

Issue #3107, opened 2025-04-04
Ability to run Augur container as non-root


JohnStrunk:
**Is your feature request related to a problem? If so, please describe the problem:**
In order to run Augur on OpenShift, it's necessary to be able to run all the containers as an arbitrary non-root user. This doesn't seem to be a problem for the rabbitmq, postgres, and redis containers, but the actual Augur container won't run as non-root.

**Potential solutions:**
We should be able to specify an arbitrary `user:group` in the docker-compose file for the `augur` service and it should "just work". Ex:
```yaml
services:
  augur:
    user: 2345:2345  # Run as an arbitrary non-root user
```

**Additional context:**
While there are certainly file permission issues:
```
$ docker logs augur-augur-1
/bin/bash: /entrypoint.sh: Permission denied
```
there also seem to be path assumptions in the code that will need to be corrected.

Related: #1389

PR #3124, opened 2025-04-08
Allow log directory to be specified

JohnStrunk:
**Description**
- `augur config init` now takes a parameter for the log file location
- Log file location in the docker container has been moved to /logs
- Log file "reset" step now deletes the *contents* of the log directory, not the directory itself

Related to #3107 

**Notes for Reviewers**
This is a step toward the non-root container in that it gets the log files out of a privileged location. The behavior for non-docker instances should be unchanged.

**Signed commits**


JohnStrunk:
I probably should have added a reason for the change to the "reset" behavior:
While it's much simpler to just delete the whole directory, that isn't possible if the directory is a mounted volume, and we're eventually going to want that... either as a temp directory or as a PVC. So, I'm just deleting the contents instead.

Ulincsys (approved):
LGTM

sgoggins (approved):
LGTM.

PR #3130, opened 2025-04-15
Configure key orchestrator to run in an independent container

Ulincsys:
**Description**
- Add docker detection to key orchestrator entrypoint
- Add keyman Dockerfile
- Update backend start command to avoid starting keyman in docker

**TESTING**
Has not been tested on bare metal, will need testing to verify bare-metal functionality remains

**Signed commits**

### DIFF
--- a/keyman/Orchestrator.py
+++ b/keyman/Orchestrator.py
@@ -1,13 +1,29 @@
-from augur.tasks.init.redis_connection import redis_connection as conn
-from augur.application.logs import AugurLogger
+import os
 import json, random, time
 
 from keyman.KeyOrchestrationAPI import spec, WaitKeyTimeout, InvalidRequest
 
+if os.environ.get("KEYMAN_DOCKER"):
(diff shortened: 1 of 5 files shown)


keyman/Orchestrator.py
```diff
@@ -1,13 +1,29 @@
...
+
+    conn = redis.Redis.from_url(os.environ.get("REDIS_CONN_STRING"))
+
+    # Just log to stdout if we're running in docker
+    logger = logging.Logger("KeyOrchestrator")
+else:
+    from augur.tasks.init.redis_connection import redis_connection as conn
+    from augur.application.logs import AugurLogger
+
+    logger = AugurLogger("KeyOrchestrator").get_logger()
```
ABrain7710, line 21: Does the logger also need to be defined when `KEYMAN_DOCKER` is false?

ABrain7710 (approved):

PR #3131, opened 2025-04-16
Add ability to run augur container as a non-root user

### LINKED ISSUE #3107: Ability to run Augur container as non-root

JohnStrunk:
**Description**
This PR provides the ability to run the augur backend container as an arbitrary non-root user. This is important for running on OpenShift where the userid is not known until runtime, and containers are not permitted to run as uid 0.

The majority of the changes are related to moving R/W files such as config and log files out of the source tree and into paths that will have data volumes mounted. The `docker-compose.yml` was updated to provide the required volumes.
File permissions during the build process were also updated so that non-root users will have read access.

I also changed the build steps for `scorecard` and `scc` to directly `go install` them instead of manually checking out the repo and running `go build` on the local copy.

This PR fixes #3107

**Notes for Reviewers**
- I *think* I've preserved the existing behavior when the new variables aren't set
- I've added a bunch of new environment variables to control file locations. I'm certainly not sold on that-- I feel like it's made things a bit complex even though they all should be optional. If you have another suggestion, please let me know.
- I added 3 new volumes, but I'm not sure they are all necessary. I'm not familiar enough with the code at this point to confidently make a call about what should be co-located. Ideally, we'd have 1 volume for persistent data, and 1 for temporary data (which would really just be the `/tmp` directory). I'm happy to modify this PR to that model if you can provide some guidance about what's important to keep and what is ephemeral.

**Signed commits**


sgoggins (approved):
LGTM

PR #3142, opened 2025-04-30
force consistency for debian version 

cdolfi:
The Augur container uses debian 11 but without specifying bullseye golang 1.23 uses 12. This impacts scc and scorecard as incompatible versions are brought in. The stable release of debian 12 (bookworm) was released on March 15th so this issue has popped up relatively recent. I will open an issue as well on if this is the right long term fix

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
 
 
-FROM golang:1.23 AS golang
+FROM golang:1.23-bullseye AS golang
 
 
 # RUN ./scripts/docker/install-go.sh


sgoggins:
I want to ensure this won't break anything that @JohnStrunk did to make Augur deployable on OpenShift. And I don't know enough to know if that is a risk or not. The change "looks" benign to something like that; which are the last words most people who crash systems utter.

cdolfi:
@sgoggins good call. I will say this is currently breaking/causing immense amount of failures if you build post 03/15 (release date of new version) so we will want to find a solution ASAP even if this isn't the one. We are using this change on our instance

JohnStrunk (approved):
This should be fine.
I would have expected static golang binaries, so I'm a bit puzzled about why the version of debian would matter. @cdolfi Is there a copy of the errors somewhere?

JohnStrunk:
I'm going to suggest we try the following:
In the current version of main, we've switched to building directly via `go install`. Instead of switching the debian version, let's try building static binaries by adding:
`ENV CGO_ENABLED=0` right below the `ENV GOBIN=/`.
This breaks the dependency on the glibc version between the builder and final containers. There are some limitations when disabling cgo, so I need someone to test that scc and scorecard both still work in the final image.

In my local branch, w/ the CGO change, I can at least run the `version` command on both of them as opposed to getting the glibc version error.

cdolfi:
@JohnStrunk Im trying to remember them exactly, the issue was with these https://github.com/cdolfi/augur/blob/2da45ae478c0beef821ca2247b942f058b3113fd/docker/backend/Dockerfile#L123-L124. SCC and scorecard were both causing a lot of errors to occur. Ill build current main and see what happens, then add the change you have described above. Would you be able to put the change you are describing in a code block for I can get the indentions and spacing right?

cdolfi:
@sgoggins @JohnStrunk Tried to build main and got the following 

`Step 17/38 : COPY --chmod=u=rw,u+X,go=r,go+X ./README.md .
the --chmod option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled`

Ill wait until the AM and I hear from @GregSutcliffe on how to handle this

GregSutcliffe:
>  I would have expected static golang binaries, so I'm a bit puzzled about why the version of debian would matter. @cdolfi Is there a copy of the errors somewhere?

@JohnStrunk they are static, but Debian 11 uses GLIBC 2.31, Debain 12 has GLIBC 2.32, so the binary cannot run.

GregSutcliffe:
Posting at midnight means I forgot some points, let me expand:
- The issue is a difference in GLIBC between the builder and the execution. I don't have the exact output to hand, but you'll get a symbol error if you try to run scc or scorecard in the augur container.
- The change to the GO build was written by @MoralCode in Feb. At that time Debian 11 was the default, so the Go build container matched the Augur container.
- Debain 12 was released on March 15th, at this point the "1.23" Go container now points to Bookworm which uses a different GLIBC
- So right now, as we don't target a specific Debian version for Go but do for Augur, the binaries created by the Go builder can't be executed by the Augur container

Right now, no one can build the container with working SCC / Scorecard. This patch *restores* the functionality as originally working & tested by ensuring we use the Bullseye version of Go 1.23. Since it's (a) broken and (b) not changing anything, I regard this is both important and not risky. Further improvements can of course be made, but bikeshedding on this delays fixes for end users. Fix the break, *then* improve the Docekrfile by all means.

BTW, Scorecard won't run without a further change - Augur is expecting to find a directory at /root/scorecard (just like SCC). This patch will fix it:
```
@@ -121,7 +121,7 @@ RUN set -x \
 RUN ./scripts/docker/install-workers-deps.sh
 
 COPY --from=golang "/root/scc" "/root/scc"
-COPY --from=golang "/root/scorecard/scorecard" "/root/scorecard"
+COPY --from=golang "/root/scorecard" "/root/scorecard"
 
 # RUN ./scripts/install/workers.sh 
```

Both this and Cali's change have been deployed to our prod instance and have resolved the errros with SCC and Scorecard

GregSutcliffe:
I fired up the vanilla build of 0.86.0 and reproduced the error, here you go:
```
$ docker run -it augur:v0.86.0 -c bash
root@73fa29492449:/augur# /root/scc/scc
/root/scc/scc: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /root/scc/scc)
/root/scc/scc: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /root/scc/scc)
root@73fa29492449:/augur# /root/scorecard 
/root/scorecard: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /root/scorecard)
/root/scorecard: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /root/scorecard)
```

sgoggins (dismissed):

sgoggins (approved):

cdolfi:
@sgoggins To summarize the conversion above, @JohnStrunk has suggested a completely different change to solve the issue, but I was unable to test whether that change works, or if this one is necessary or fixes the problem with the updates from dev. I opened this PR before dev was merged into main. I opened a different issue detailing that I was not able to test out this problem as I ran into a completely different issue trying to run the main branch with the updates from dev. #3146

PR #3148, opened 2025-05-05
Switch from venv/pip to uv

JohnStrunk:
**Summary**
This switches from using venv and pip for managing the python dependencies to using [uv](https://docs.astral.sh/uv/)

**Benefits**
- uv is much faster
- uv maintains a lockfile (`uv.lock`) that includes the exact versions of all dependencies for the project, not just direct ones. This ensures that the environment is reproducible.
- We now get consistent versions of all dependencies (previously, the order in which the `setup.py` scripts were called led to different package versions being installed)

**Changes**
- Project direct dependencies are now in `pyproject.toml`, and all `setup.py` files have been removed. I did my best to migrate the metadata across where possible. setuptools is still used for the actual install.
- The python version is now locked for the project via `.python-version`. uv will install this version of python when it creates the environment, so it no longer matters what python version is installed on the user's machine.
- The "sub-packages" from `augur/tasks/data_analysis` have been pulled into the main project. I suspect this may be controversial, and I can separate them back out if necessary.
- Backend Dockerfile has been updated to use uv based on uv's best-practices docs.

Assuming you all are good w/ this change, I'd like to update the docs prior to merge.

To try this out, the docker version should work as before, but if you're running outside of docker, you should:
- Install `uv`: `$ pip install --user uv`
- Prefix your "augur" commands w/ `uv run`, e.g., `$ uv run augur backend start`.

It's no longer necessary to manage the virtual environment or use pip directly, though you probably want to delete your old venv prior to starting w/ uv.

**To do**
- [x] GH action to ensure the lockfile is up-to-date
- [x] Documentation


JohnStrunk:
@sgoggins @Ulincsys Please take a look when you get a chance. If you're good w/ it, I'll get to work on revising the docs.

erikerlandson (commented):
LGTM

sgoggins (dismissed):
@JohnStrunk : Candidly I am outside my element for being able to provide a meaningful review here. 

I do understand, though, that this will stabilize our dependencies and make them easier to maintain in one place. And there are other benefits you outline. 

I am inclined to merge this into dev and test it vigorously. 

@Ulincsys : Any thoughts?

JohnStrunk:
@sgoggins I've made the doc changes. This should be ready.

PR #3149, opened 2025-05-06
Fix gihtub ratelimit issues in facade contributor resolution

ABrain7710:
**Description**
- Ensure that the github search ratelimit is respected so we don't just keep using the key

**Signed commits**


IsaacMilarky (approved):

Ulincsys (approved):
LGTM

PR #3151, opened 2025-05-07
Add smoke test to PRs

JohnStrunk:
- Add smoke test that starts Augur using the Docker compose file with the images that were just built.
- Update the backend CLI to accept the `AUGUR_DISABLE_COLLECTION` environment variable.

**Notes for Reviewers**
I had to disable the collection since I'm not aware of a way to securely use a GH or GL token to actually test collection. Repo secrets aren't available to PRs from forked repos, and even if they were, it'd be trivial for a user to extract them w/ a malicious PR.


.github/workflows/build_docker.yml
```diff
@@ -5,9 +5,6 @@ on:
       - main
       - dev
   pull_request:
-    branches:
```
erikerlandson, line 8: Is removing branch spec here deliberate?

erikerlandson (commented):
Overall LGTM, added one other inline comment.

Possibly out of scope for this PR but should maybe fix these warnings:
```
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value"
```

.github/workflows/build_docker.yml
JohnStrunk: Yeah. I see no reason to skip these tests for any PR. In the event we get a long-lived feature branch, this test should probably still run.

JohnStrunk:
> Possibly out of scope for this PR but should maybe fix these warnings:
> 
> ```
> LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value"
> ```

#3154

PR #3153, opened 2025-05-17
A collection of fixes for identified/new data anomalies from GitHub API

sgoggins:
**Description**
-  This fixes errors when a PR is collected and deleted, or when a PR contains no commit records. This is an occasional anomaly we need to address. 
-  This fixes errors on `facade_contributors` logic with tighter alignment with the API and myriad of possible values returned. Also added error handling. 
- Libyear Dependencies had some issues with emerging variance in PyPi declarations. The known possibilities are now handled. 
- GithubDataAccess: Added error handling to function in cases where there are PRs with missing commits, or (PRs or Issues that were deleted before we go get the related messages). It is worth noting that we do **not** have any issues for PR files, which is because the GraphQL API we use there returns "None" if there are no files, instead of a 404 like the commits endpoint.
- repo_info: Added code to handle emergent new ways or structures for data returned (or not returned). 
- Fixed merge issue in backend/Dockerfile with help from @JohnStrunk 
- analyze_commit methods updated to handle missing data

This PR fixes #

All of the fixes were identified using a large GitHub data collection test set where we find emergent anomalies. 

**Signed commits**


sgoggins:
I think the Docker issue is related to storing secrets OR it could be that the golang-builder image we are referencing doesn't exist anymore. @JohnStrunk ... what do you think?

JohnStrunk:
> I think the Docker issue is related to storing secrets OR it could be that the golang-builder image we are referencing doesn't exist anymore. @JohnStrunk ... what do you think?

There was a bad merge along the line someplace. In the backend Dockerfile, the name of the stage that builds the golang executables doesn't match the name being used to copy them into the final image:

```
FROM golang:1.23-bullseye AS golang
```
vs
```
COPY --from=golang-builder --chmod=u=rw,u+X,go=r,go+X "/scc" "/scc/scc"
```

Change the FROM line to be `... AS golang-builder`

augur/tasks/github/messages.py
```diff
@@ -114,22 +114,26 @@ def process_large_issue_and_pr_message_collection(repo_id, repo_git: str, logger
...
+            all_data += messages
         except UrlNotFoundException as e:
-            logger.warning(e)
-            continue
-
-        all_data += messages
+            logger.warning(f"{task_name}: 404 on comment URL {comment_url}. Skipping. Reason: {e}")
+            skipped_urls += 1
+        except Exception as e:
+            logger.error(f"{task_name}: Unexpected error on comment URL {comment_url}: {e}", exc_info=True)
```
ABrain7710, line 127: We don't want this because this will potentially obfuscate errors that are happening. We could potentially do this in the future, but right now we don't really look at logs to see the health of collection, we look at flower. So this would hide the error from flower

augur/tasks/github/pull_requests/commits_model/core.py
```diff
@@ -50,23 +50,32 @@ def pull_request_commits_model(repo_id,logger, augur_db, key_auth, full_collecti
...
+                    'tool_source': 'pull_request_commits_model',
+                    'tool_version': '0.41',
+                    'data_source': 'GitHub API',
+                    'repo_id': repo.repo_id,
+                }
+                all_data.append(pr_commit_row)
+        except UrlNotFoundException:
+            logger.warning(f"{task_name}: PR #{index + 1} returned 404 on commit data. Skipping.")
+        except Exception as e:
+            logger.error(f"{task_name}: Unexpected error while processing PR #{index + 1}: {e}", exc_info=True)
```
ABrain7710, line 76: Same here

augur/tasks/github/util/github_data_access.py
```diff
@@ -111,10 +111,26 @@ def make_request(self, url, method="GET", timeout=100):
...
+                query = parsed.query.lower()
+
+                if any(k in path for k in ["commits", "files", "comments"]) or \
+                any(k in query for k in ["commits", "files", "comments"]):
+                    self.logger.warning(
+                        f"Github response with 404 for PR files, PR commits or messages. "
+                        f"This is a data anomaly in the platform API, not an error. URL: {url}. Response: {response.text}"
+                    )
+                else:
+                    raise UrlNotFoundException(f"Could not find {url}")
```
ABrain7710, line 133: This logic should live in the pr files, pr commits, and pt comments collection files. 

It's not the responsibility of GitHub data access to understand how to handle responses from the GitHub api. Its responsibilities are to communicate the result of data retrieval (success by not throwing an exception or failure by throwing an exception), handle authentication and ensure fault tolerance from rate limits.

Essentially each of those needs to wrap the call to GitHub data access in a try catch and the specifically catch the UrlNotFoundException. Then the the catch you can do whatever logic that's needed for the specific data that's being collected.

It looks like it is already handled for files and messages, but not for commits

ABrain7710 (changes requested):

augur/tasks/github/repo_info/core.py
```diff
@@ -54,7 +54,32 @@ def get_repo_data(logger, url, response):
...
+    if 'errors' in data:
+        logger.error(f"GitHub API returned errors: {data['errors']}")
+        raise Exception(f"GitHub returned error response! {data['errors']}")
+
+    if 'id' not in data and 'message' in data:
+        logger.warning(f"Unexpected response structure from {url}: {data}")
+        if data['message'] == 'Not Found':
+            raise Exception(f"GitHub repo was not found or does not exist for endpoint: {url}")
+
+    return data
```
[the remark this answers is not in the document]
sgoggins: So rewrite the repo_info logic to hit graphql instead of the REST API?

augur/tasks/github/util/github_data_access.py
ABrain7710: Are you saying when we request all messages for a repo and it doesn't have any it will return a 404? 

If so we should catch the UrlNotFoundException in the call to /<owner>/<repo>/messages

augur/tasks/github/repo_info/core.py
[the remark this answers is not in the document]
ABrain7710: I didn't realize this was using the GitHub api. If it is then it should be using the GithubDataAccess class

PR #3154, opened 2025-05-21
Fix DB dockerfile env var warning

JohnStrunk:
**Description**

Fixed warning about "Legacy key/value format with whitespace separator should not be used" in Dockerfiles

This PR fixes #

**Notes for Reviewers**

**Signed commits**


sgoggins (approved):
I've seen this error and agree it should be fixed. Thanks @JohnStrunk !!

PR #3169, opened 2025-06-03
Tidy up some items around exception handling and repo_info logic

ABrain7710:
**Signed commits**

### DIFF
--- a/augur/application/schema/alembic/versions/32_update_openssf_deps.py
+++ b/augur/application/schema/alembic/versions/32_update_openssf_deps.py
@@ -1,0 +1,45 @@
+"""Remove old OpenSSF Scorecard Unique Constraint; Add new one.
+
+Revision ID: 32
+Revises: 31
+Create Date: 2025-06-06
+
+"""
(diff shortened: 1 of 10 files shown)


sgoggins (approved):
I think this fixes the design issues raised in the recent PR I merged?

sgoggins:
The fixes are causing files 404 failures to error out all of secondary collection, which was not happening prior to this patch. 

So, any time there is a file not at the endpoint, the whole secondary collection fails for the repo. These missing files are common, and not errors. We need a solution that solves the issue without propagating these as errors @ABrain7710 ...

sgoggins (approved):
I think we need a solution that doesn't cause the 404's for PRFiles, PRCommits or issue/pr messages to error out the entire secondary thread for a repository because these 404's legitimately reflect non-present artifacts.

ABrain7710:
@sgoggins Do you have a stack trace for these 404 errors? They should be handled but it is hard to tell what is happening without a stack trace

PR #3180, opened 2025-06-10
Add index on pull_request_files.repo_id

jberkus:
**Description**

Add an index on pull_request_files.repo_id, since that column supports a foriegn key, and it's a good idea to index any FK column on a large table unless it's extremely low cardinality.

This PR fixes #

No issue filed; change discussed in Slack.

**Notes for Reviewers**

Are there migrations somewhere I also need to update?  Or will folks just pick this up from base SQLAlchemy migrations?

**Signed commits**


jberkus:
Closing this, apparently y'all have a TON of trailing spaces in the file which my editor cleaned up automatically.

PR #3181, opened 2025-06-10
Add index pull_request_files.repo_id

jberkus:
**Description**

Add an index on pull_request_files.repo_id, since that column supports a foriegn key, and it's a good idea to index any FK column on a large table unless it's extremely low cardinality.

This PR fixes #

No issue filed; change discussed in Slack.

**Notes for Reviewers**

Are there migrations somewhere I also need to update?  Or will folks just pick this up from base SQLAlchemy migrations?

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -3052,6 +3052,7 @@ class PullRequestEvent(Base):
 class PullRequestFile(Base):
     __tablename__ = "pull_request_files"
     __table_args__ = (
+        Index("repo_id"),
         UniqueConstraint("pull_request_id", "repo_id", "pr_file_path"),
         {
             "schema": "augur_data",


augur/application/db/models/augur_data.py
```diff
@@ -3600,4 +3601,4 @@ class RepoClone(Base):
     count_clones = Column(BigInteger)
     clone_data_timestamp = Column(TIMESTAMP(precision=6))
 
-    repo = relationship("Repo")
\ No newline at end of file
+    repo = relationship("Repo")
```
jberkus, line 3604: sorry about that bit, here my Python checker is just removing a trailing newline.

Ulincsys:
We use Alembic for db migrations, so this change may also need to be added to a revision there.

Our revisions directory is located in `augur/application/schema/alembic/versions`

sgoggins:
LGTM

sgoggins:
PR #3184 is being tested to ensure there's an upgrade and downgrade path for this change.

sgoggins:
I made an incremental change following our usual strategy in PR #3184 ... Whenever Augur is installed, it runs the script referenced here, and then all of the update scripts since this was created. Keeping the ORM stable and aligned is easier if we add an increment "down the road" instead of changing the core install; especially for Augur instances currently in place. THANK YOU @jberkus ! The maintenance guideposts on our part need to be more clear.

jberkus:
Um, why did you close this PR?  What does it have to do with a psycopg2 update?

PR #3202, opened 2025-06-27
Initial conversion of main augur package to uv

JohnStrunk:
**Description**
- This revives #3148, targeting main.
- This supersedes (the uv portion of) #3198

This PR fixes #

**Notes for Reviewers**

**Signed commits**
- [ ] Yes, I signed my commits.

PR #3198, opened 2025-06-25
Fix docs --> Update to Main with Updated Build

sgoggins:
**Description**
- Placeholder for implementing these changes when the new Augur branch flow occurs.

PR #3183, opened 2025-06-11
Fix warnings/errors in docs & ensure clean builds

JohnStrunk:
**Description**
- This PR cleans up the documentation build, fixing all warnings and errors.
- It also adds a new check to PRs that builds the docs (unassociated w/ RTD's build process), stopping if any warnings are added to the documentation.

This PR fixes #3155 

**Notes for Reviewers**

**Signed commits**
- [x] Yes, I signed my commits.

PR #3203, opened 2025-06-27
Try 2: Fix warnings/errors in docs & ensure clean builds

JohnStrunk:
**Description**
- Depends on #3202 <== Merge that first!
- Re-targets #3183 at main
- Supersedes #3198


MoralCode:
> Depends on https://github.com/chaoss/augur/pull/3202 <== Merge that first!

I guess this is off-topic, but is this something you think Mergeify can help enforce for us in the future? IMO it would be nice to be able to have a red failing CI job present on open PRs that are somehow marked as dependent on other yet-to-be-merged PRs

MoralCode:
Also, it looks like this PR still contains some UV related changes - should those be moved to #3202 ?

JohnStrunk:
Yes, mergify can handle dependencies: https://docs.mergify.com/merge-protections/#depends-on (If I would have used the right syntax :cry:)

The uv-related commits are because this branch is build on top of my uv branch from #3202. Those will either go away or I can rebase once that merges. I originally tried to base this one on main directly, and I got a bunch of conflicts wrt uv.

MoralCode:
Docs build is failing due to a reference to `setup.py`

JohnStrunk:
> Docs build is failing due to a reference to `setup.py`

#3210 is the fix for the doc build on RTD

sgoggins (approved):
LGTM! Thank you!

docs/source/conf.py
```diff
@@ -49,7 +49,6 @@
     'sphinx_rtd_theme',
     'sphinxcontrib.openapi',
     'sphinxcontrib.redoc',
-    'sphinx.ext.autosectionlabel'
```
Ulincsys, line 52: Was this extension removed because it was unused, or for some other reason?

docs/source/index.rst
```diff
@@ -77,13 +77,13 @@ GSoC 2022 participants
 - `Mabel F <https://github.com/mabelbot>`_
 - `Priya Srivastava <https://github.com/Priya730>`_
 - `Ramya Kappagantu <https://github.com/RamyaKappagantu>`_
-- `Yash Prakash <https://gist.github.com/yash-yp>`_
+- `Yash Prakash <https://gist.github.com/yash-yp>`__
```
Ulincsys, line 80: What is the significance of the double underscore here?

docs/source/login.rst
```diff
@@ -73,7 +73,7 @@ The Client Application must make the following request in order to facilitate th
 
 The authorization server will respond with the following on success:
 
-.. code:: json
+.. code:: text
```
Ulincsys: Is there a particular reason to remove syntax highlighting for this block?

docs/source/conf.py
JohnStrunk: It was the source of many warnings about duplicate anchors since it creates an anchor w/ the name of every section. I think I only had to correct 1 or 2 links in all of the docs as a result of disabling this. The fix was to add manual anchors to the couple of sections.

docs/source/index.rst
JohnStrunk: It avoids creating a target, preventing warnings about duplicates. Explanation: https://stackoverflow.com/questions/27420317/restructured-text-rst-http-links-underscore-vs-use

docs/source/login.rst
JohnStrunk: It's not valid json, due to the `[integer:...]` line. That results in a warning:
```
/home/jstrunk/src/github.com/chaoss/augur/docs/source/login.rst:110: WARNING: Lexing literal_block '{\n    "status": "Validated",\n    "access_token": "the new Bearer token",\n    "refresh_token": "the new refresh token",\n    "token_type": "Bearer",\n    "expires": [integer: seconds until this access_token expires]\n}' as "json" resulted in an error at token: 'i'. Retrying in relaxed mode.
```

I chose to just make it text instead of try and figure out how to explain the integer parameter separately since comments aren't allowed in proper json either :cry:

Ulincsys: Does RST support HJSON highlighting? That format allows comments, so if it can then it may be useful to switch it to that

JohnStrunk: Looks like pygments will take json5. I switched it to that.

Ulincsys: Great! There's still one more text block in the file that needs updated to json5, but other than that it looks good to me

JohnStrunk: got them both now. :facepalm:

sgoggins (approved):
With all the recent changes I don't know if I should believe the build failure on READTHEDOCS.IO ... The regular docs are building, and nothing here changes anything so I think the failed docs build is likely a byproduct of something else that is out of synce.

PR #3204, opened 2025-06-27
Push the keyman container image

JohnStrunk:
**Description**
- Add keyman image and remove conditional login to GitHub Container Registry

Reopens #3193 against main. The original got closed when dev was removed.

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -145,6 +145,7 @@ jobs:
         image:
           - backend
           - database
+          - keyman
           - rabbitmq
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


MoralCode:
was the conditional login supposed to be a conditional push? I imagine we would only want to push updates to keyman for either merges to `main` or merges to `release`, right?

JohnStrunk:
The conditional login was a leftover from an earlier iteration of the workflow. That same condition actually gates the entire job: https://github.com/chaoss/augur/pull/3204/files#diff-c7f993a522ae12469fcfb15fc43197a8be4d9ef9e1350d98e775b2603eec9bd0R138-R139

The branch specification does need to be updated. #3200 will get that.

MoralCode:
Does this conflict with or depend on 3200? or can they both be merged independently?

JohnStrunk:
> Does this conflict with or depend on 3200? or can they both be merged independently?

Doesn't conflict, but given the choice, I'd do #3200 1st just to make sure the containers get tagged correctly when this goes in.

sgoggins (approved):
LGTM

ABrain7710 (approved):

Ulincsys (approved):
LGTM

PR #3201, opened 2025-06-25
Update documentation for new branching model

MoralCode:
**Description**
This updates as many docs as i can find through simple searches to correctly refer to the new branches

**Signed commits**
- [X] Yes, I signed my commits.

PR #3206, opened 2025-07-02
UPdate release branch? 

sgoggins:
**Description**
- We think the release branch would get updated before doing the release? Or, is mergify already activated?


MoralCode:
I don't understand what this PR is doing

It purports to be merging main into release, but it seems as though, in the current state of the repo (main is cf69b9ea9ebcbe235404d856c914947a8daa2a39, release is 47ceea4b941811df907569cc41f36857eff98e6d), release is only behind main by 3 merge commits (the merges for prs 3200, 3202, and 3204 in that order).

1) This PR seems like it contains just the straight up commits - is that GitHub's UI trying to be clever and unraveling each of the merges in the UI? 

2) Even if so, I think #3203 should be rebased and merged since it fixes the failing docs builds and depends on the now-merged 3202

3) I think #3201 should be in this release as it documents the new branching model. We dont want any new contributors that look at the release docs between now and the release after the one in this PR to be trying to use the old process or basing their changes on the now-deprecated `dev` branch

4) since this is a regular release, mergify should not be needed. According to my understanding, we were only planning to use mergify in the event of a hotfix, so that when a hotfix is proposed to be merged to main, we can set mergify to automatically open a PR from that feature branch to release _after_ it merges to main

JohnStrunk:
@MoralCode You're right that the hashes aren't the same. It appears to be that it's not including the merge commits, just the underlying commits themselves. I *think* this is ok for our purposes.

Another option would be to have a workflow that does a `git branch -f release main` to really set them equal.

MoralCode:
id rather check if theres an actual setting that controls that - mostly so we can turn it off and keep the commits organized/grouped by which merge/feature they were for to avoid confusion and/or weird bugs

sgoggins:
> There are unaddressed questions about what this PR aims to do

This PR is to create a release from what is in main @MoralCode

sgoggins:
> I don't understand what this PR is doing
> 
> It purports to be merging main into release, but it seems as though, in the current state of the repo (main is [cf69b9e](https://github.com/chaoss/augur/commit/cf69b9ea9ebcbe235404d856c914947a8daa2a39), release is [47ceea4](https://github.com/chaoss/augur/commit/47ceea4b941811df907569cc41f36857eff98e6d)), release is only behind main by 3 merge commits (the merges for prs 3200, 3202, and 3204 in that order).
> 
>     1. This PR seems like it contains just the straight up commits - is that GitHub's UI trying to be clever and unraveling each of the merges in the UI?
> 
>     2. Even if so, I think [Try 2: Fix warnings/errors in docs & ensure clean builds #3203](https://github.com/chaoss/augur/pull/3203) should be rebased and merged since it fixes the failing docs builds and depends on the now-merged 3202
> 
>     3. I think [Update documentation for new branching model #3201](https://github.com/chaoss/augur/pull/3201) should be in this release as it documents the new branching model. We dont want any new contributors that look at the release docs between now and the release after the one in this PR to be trying to use the old process or basing their changes on the now-deprecated `dev` branch
> 
>     4. since this is a regular release, mergify should not be needed. According to my understanding, we were only planning to use mergify in the event of a hotfix, so that when a hotfix is proposed to be merged to main, we can set mergify to automatically open a PR from that feature branch to release _after_ it merges to main

I think it is now more clear what the changes are, and the PRs are all in `main`.  

I recall the purpose of Mergify the same as you do. We have it enabled and welcome help getting that setup completed.

sgoggins:
We are currently conducting the high-volume collection test on `main` before the merge.

MoralCode:
> This PR is to create a release from what is in main @MoralCode

That part makes sense - I'm just worried about the risk of creating regressions given that this PR doesn't seem to line up with what's actually on main

> @MoralCode You're right that the hashes aren't the same. It appears that it's not including the merge commits, just the underlying commits themselves.

Did we ever find a solution to the issue of why the commits in this PR don't line up with what is on `main`? (i.e. the early history of this PR shows a bunch of commits that should have been part of merges)

Example: commit 706e5dcd217d096b0400e657be97979dbabaa5db should not be listed on this page as being directly part of this PR because it's a commit on a feature branch. It should instead be indirectly included via the merge commit 4e64a177dc33b8e48ecbcca9697d15a1a9910062 that was already approved and merged to main. If you ctrl-F this page for "706e5d" you can see it is directly part of this PR. I think something is wonky with this PR (maybe also related to the failing end to end test).

I would say, to be safe, it may be worth re-creating a new PR from main to release to see if that happens to fix both the commits and the E2E test

MoralCode (commented):
At this point I'm convinced it's just github weirdness. 

We should merge a PR to bump the version number and edit the changelogs and other metadata fixes at the time we want to cut this release, but I'll stop being the squeaky wheel

augur/tasks/github/util/github_data_access.py
```diff
@@ -145,13 +163,21 @@ def make_request_with_retries(self, url, method="GET", timeout=100):
...
+    def _decide_retry_policy(exception: Exception) -> bool:
+        """Defines whether or not to retry a failed request based on the exception thrown
+
+        Returns:
+            bool: Boolean describing whether or not the request should be retried
+        """
+        return not isinstance(num, (UrlNotFoundException, ResourceGoneException))
         
-    @retry(stop=stop_after_attempt(10), wait=wait_fixed(5), retry=retry_if_exception(lambda exc: not isinstance(exc, UrlNotFoundException)))
+    @retry(stop=stop_after_attempt(10), wait=wait_fixed(5), retry=retry_if_exception(_decide_retry_policy))
```
sgoggins, line 175: I think this may be causing an issue where not found URL's are causing breaking errors.

augur/api/gunicorn_conf.py
```diff
@@ -19,7 +20,22 @@
 workers = multiprocessing.cpu_count() * 2 + 1
 umask = 0o007
 reload = True
-reload_extra_files = glob(str(Path.cwd() / '**/*.j2'), recursive=True)
+
```
sgoggins, line 23: Maybe some issues here ... related to the disposal of database connections.

augur/tasks/github/util/github_data_access.py
```diff
@@ -22,6 +22,17 @@ class UrlNotFoundException(Exception):
 class NotAuthorizedException(Exception):
     pass
 
+class ResourceGoneException(Exception):
```
sgoggins, line 25: This is generating errors in testing. @MoralCode .. basically we are getting errors when these resource exceptions are hit and the job fails. Perhaps that is intended. Just checking.

keyman/KeyClient.py
```diff
@@ -150,14 +152,15 @@ class KeyPublisher:
...
         the publisher and the orchestrator.
     """
+    
     def __init__(self) -> None:
         # Load channel names and IDs from the spec
         for channel in spec["channels"]:
             # IE: self.ANNOUNCE = "augur-oauth-announce"
             setattr(self, channel["name"], channel["id"])
-            
+        self.conn = get_redis_connection()
```
sgoggins, line 161: This seems possibly wrong. @MoralCode / @Ulincsys ... we have changed the import from `as conn` to simply be the library. I don't see where `conn` exists as an import any longer. If this fails, it seems it could be the source of the database connection leaks on `main` right now.

keyman/KeyClient.py
```diff
@@ -172,7 +175,7 @@ def publish(self, key: str, platform: str):
             "key_platform": platform
         }
 
-        conn.publish(self.ANNOUNCE, json.dumps(message))
+        self.conn.publish(self.ANNOUNCE, json.dumps(message))
```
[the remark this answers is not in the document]
sgoggins: @MoralCode / @Ulincsys : Here as well, `conn` is referenced but I don't think it exists any longer.

keyman/KeyClient.py
```diff
@@ -1,4 +1,4 @@
-from augur.tasks.init.redis_connection import redis_connection as conn
+from augur.tasks.init.redis_connection import get_redis_connection
```
sgoggins, line 1: here is where the `as conn` is removed, but is still in use through the rest of the file.

keyman/KeyClient.py
```diff
@@ -215,7 +218,7 @@ def wait(self, timeout_seconds = 30, republish = False):
         }
         
         listen_delta = 0.1
-        conn.publish(self.ANNOUNCE, json.dumps(message))
+        self.conn.publish(self.ANNOUNCE, json.dumps(message))
```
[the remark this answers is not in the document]
sgoggins: no `conn`

keyman/Orchestrator.py
```diff
@@ -21,13 +21,15 @@
     logger.addHandler(handler)  # Attach the handler to the logger
     logger.setLevel(logging.DEBUG)
 else:
-    from augur.tasks.init.redis_connection import redis_connection as conn
+    from augur.tasks.init.redis_connection import get_redis_connection
     from augur.application.logs import AugurLogger
 
     logger = AugurLogger("KeyOrchestrator").get_logger()
+    conn = get_redis_connection()
```
sgoggins, line 28: ok .. here `conn` exists as a call to a method.

keyman/KeyClient.py
```diff
@@ -29,8 +29,10 @@ def __init__(self, platform: str, logger: Logger):
         if not platform:
             raise ValueError("Platform must not be empty")
         
-        self.stdout = conn
-        self.stdin: PubSub = conn.pubsub(ignore_subscribe_messages = True)
+        self.conn = get_redis_connection()
```
sgoggins, line 32: The change here is moving the refernce from the library to this call to the `get_redis_connection()` method.

augur/tasks/github/util/github_data_access.py
MoralCode: fixed in #3235

augur/api/gunicorn_conf.py
MoralCode: Can you elaborate on the issues you are seeing? This is a gunicorn config, so it shouldn't be affecting DB connections

augur/tasks/github/util/github_data_access.py
MoralCode: I suspect this is fixed by #3235 (TL;DR my fault).

If a resource returns 410 Gone from github it is handled similarly to a 404 (i.e. bail out of the retry loop and keep going with collection rather than propagating the exception up to the worker and stopping everything)

sgoggins: I think you are right.

keyman/KeyClient.py
MoralCode: That sounds plausible to me. I remember seeing that this keyclient file got refactored, but maybe it needs a deeper look

sgoggins: Later on in my review it seemed like the object was getting used and reassigned in the new class made specifically for that ... I am suspicious of Redis in general, perhaps being used in a way that creates a database connection but doesn't let go of it ... or the process loses it ... Like I said, there is no change that appears to directly alter the way we are connecting to the database ... which is everyone's favorite type of bug, right?

augur/api/gunicorn_conf.py
sgoggins: @MoralCode : In short, I have my POstgresql instance set to 3,000 connections .. and within one hour of starting Augur, they are all consumed. Which is bad in myriad ways, right? For comparison, the version of Augur in release uses max 200 connections over weeks.

Issue #1199, opened 2021-03-20
Implement the hdpmodel module from gensim for topic modeling in the clustering_worker


sgoggins:
There is significant diversity in the optimal number of topics in a topic model for any given collection of repositories. The hdpmodel module from gensim accommodates this diversity by determining the optimal number of topics for a given collection. Currently, it is preset, and could easily be configurable. However, the number is going to be different for each collection of repositories we do topic modeling for; ergo, and automated way to determine that number is a better approach than what Augur currently does. 

**Potential solutions:**
The description in this post makes it clear that the hdpmodel from gensim will be a better approach, though other implementations of this optimization exist:  https://www.kaggle.com/akashram/topic-modeling-intro-implementation

@sarit-adh may be able to do this as the original author of the worker. 

The current version of the worker is in the `main` branch, though will likely be in the release on Monday, April 8th, 2022.

sgoggins:
Here's another useful link: https://towardsdatascience.com/unsupervised-nlp-topic-models-as-a-supervised-learning-input-cf8ee9e5cf28

sgoggins:
Another useful link: https://nlpforhackers.io/topic-modeling/

sgoggins:
And Another: https://www.machinelearningplus.com/nlp/topic-modeling-gensim-python/#15visualizethetopicskeywords

WhiteWolf47:
I want to work on this issue, I have never used topic modelling. Where can i find the data?

WhiteWolf47:
@sgoggins lets discuss this issue!

WhiteWolf47:
@sgoggins okay, so we are looking forward to replace the current worker model with a hdpmodel (task model). I dont really know how we get the data and what we do with the analyse data, I have searched about hdpmodel and tried its basic implementation , so lets discuss this in depth on saturday (during the CHAOSS Software meeting)

sgoggins:
@WhiteWolf47 : if you are still interested I'll tag this as a first-timers-only issue which gets you direct support and access to the maintainers.

sgoggins:
Turns out this is not the best approach we could come up with! #fixed.

Tahamallick:
I'm interested in contributing to this issue as part of GSoC microtasks. Is anyone actively working on it?
If it's available, can I take this up?

Godzilaa:
@sgoggins  can i contribute anything on this or is it done if theree is to contribute pplease assign me this

PR #3214, opened 2025-07-07
Topic model versioning 3207

### LINKED ISSUE #3207: feat: Add topic model versioning and visualization (dummy data prototype)

Currently, Augur's topic modeling functionality does not support model versioning, metadata tracking, or visualization of topic models. This leads to several issues:
- It is difficult to track which model version produced which results.
- There is no way to compare different model runs or configurations.
- Users cannot easily visualize or interpret the topics generated by the model.
- Model evaluation metrics (such as coherence and perplexity) are not stored or accessible.

**Potential solutions:**

- Add a new `TopicModelMeta` table to store metadata for each topic model run (including method, parameters, scores, file paths, etc.).
- Add a `model_id` foreign key to the `TopicWord` and `RepoTopic` tables to associate words and topic distributions with a specific model version.
- Save model artifacts (pickle files, pyLDAvis HTML, wordcloud images) in a versioned directory structure under `artifacts/`.
- Generate and store topic visualizations (pyLDAvis, wordclouds) for each model.
- Store evaluation metrics (coherence, perplexity) in the database for each model.
- Provide helper functions to compare different models and retrieve their metadata.

**Additional context:**

- This feature is currently prototyped and tested on dummy data, not yet validated on production data.
- Main code changes are in:
  - `augur/application/db/models/augur_data.py`
  - `augur/application/db/models/__init__.py`
  - `augur/tasks/data_analysis/clustering_worker/tasks.py`
- Example visualizations and model files are saved in the `artifacts/` directory.
- The goal is to make topic modeling results more reproducible, comparable, and interpretable for users and researchers.

Xiaoha-cloud:
**Description**
- Integrate HDP-based topic number estimation into clustering_worker, enabling automatic topic count for LDA based on real data instead of a fixed preset

This PR fixes #3207

**Notes for Reviewers**
- The HDP model is used to estimate the optimal number of topics before LDA training, making the topic modeling pipeline fully data-driven and adaptive.

**Signed commits**
-[x] Yes, I signed my commits.


xiaoha-cloud:
Thanks @sgoggins for pointing this out! 

I've reviewed the current setup and understand that dependency management has moved to `pyproject.toml` with UV for locking. I'll go ahead and update the PR by removing the `setup.py` change and adding `pyLDAvis` under `[tool.poetry.dependencies]` in `pyproject.toml`.

After making that change, I will have a quick sync with @MoralCode to confirm the dependency is added correctly and follows the updated standards.

Thanks again for the feedback!

sgoggins:
@Xiaoha-cloud : You need to follow the instructions for reconciling the conflict with main. This reconciliation in your fork is possibly going to break your instance until we reconcile other changes to main. 

The "main thing" that needs to happen is that we have gotten rid of the setup.py architecture ... So, that needs to go ... Then we need to add things to pyproject.toml, BUT ... 

BEFORE you add things to pyproject.toml we need to fix the conflict ... then I need to talk to @JohnStrunk or @MoralCode to determine HOW to modify `pyproject.toml` and `uv.lock` ...

xiaoha-cloud:
Thanks again @sgoggins for the clarification!

I've noted that the `setup.py` architecture has been deprecated and that we should now use `pyproject.toml` + `uv.lock` for all dependency management.

 Next Steps I’ll Take:
1. Resolve the merge conflict with `main`, especially removing the `setup.py` file from `clustering_worker`.
2. Hold off on editing `pyproject.toml` or `uv.lock` until the conflict is fully resolved and I’ve synced with @MoralCode or @JohnStrunk for the correct modification process.
3. Test my instance locally post-merge to ensure the refactored worker is still functioning as expected.

MoralCode:
@Xiaoha-cloud Could you instead rebase your branch so that it is based on the current `main` branch? It seems like this PR is based on an augur version from April.

Doing so would also pretty much do all of the `setup.py` -> `uv` migration for you, so all you should need to change is to adjust your "Add pyLDAvis==3.3.1 to clustering_worker dependencies" to instead commit the results generated by `uv add pyLDAvis==3.3.1`

augur/tasks/init/celery_app.py
```diff
@@ -256,6 +256,10 @@ def setup_periodic_tasks(sender, **kwargs):
         one_day_in_seconds = 24*60*60
         sender.add_periodic_task(one_day_in_seconds, create_collection_status_records.s())
 
+        one_hour_in_seconds = 60*60
+        sender.add_periodic_task(one_hour_in_seconds, process_contributors.s())
```
MoralCode: Looks like this may in fact be a commit from Andrew that got mistakenly included in the rebase

xiaoha-cloud:
@MoralCode The Docker build is failing due to geckodriver version resolution returning null from GitHub API, resulting in 404 errors.

.github/workflows/build_docker.yml
```diff
@@ -344,7 +344,7 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64
           # Only push if we've tagged the image in the metadata step
-          push: ${{ steps.meta.outputs.tags != '' }}
+          push: ${{ github.event_name != 'pull_request' && steps.meta.outputs.tags != '' }}
```
MoralCode: Why was this changed?

augur/application/db/models/augur_data.py
```diff
@@ -3197,7 +3201,8 @@ class PullRequestMessageRef(Base):
             onupdate="CASCADE",
             deferrable=True,
             initially="DEFERRED",
-        )
+        ),
+        nullable=False,
```
MoralCode: This seems to make a change to the nullable status of another table. Was there a particular reason for this change? How would you recommend existing production databses, some with tens or hundreds of thousands of repos perform this migration if their tables have null values in them?

augur/application/db/models/augur_data.py
```diff
@@ -27,6 +27,8 @@
 import re
 import json
 import urllib.parse
+import pyLDAvis
+import numpy as np
```
MoralCode: You seem to be importing numpy and PyLDAvis in a models file. Are these intended for performing business logic in the models? As far as I can tell they seem unused in this file

tox.ini
MoralCode, line 1: this file seems to be for unit testing - why was it removed as part of this PR?

augur/schema/create_tables.sql
MoralCode, line 1: It seems like you created a new file with a generic `create_tables` name that only creates your tables. Was this just something you were using in development? or is it part of how this PR operates?

TOPIC_MODEL_MIGRATION_README.md
MoralCode, line 1: Love to see that this is documented - i just wonder whether a README.md in the root of the repo is the best place for it. I'd recommend either:
- having it be in comments within the migration file itself
- putting it in the readthedocs (we may need to make a new subsection for migrations if theres other migrations that could use documenting

TOPIC_MODEL_MIGRATION_README.md
```diff
@@ -0,0 +1,148 @@
+# Topic Model Meta Migration
+
+This migration adds support for topic modeling metadata and versioning to the Augur database.
```
MoralCode: I think it would be nice to include documentation on what topic modeling is and/or why it may be helpful to the augur codebase (and ideally "how" if theres anything special happening there) in addition to just "what" happened.

As someone who hasn't interacted with topic modeling before, I think it would be nice to ensure that the documentation is written with non-experts in mind. It seems to me like the goal of this is to automatically (rather than manually) group items in augur, such as repositories or words from issue/PR comments, into topic categories. It would be great to have your help in understanding whether or not this is correct and making sure the knowledge is documented for future contributors.

MoralCode (changes requested):

MoralCode:
This seems maybe related to https://github.com/chaoss/augur/issues/1199

.github/workflows/build_docker.yml
xiaoha-cloud: I added this condition to prevent Docker images from being pushed during pull request builds, which helps keep the registry clean and avoids publishing images for unreviewed code.
If the community has a different workflow or specific requirements (such as pushing images for PR previews), this change can be adjusted or reverted as needed.
Please advise on the preferred approach.

augur/schema/create_tables.sql
xiaoha-cloud: This file was created for local development and testing purposes, to quickly set up the new topic modeling tables and fields. It is not intended to be part of the production schema or the main migration process. I have removed it from the PR to avoid confusion.

augur/application/db/models/augur_data.py
xiaoha-cloud: The nullable=False settings in TopicModelMeta are for a new table introduced by this migration. These fields are required for each topic model record and do not affect any existing tables or data. No changes were made to the nullable status of existing columns in production tables.

xiaoha-cloud:
- Removed unused imports from the models file.
- Deleted development-only SQL file.
 - Restored tox.ini for unit testing.
- As suggested, I have moved the topic model migration documentation to docs/source/migrations/topic_model_meta.rst.
 - No changes were made to the nullable status of existing columns in production tables.
- I have also added an introductory section explaining what topic modeling is, why it is useful for Augur, and how it works, written for non-experts.
- This PR directly addresses the requirements described in #1199 by implementing HDP-based topic number estimation in the clustering_worker. With this change, the number of topics for LDA is now determined automatically based on the data, rather than being preset, which should improve the flexibility and accuracy of topic modeling in Augur.

.github/workflows/build_docker.yml
MoralCode: This is already filtered for on this line https://github.com/chaoss/augur/blob/2f860d8cae2ba95bd2450f5987dee7e9a14f30c5/.github/workflows/build_docker.yml#L292

augur/application/db/models/augur_data.py
MoralCode: it looks like this particular line is affecting a class called `PullRequestMessageRef` - is that one that you added?

MoralCode:
> As suggested, I have moved the topic model migration documentation to docs/source/migrations/topic_model_meta.rst.

Is this added to the nav pages in readthedocs so that someone can navigate to it (do we even have a section in the docs currently for documenting migrations? 

CC @sgoggins for opinions on what to do with this particular page of docs

augur/application/schema/alembic/versions/32_add_topic_model_meta_table.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+from sqlalchemy import text
+
+
+# revision identifiers, used by Alembic.
+revision = '32'
+down_revision = '31'
```
MoralCode: It seems like this PR has two revisions numbered 32 - is that intentional? Seems incorrect to me

sgoggins:
@MoralCode and @Xiaoha-cloud : I ran this on OSX and also ran the current version of main on OSX. Locally, both complain about tensorflow not being installed, and both also run. Frankly I am confused about how this could be the case since the OSX test is passing on this PR, but the end to end ones are not. 

The end to end testing fails because of a timeout. I am wondering if the failure is actually a timeout issue.

sgoggins:
A few other notes: 

- I had to run `uv lock` locally to get this branch to run
- When I run `make docs` locally I get no errors.

docs/source/index.rst
```diff
@@ -17,6 +18,8 @@ Augur Documentation
    schema/toc
    login
    procedures/toc
+   migrations/toc
```
MoralCode: You added a reference to a file that doesnt exist - this is causing a warning in the docs build (and presumably causing it to fail)

Might be nice to add a toc page (similar to the other TOC pages) to act as both a Table of contents for the migrations pages and also a general page explaining what the migrations sections of the docs are for

MoralCode: `/home/runner/work/augur/augur/docs/source/index.rst:7: WARNING: toctree contains reference to nonexisting document 'migrations/toc'`

xiaoha-cloud:
Hey @MoralCode !
Task 1: NMF+Count vs LDA Performance
-  Replaced HDP with NMF+Count algorithm
-  85.1% faster than LDA (0.020s vs 0.135s)
-  92.9% better topic coherence (0.915 vs 0.474)
-  8.09x speedup with larger datasets
Task 2: JSON Configuration
-   Replaced complex database config with config.example.json
-   Easier for team members to configure topic modeling parameters
Task 3: Cross-Platform TensorFlow
-   Linux: tensorflow>=2.15.0,<2.16.0
-   Apple Silicon: tensorflow-macos + tensorflow-metal
-   Using uv dependency groups for platform compatibility
Test Results: All 37 tests passed (22 unit + 6 integration + 6 system + 3 performance)
Thanks for helping with rebase and local testing! This version should run successfully now.

xiaoha-cloud:
Superseded by #3254 . Closing this PR to reduce review noise. The new PR rebases on current main, migrates dependencies to pyproject.toml + uv, adds migrations 34/35, implements NMF with 4‑dimension auto‑retraining, provides API + web UI, and addresses prior review feedback.

Issue #3212, opened 2025-07-03
podman compose unable to run docker lifecycle hooks


MoralCode:
When trying to run the compose file, even with patches from #3211, i run into this error:

```
[augur]        | Python 3.11.13
[augur]        | 2025-07-03 13:39:57 a91d132b829c augur[1] INFO Flushing all redis databases this instance was using
[augur]        | Traceback (most recent call last):
[augur]        |   File "/augur/.venv/bin/augur", line 10, in <module>
[augur]        |     sys.exit(run())
[augur]        | (augur) Reseting log files
[augur]        | Failed to create file handler: [Errno 13] Permission denied: '/logs/augur.info'
[augur]        | Failed to create file handler: [Errno 13] Permission denied: '/logs/augur.error'
[augur]        |              ^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 1128, in __call__
[augur]        |     return self.main(*args, **kwargs)
[augur]        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 1053, in main
[augur]        |     rv = self.invoke(ctx)
[augur]        |          ^^^^^^^^^^^^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 1659, in invoke
[augur]        |     return _process_result(sub_ctx.command.invoke(sub_ctx))
[augur]        |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 1659, in invoke
[augur]        |     return _process_result(sub_ctx.command.invoke(sub_ctx))
[augur]        |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 1395, in invoke
[augur]        |     return ctx.invoke(self.callback, **ctx.params)
[augur]        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[augur]        |   File "/augur/.venv/lib/python3.11/site-packages/click/core.py", line 754, in invoke
[augur]        |     return __callback(*args, **kwargs)
[augur]        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #3195, opened 2025-06-24
Unable to build Augur in podman due to `--chmod`


MoralCode:
**Description:**
Due to some current bugs relating to podman's support of both symbolic notation in the `--chmod` argument to the dockerfile COPY instruction, and the handling of the `X` (only apply executable permission to directories) portion of the flag, Augur is unable to be built using podman tooling. 

**How to reproduce:**
1. Clone the augur repo in an environment with podman set up
2. follow instructions at https://oss-augur.readthedocs.io/en/main/docker/quick-start.html to build the container, substituting `docker` for `podman` in commands as needed

**Expected behavior:**
Since Podman is generally a drop-in replacement for docker, it should work the same

**Actual behavior:**

The podman build fails during a COPY step with an error like this:

`Error: building at STEP "COPY --chmod=u=rw,u+X,go=r,go+X ./README.md .": Error parsing chmod u=rw,u+X,go=r,go+X`

**Outstanding questions**

**Log files**
N/A

**Software versions:**
 - Augur: v0.89.3
 - OS: Fedora 41
 - Browser: N/A

**References**:

Upstream podman issue: https://github.com/containers/buildah/issues/6066

Issue/PR that caused this issue in augur: https://github.com/chaoss/augur/issues/3107

MoralCode:
based on my testing, using the octal notation resolves the issue with building in podman, however this comes at the cost of not-quite-identical semantics with regard to the `X` value affecting recursive copies (i.e. only make directories being copied executable, not every file within them).

enabling podman to support this symbolic format for `--chmod` arguments seems to be the best path forward for resolving this, however it may be more effort than the current desire to support podman is worth, so worst case I hope this issue just serves as documentation for the next person who runs into this


I also think, once this is fixed, it may be useful to add tests that cover the podman builds in a similar way to the current docker based CI jobs (even if their failure doesn't block a PR) just to create a point of additional information about whether future PRs cause issues for podman so issues can be fixed in code review

JohnStrunk:
It may also be possible to resolve this by replacing the copy's chmod with a fancy `RUN find ...` after the copy to properly search and set the permissions.

PR #3216, opened 2025-07-07
add podman copy of the E2E test

MoralCode:
**Description**
This PR adds a copy of the end to end CI job that uses podman instead of docker to ensure that any changes are compatible with both

**Signed commits**

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -52,7 +52,7 @@ jobs:
 
 
   test-e2e:
-    name: End-to-end test
+    name: End-to-end test (Docker)
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


MoralCode:
ok so the podman job is failing for expected reasons (i.e. #3195)

sgoggins (dismissed):
LGTM, and I don't really have a way to test this without merging it, so here we go! :D 

Update: Ah. I can see the PR has the podman test on it and its running right now. Will wait of course.

MoralCode:
@sgoggins note I'm expecting this is likely to fail due to #3212 as well - I don't currently have a good workaround for that lifecycle hook though (at least one that won't break the openshift usecase based on my knowledge of it).

sgoggins:
@MoralCode : Waiting to see what happens in the test, but looking at that issue I wonder if it is simply an issue of those two files not existing in the container yet, and maybe it could be solved with `touch` ? This could be all wrong, but I have seen failures like this before solved by simply making the file exist.

JohnStrunk:
buildah is using too much disk space:
```
Error: committing container for step {Env:[PATH=/augur/.venv/bin:/usr/bin/:/usr/local/bin:/usr/lib:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin LANG=C.UTF-8 GPG_KEY=A035C8C19219BA821ECEA86B64E628F8D684696D PYTHON_VERSION=3.11.13 PYTHON_SHA256=8fb5f9fbc7609fa822cb31549884575db7fd9657cbffb89510b5d7975963a83a DEBIAN_FRONTEND=noninteractive UV_COMPILE_BYTECODE=1 UV_LINK_MODE=copy UV_LOCKED=1 SCC_DIR=/scc SCORECARD_DIR=/scorecard] Command:cmd Args:[/init.sh] Flags:[] Attrs:map[json:true] Message:CMD /init.sh Original:CMD ["/init.sh"]}: copying layers and metadata for container "729ca85e50830fe3342a9a986119b6d26c3d62bbefa92c478ca33a070fb40cce": writing blob: adding layer with blob "sha256:eafd5cd469df775bd8d1792a90d4b7aaae47dbdb93bd80a8adecc5e506b65b69": processing tar file(write /scorecard/scorecard: no space left on device): exit status 1
Error: Error: buildah exited with code 1
```

Ulincsys (approved):
LGTM

Issue #3185, opened 2025-06-13
GitLab Data Collection Error for PR Metadata


sgoggins:
Please help us help you by filling out the following sections as thoroughly as you can.

**Description:**

Key Issue: Pretty clear there is some kind of logic error identifying the repo_id for a git_url for GitLab repositories. Likely buried inside the logic of `augur/tasks/gitlab/merge_request_task.py` and its calling of `augur/tasks/util/collection_util.py` 

Core error: 
```
psycopg2.errors.UndefinedFunction: operator does not exist: character varying = integer[]
LINE 3: WHERE augur_data.repo.repo_git = ARRAY[897,896,895,894,893,8...
                                       ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.


```

Whole Error Message: 
```
Traceback (most recent call last):
  File "/home/sean/github/virtualenv/ai/lib/python3.11/site-packages/celery/app/trace.py", line 451, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/home/sean/github/virtualenv/ai/lib/python3.11/site-packages/celery/app/trace.py", line 734, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sean/github/ai.chaoss/augur/tasks/gitlab/merge_request_task.py", line 306, in collect_merge_request_metadata
    process_mr_metadata(metadata_list, f"{owner}/{repo}: Mr metadata task", repo_id, logger, session)
  File "/home/sean/github/ai.chaoss/augur/tasks/gitlab/merge_request_task.py", line 339, in process_mr_metadata
    all_metadata.extend(extract_needed_mr_metadata(metadata, repo_id, pull_request_id, tool_source, tool_version, data_source))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sean/github/ai.chaoss/augur/application/db/data_parse.py", line 1091, in extract_needed_mr_metadata
    head = {'sha': mr_dict['diff_refs']['head_sha'],
                   ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable

[...]

PR #3217, opened 2025-07-08
Gitlab Mr Metadata Improvements

ABrain7710:
**Description**
- The issue #3185 shows a NoneType exception happenening when trying to retrieve the `head_sha`, so I fixed this by providing a default value of None if the diff_refs are missing or head_sha or base_sha.
- Additionally the issue pointed out that the failure handler was assuming that the first paramater to the task would be repo_git, but it this task it can't be the first paramater the celery chain is passing the mr_ids to the first param in the task. To fix this I implemented a method to search the methods parameters for the repo_git parameter. 

This PR fixes #
#3185

**Signed commits**

### DIFF
--- a/augur/tasks/init/celery_app.py
+++ b/augur/tasks/init/celery_app.py
@@ -5,6 +5,7 @@ from typing import List, Dict
 import os
 import datetime
 import traceback
+import inspect
 import celery
 from celery import Celery
 from celery import current_app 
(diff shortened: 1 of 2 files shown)


augur/tasks/init/celery_app.py
```diff
@@ -101,27 +102,43 @@ def augur_handle_task_failure(self,exc,task_id,repo_git,logger_name,collection_h
                 session.commit()
 
     def on_failure(self,exc,task_id,args, kwargs, einfo):
-        repo_git = args[0]
+        repo_git = self._extract_repo_git(args, kwargs)
         # log traceback to error file
         self.augur_handle_task_failure(exc, task_id, repo_git, "core_task_failure")
 
+    def _extract_repo_git(self, args, kwargs):
```
Ulincsys, line 109: Would it maybe be a better option to standardize the calling convention rather than dynamically locating the argument post-call? Or, is there some reason why enforcing call-side standardization is not possible?

PR #3222, opened 2025-07-09
Allow DB connection specification via components

JohnStrunk:
**Description**
This provides an additional method for specifying the DB connection information.

A user may now (instead of using the current `AUGUR_DB` connection string) specify the various connection string components separately: `AUGUR_DB_HOST`, `AUGUR_DB_USER`, `AUGUR_DB_PASSWORD`, `AUGUR_DB_NAME`

This change was made to support integration with the Postgres operator in Kubernetes environments. The operator handles these connection settings as separate fields within a Secret, and by allowing Augur to accept them individually, those Secret fields can be directly mapped into the Augur container.

### DIFF
--- a/docker/backend/entrypoint.sh
+++ b/docker/backend/entrypoint.sh
@@ -2,6 +2,16 @@
 #SPDX-License-Identifier: MIT
 set -e
 
+if [[ -z "$AUGUR_DB" ]]; then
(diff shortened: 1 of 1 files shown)


sgoggins (approved):
Thank you! Any step towards Kubernetes is a good one!

PR #3223, opened 2025-07-10
Removed Files Not Needed

sgoggins:
**Description**
- Some files existed in the root that no longer need to exist there. Removed. 

**Signed commits**


MoralCode (approved):
Seems like all files are either markdown files or are files that you originally added, so I'm pretty confident they arent inadvertently tied into anything.

LGTM

Ulincsys (approved):

PR #3224, opened 2025-07-11
Address warnings from linter

MoralCode:
**Description**
Fixes linter warnings to make one of the bots be quiet

This PR pretty much just removed unused imports and renames a couple variables to prevent shadowing

**Signed commits**

### DIFF
--- a/augur/application/cli/config.py
+++ b/augur/application/cli/config.py
@@ -237,7 +237,7 @@ def config_get_all_json():
 
             for row in session.query(Config).all():
                 data[row.section_name][row.setting_name] = row.value
-    except:
+    except Exception:
         pass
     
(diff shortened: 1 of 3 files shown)


augur/application/cli/config.py
Ulincsys, line 1: If possible, I would be greatly interested to figure out how to turn off the "Unnecessary else after X" warnings. IMO they are not helpful, and they encourage a reduction in readability.

MoralCode: If we dont already have one, we can add a pylint config file to the repo to define a set of linting rules to enforce.

https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html

PR #3229, opened 2025-07-15
deduplicate convert_type_of_value

MoralCode:
**Description**
This PR takes two identical functions found in the codebase and refactors them to both import that function from the `utils.py` file in the application module

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -4,8 +4,7 @@ import json
 from typing import List, Any, Optional
 import os
 from augur.application.db.models import Config 
-from augur.application.db.util import execute_session_query
-from augur.application.util import convert_type_of_value
+from augur.application.db.util import execute_session_query, convert_type_of_value
 
 def get_development_flag_from_config():
     
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -5,6 +5,7 @@ from typing import List, Any, Optional
 import os
 from augur.application.db.models import Config 
 from augur.application.db.util import execute_session_query
+from augur.application.util import convert_type_of_value
 
 def get_development_flag_from_config():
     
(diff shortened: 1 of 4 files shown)


MoralCode:
i wouldnt exactly call this a bug fix, more just cleanup of potentially old tech debt or something - this doesnt fix anything (augur should operate exactly the same with or without this) - it just helps make the code more maintainable

Ulincsys:
I think that the preferable location for this function would be in the `db/lib.py` file. The `augur.application.db` module is intended to be independent, and so it should not import anything from parent or sibling modules.

We struggled a lot for a long time with circular imports, and the practice of isolating this module was intended to fix that. So, instead, it's better for functionality required by the module to be included in the module, and anything outside of the module that needs it should import it from `db/`.

sgoggins:
@MoralCode : Have you seen @Ulincsys 's comment?

sgoggins:
> I think that the preferable location for this function would be in the `db/lib.py` file. The `augur.application.db` module is intended to be independent, and so it should not import anything from parent or sibling modules.
> 
> We struggled a lot for a long time with circular imports, and the practice of isolating this module was intended to fix that. So, instead, it's better for functionality required by the module to be included in the module, and anything outside of the module that needs it should import it from `db/`.

This comment ^ @MoralCode

MoralCode:
Yep, just been distracted and haven't gotten around to moving the function

MoralCode:
Oops, I don't think I was thinking when I made the code change, so I ended up moving it to db.utils.

It seems like it's a pretty standalone utility function, so it seems like it fits well there

Ulincsys (approved):
LGTM

sgoggins (approved):

Issue #3194, opened 2025-06-23
augur.tasks.github.releases.tasks.collect_releases: psycopg2.errors.UniqueViolation


cdolfi:
IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "releases_pkey"\nDETAIL: Key (release_id)=(RE_kwDOI7xefs4HLIld ) already exists.\n') 


```
[...]

PR #3245, opened 2025-07-27
fix: resolve UniqueViolation error in GitHub releases collection

mohsinm-dev:
Fix for GitHub Issue #3194 where releases collection was failing with psycopg2.errors.UniqueViolation on releases_pkey constraint.

Root cause: GitHub API returns release IDs with trailing spaces that don't match existing trimmed database records.

Changes:
- Add str().strip() to release_id processing in get_release_inf()
- Enhance duplicate detection in insert_release() with proper trimming
- Add early duplicate detection to prevent unnecessary database operations

**Description**

This PR fixes #

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/github/releases/core.py
+++ b/augur/tasks/github/releases/core.py
@@ -23,7 +23,7 @@ def get_release_inf(repo_id, release, tag_only):
 
 
         release_inf = {
-            'release_id': release['id'],
+            'release_id': str(release['id']).strip(),
             'repo_id': repo_id,
             'release_name': release['name'],
             'release_description': release['description'] if release['description'] is not None else '',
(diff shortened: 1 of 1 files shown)


augur/tasks/github/releases/core.py
```diff
@@ -23,7 +23,7 @@ def get_release_inf(repo_id, release, tag_only):
 
 
         release_inf = {
-            'release_id': release['id'],
+            'release_id': str(release['id']).strip(),
```
sgoggins, line 26: I think this helps to ensure that strings are interpreted as strings, and spaces are removed. So, it should eliminate "false duplicates".

sgoggins (approved):
Looks like this will identify "false duplicates".

sgoggins:
@cdolfi : I think this addresses false duplicate types of issues.

cdolfi (approved):
lgtm!

MoralCode:
Seems like this needs to be merged still? any reason it hasnt been?

PR #3253, opened 2025-08-27
add date filter to contributer resolution logic queries

IsaacMilarky:
**Description**
- Add a date filter to the Facade contributer resolution logic
- Now, we will only use commits that have appeared since last time facade was ran in contributer resolution
- I have also removed the start_date setting because it wasn't doing anything 
- Added a helper function for if we need to get the date that the last commit was added

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -112,8 +112,6 @@ def trim_commits_post_analysis_facade_task(repo_git):
     repo = repo = get_repo_by_repo_git(repo_git)
     repo_id = repo.repo_id
 
-    start_date = facade_helper.get_setting('start_date')
-    
     logger.info(f"Generating sequence for repo {repo_id}")
 
     repo = get_repo_by_repo_git(repo_git)
@@ -123,7 +121,7 @@ def trim_commits_post_analysis_facade_task(repo_git):
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc, start_date)
+    parent_commits = get_parent_commits_set(repo_loc)
 
     # Grab the existing commits from the database
     existing_commits = get_existing_commits_set(repo_id)
@@ -237,7 +235,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     repo = get_repo_by_repo_git(repo_git)
     repo_id = repo.repo_id
 
-    start_date = facade_helper.get_setting('start_date')
+    start_date = facade_helper.get_last_collected_commit_date(repo_id)#.get_setting('start_date')
 
     logger.info(f"Generating sequence for repo {repo_id}")
     
@@ -438,11 +436,6 @@ def generate_analysis_sequence(logger,repo_git, facade_helper):
 
     analysis_sequence = []
 
-    #repo_list = s.sql.text("""SELECT repo_id,repo_group_id,repo_path,repo_name FROM repo WHERE repo_git=:value""").bindparams(value=repo_git)
-    #repos = fetchall_data_from_sql_text(repo_list)
-
-    start_date = facade_helper.get_setting('start_date')
-
     #repo_ids = [repo['repo_id'] for repo in repos]
 
     #repo_id = repo_ids.pop(0)
@@ -472,8 +465,6 @@ def facade_phase(repo_git, full_collection):
     #Get the repo_id
     #repo_list = s.sql.text("""SELECT repo_id,repo_group_id,repo_path,repo_name FROM repo WHERE repo_git=:value""").bindparams(value=repo_git)
     #repos = fetchall_data_from_sql_text(repo_list)
-
-    start_date = facade_helper.get_setting('start_date')
 
     #repo_ids = [repo['repo_id'] for repo in repos]
 
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -235,8 +235,6 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     repo = get_repo_by_repo_git(repo_git)
     repo_id = repo.repo_id
 
-    start_date = facade_helper.get_last_collected_commit_date(repo_id)#.get_setting('start_date')
-
     logger.info(f"Generating sequence for repo {repo_id}")
     
     repo = get_repo_by_repo_id(repo_id)
@@ -246,7 +244,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc, start_date)
+    parent_commits = get_parent_commits_set(repo_loc)
 
     # Grab the existing commits from the database
     existing_commits = get_existing_commits_set(repo_id)
--- a/augur/tasks/git/util/facade_worker/facade_worker/config.py
+++ b/augur/tasks/git/util/facade_worker/facade_worker/config.py
@@ -244,6 +244,17 @@ class FacadeHelper():
             return
     def inc_repos_processed(self):
         self.repos_processed += 1
+    
+    def get_last_collected_commit_date(self,repo_id):
+        commit_date_query = s.sql.text("""
+        SELECT cmt_committer_timestamp FROM commits 
+        WHERE repo_id=:repo_id 
+        ORDER BY data_collection_date DESC
+        LIMIT 1;
+        """).bindparams(repo_id=repo_id)
+
+        result = execute_sql(commit_date_query).fetchone()
+        return result[0]
 
 """
 class FacadeConfig:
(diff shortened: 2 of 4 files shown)


augur/tasks/git/util/facade_worker/facade_worker/config.py
```diff
@@ -244,6 +244,17 @@ def insert_or_update_data(self, query, **bind_args)-> None:
             return
     def inc_repos_processed(self):
         self.repos_processed += 1
+    
+    def get_last_collected_commit_date(self,repo_id):
```
sgoggins: I suggest this logic include a pull back of at least 1 day from the most recent commit in the table, as that will ensure we don't miss mid-day commits. I am not sure that this is the right location for that logic. 

We could use the timestamp, but I trust timestamps less than I trust just moving the window back one day. 

Finally: There are a small number of repos where some dufus created a commit or two into the future. Like way into the future. So, that should be checked I think. Perhaps the most reliable failsafe would be to triangulate the last commit date with the `facade_data_last_collected` stamp in `augur_operations.collection_status` to ensure that something is not wildly out of whack. On inactive repos, the most recent commit timestamp could be years ago; but we could just use whichever date is farthest backward from today to ensure that committer misbehavior does not undermine data quality. 

Thank you @IsaacMilarky !! This is FANTASTIC!!

augur/tasks/github/facade_github/tasks.py
```diff
@@ -231,6 +236,7 @@ def insert_facade_contributors(self, repo_git):
                 commits
             WHERE
                 commits.repo_id = :repo_id
+                AND (:since_date is NULL OR commits.data_collection_date > :since_date)
```
sgoggins, line 242: This relies on the since_data not be garbaged up by dufus committers putting random dates in, similar to my other comment but possibly even more insane to troubleshoot in the future. I think it works if we undertake some of the bulletproofing I mentioned previously.

augur/tasks/github/facade_github/tasks.py
```diff
@@ -239,7 +245,7 @@ def insert_facade_contributors(self, repo_git):
                 commits.cmt_author_raw_email
             ORDER BY
             hash
-    """).bindparams(repo_id=repo_id)
+    """).bindparams(repo_id=repo_id,since_date=last_collected_date)
```
sgoggins, line 251: Making sure `last_collected_date` is in fact the date we calculate for the most recent commit in our record.

augur/tasks/github/facade_github/tasks.py
```diff
@@ -271,6 +276,7 @@ def insert_facade_contributors(self, repo_git):
             commits
         WHERE
             contributors.cntrb_canonical = commits.cmt_author_raw_email
+            AND (:since_date is NULL OR commits.data_collection_date > :since_date)
```
sgoggins, line 282: Making sure `last_collected_date` is in fact the date we calculate for the most recent commit in our record.

augur/tasks/github/facade_github/tasks.py
```diff
@@ -286,7 +292,8 @@ def insert_facade_contributors(self, repo_git):
             contributors_aliases.alias_email = commits.cmt_author_raw_email
                             AND contributors.cntrb_id = contributors_aliases.cntrb_id
             AND commits.repo_id = :repo_id
-    """).bindparams(repo_id=repo_id)
+            AND (:since_date is NULL OR commits.data_collection_date > :since_date)
+    """).bindparams(repo_id=repo_id,since_date=last_collected_date)
```
sgoggins, line 299: Making sure `last_collected_date` is in fact the date we calculate for the most recent commit in our record.

Makefile
```diff
@@ -129,7 +129,7 @@ test-api:
 #
 .PHONY: uv
 uv:
-	@ command -v uv >/dev/null 2>&1 || { echo "Installing uv..."; pip install --user uv; }
+	@ command -v uv >/dev/null 2>&1 || { echo "Installing uv..."; pip3 install uv; }
```
Ulincsys: This change was not referenced in your PR message. What purpose does it serve?

sgoggins (approved):
LGTM - Lets let it fly!

sgoggins:
The test would be: 
1. New Augur instance
2. Identify 3-5 very large repositories
3. Do collection for them (note that for Facade collection it typically requires an Augur restart)
4. Truncate the augur_operations.collection_status table
5. Restart collection after commits were collected the first time and see if it goes insanely fast, which it should if this branch does what we claim it does.

Makefile
sgoggins: @MoralCode ; This looks wrong.

MoralCode: id say maybe just revert this piece then? Seems like the makefile isn't really in use anyway

ABrain7710 (changes requested):
To ensure we can still support full collection on demand I think it would good to pass a full_collection flag and if it is true then grab all commits, otherwise do this new logic to only get new commits

MoralCode:
in poking through the code, im actually not sure this even affects analyze_commits_in_parallel beyond removing some date related things

Makefile
IsaacMilarky: Reverting this

IsaacMilarky:
> in poking through the code, im actually not sure this even affects analyze_commits_in_parallel beyond removing some date related things

The partial commit collection was already supported previously in the body of the `analyze_commits_in_parallel` method:

```python

    repo = get_repo_by_repo_id(repo_id)

    #Get the huge list of commits to process.
    absolute_path = get_absolute_repo_path(facade_helper.repo_base_directory, repo.repo_id, repo.repo_path, repo.repo_name)
    repo_loc = (f"{absolute_path}/.git")
    # Grab the parents of HEAD

    parent_commits = get_parent_commits_set(repo_loc)

    # Grab the existing commits from the database
    existing_commits = get_existing_commits_set(repo_id)

    # Find missing commits and add them
    missing_commits = parent_commits - existing_commits

    facade_helper.log_activity('Debug',f"Commits missing from repo {repo_id}: {len(missing_commits)}")
```

These changes just effect the contributor resolution logic to only go through commits that we haven't before by default.

MoralCode:
What would the behavior of this PR be in the case of a repo deciding to rewrite the history of their git repo? Is that something we can detect and maybe just fall back to the full recollection?

IsaacMilarky:
> What would the behavior of this PR be in the case of a repo deciding to rewrite the history of their git repo? Is that something we can detect and maybe just fall back to the full recollection?

The logic is based on git commit hashes for the main analysis logic and for the contributor resolution logic it's based on the date that the commit record was inserted. 

So, if a repository is rewriting their history the hashes would presumably be different and would be collected by the main logic into the commits table. Then that new commit would be collected later and therefore not be filtered by the new logic. The date that it is filtering by is the date that the commit record was collected not the date the commit record was committed.

augur/application/schema/alembic/versions/34_add_facade_config_full_recollect.py
```diff
@@ -0,0 +1,37 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    #Add toggle for facade collection.
+    conn = op.get_bind()
+
+    conn.execute(text(f"""
+        INSERT INTO "augur_operations"."config" ("section_name", "setting_name", "value", "type") VALUES ('Facade', 'facade_contributor_full_recollect', '{0}', 'int');
+        """))
+    # ### end Alembic commands ###
```
MoralCode: A migration is not needed if you are simply adding a value to the database - its really only needed for changing the structure of the DB tables/columns themselves. I suspect you could add this to augur/application/config.py:~71 or so and it would get included in the database for anyone who has augur set to insert that config on startup

IsaacMilarky: I gotcha, I just wasn't sure where we create the config in the code. Have changed to reflect.

augur/application/config.py
```diff
@@ -50,7 +50,8 @@ def get_development_flag():
                 "pull_repos": 1,
                 "rebuild_caches": 1,
                 "run_analysis": 1,
-                "run_facade_contributors": 1
+                "run_facade_contributors": 1,
+                "facade_contributor_full_recollect": 1
```
sgoggins: Added at @ABrain7710 's request. The goal is to have a specific option to ensure that previously uncollected contributors are gathered. Probably the default here should be 0 ...

augur/tasks/git/facade_tasks.py
```diff
@@ -123,7 +121,7 @@ def trim_commits_post_analysis_facade_task(repo_git):
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc, start_date)
+    parent_commits = get_parent_commits_set(repo_loc)
```
sgoggins, line 124: Per @MoralCode 's prior comment; The analyze commits in parallel method would go from the last time facade ran, but this doesn't really get used. RIght now it gets the parent commits first, and gets the existing commits in the database already, and only goes through the commit hashes for commits that are "missing commits" =`parent commits - existing commits`

sgoggins: This is not intended to affect the main contributor analysis process. It is less about 'commits' per se.

sgoggins: Also note that the `start_date` is never used in the function called with it, which is another reason to simplify/remove it ...

PR #3268, opened 2025-09-11
set up postgres monitoring with grafana and prometheus

MoralCode:
**Description**
this adds some containers to the stack to enable better monitoring of the postgres DB

This is intended to be something similar to flower in that its mainly (for now) intended for development use.

Grafana in particular takes a hot sec to start up (probably about a minute) before its accessible via the web interface (it has db migrations to do apparrently - the logs for this should be hidden by the instruction to only show error logs)

**Usage:**

1. start up the augur stack
2. navigate to http://localhost:3000 in your browser
3. login with the credentials admin/password
4. navigate to http://localhost:3000/d/000000039/postgresql-database (or via the UI: Dashboards > services >PostgreSQL)

Observe: data

Screenshot_20250911_151351

The usecases for this is to generally provide more observability into what is going on with postgres. This may specifically help with debugging facade things, especially with respect to how many deadlock events are happening, how frequently postgres is checkpointing, how many idle processes/connections there are, etc.

I also hope that, in the future we can add some openTelemetry type reporting into augur so that augur itself can be monitored through a dashboard like this (maybe we can pipe [Celery data](https://github.com/danihodovic/celery-exporter) into this?)

**Signed commits**

### DIFF
--- a/monitoring/docker-compose.yml
+++ b/monitoring/docker-compose.yml
@@ -18,6 +18,8 @@ services:
     image: prom/prometheus
     volumes:
       - ./prometheus.yml:/etc/prometheus/prometheus.yml:z
+    extra_hosts:
+     - "host.containers.internal:host-gateway"
 
   postgres-exporter:
     image: quay.io/prometheuscommunity/postgres-exporter


MoralCode:
Reviewers: if you want i can comment out the new config in the dockerfile to match the convention we use for flower

docker-compose.yml
```diff
@@ -110,6 +110,35 @@ services:
...
+    ports:
+      - 9090:9090
+    volumes:
+      - ./prometheus.yml:/etc/prometheus/prometheus.yml:z
+
+  postgres-exporter:
+    image: quay.io/prometheuscommunity/postgres-exporter
+    ports:
+      - 9187:9187
+    env_file: .env
```
MoralCode: This line is breaking CI.

sgoggins (dismissed):
Excellent!

MoralCode:
current status (as a draft): postgres monitoring seems to work (just do `docker compose up` in the `monitoring` directory once augur is running (it assumes you use the compose file in the root of the repo).

Flower also (apparrently) ships with a /metrics endpoint that prometheus can ingest so im working on getting that hooked up.

feel free to play with this if anyone is interested!

MoralCode:
Seems to be working under both podman and docker environments

MoralCode:
This also includes both postgres and celery monitoring dashboards.

To use this you need to add a .env file in the monitoring folder containing augur db credentials (`AUGUR_DATABASE`, `AUGUR_HOST`, `AUGUR_DB_USER`, `AUGUR_DB_PASSWORD`), then you can just do a regular `<tool> compose up -d` in the monitoring dir and it will start the monitoring stack. 

this is intended to run independently of augur so you can take augur down and back up and the monitoring should keep going and simply reflect a gap in the data

MoralCode:
@sgoggins i think my force pushing dismissed your review

MoralCode:
Given this is entirely independent, im going to maintain this separately to avoid cluttering the augur repo with extra utilities.

Code will be maintained at https://github.com/oss-aspen/augur-monitoring

PR #3271, opened 2025-09-15
scheduling: refactor send_messages to de-duplicate code path

MoralCode:
**Description**
in looking for the facade scheduling issues, I noticed a couple structural (not functional) issues that arent directly going to solve the core problem but just help to clean up the code so its more readable when trying to debug the facade issues

1. This code path was almost entirely duplicated in two branches of an if-else chain.
2. the catch-all else block was being used for gitlab, potentially blocking any future additional sources

This essentially moves the actual "meat" of this function to below the if-else chain and extracts a couple new variables to abstract the differences between the two implementations (i.e. changing which list of phases gets used, and also printing the correct platform name in log messages)

**Signed commits**

### DIFF
--- a/augur/tasks/util/collection_util.py
+++ b/augur/tasks/util/collection_util.py
@@ -597,37 +597,33 @@ class AugurTaskRoutine:
             for repo_git, full_collection in col_hook.repo_list:
 
                 repo = get_repo_by_repo_git(repo_git)
+                platform_name = "github"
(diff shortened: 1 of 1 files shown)


MoralCode:
interestingly, this change was discovered when I asked gemini for some help tracking down the facade changes. It made a code change and, when describing the changes, said "The same logic is applied to GitLab phases." - prompting me to go see what the differences were between the two implementations before realizing they could be deduplicated

Issue #3272, opened 2025-09-15
Use Classes to enforce consistent APIs


MoralCode:
https://github.com/chaoss/augur/blob/2c9a5f1f2a9b58cfec035982fd0b27bd31cb0fbe/augur/tasks/start_tasks.py#L159

These builders are all intended to use the same APIs and conventions, which is why many of them refer to "enabled phases" when they dont need the ability to have disabled phases.  All collection request builders must be able to accept a list of 'enabled_phase_names', even if they are not used.

these should maybe be refactored into instances  or subclasses so that the convention of matching signatures is more explicit in its purpose and able to be technically enforced.

MoralCode:
same applies here https://github.com/chaoss/augur/blob/2c9a5f1f2a9b58cfec035982fd0b27bd31cb0fbe/augur/tasks/util/collection_util.py#L605

lists of tasks, such as can be seen in facade_phase (variable `facade_core_collection`)

Sukuna0007Abhi:
I will give it a try @MoralCode , pr #3583

PR #3273, opened 2025-09-15
facade: refactor facade_phase to return the group itself so that celery can properly convert things into a chord

MoralCode:
**Description**
While I haven't been certain that I can see the facade recollection issue locally, this change allowed me to successfully find `facade_task_success_util` via a flower task search in my local instance, meaning the tasks that were not running are now running again,

The thing that helped me get through this issue was this analysis from GPT5:
> I traced the facade orchestration. The issue is how the facade phase is chained: your facade_phase returns a chain that ends with a group, and then the outer routine chains the “success” util after that. In Celery, when you want to run something after a group completes, you must form a chord (e.g., group(...) | next_task). A nested chain that returns only a group won’t implicitly convert to a chord, so the subsequent facade_task_success_util never fires.

**Notes for Reviewers**
In doing this refactor, I didnt see any other uses of `facade_phase` besides in `build_facade_repo_collect_request`, so it doesnt seem like the specific return type here is being super heavily relied on (although maybe thats due to an unwritten convention as noted in #3272)

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -491,7 +491,6 @@ def facade_phase(repo_git, full_collection):
     #force_analysis = session.force_analysis
     run_facade_contributors = facade_helper.run_facade_contributors
 
-    facade_sequence = []
     facade_core_collection = []
 
     if not limited_run or (limited_run and pull_repos):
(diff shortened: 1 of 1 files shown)


augur/tasks/git/facade_tasks.py
```diff
@@ -509,14 +508,14 @@ def facade_phase(repo_git, full_collection):
...
             process_libyear_dependency_metrics.si(repo_git),
             process_scc_value_metrics.si(repo_git)
         )
-    )
+    ]
 
     logger.info(f"Facade sequence: {facade_sequence}")
-    return chain(*facade_sequence)
\ No newline at end of file
+    return facade_sequence[0]
```
Ulincsys: Is there a particular reason to keep the list if we are just going to peek the first item from it on return?

MoralCode: i only did this to preserve the log statement. i guess if the print statement can handle the list of objects just fine, it should be fine to just throw the object at it.

MoralCode: thanks for calling me out on my addition of tech debt lol - this probably isnt necessary, i think i was just like "eh whatever it works" when i wrote it and forgot to revisit

MoralCode:
oh fun, i guess one of my rebases or amends must have stripped out the DCO...

PR #3280, opened 2025-09-20
Add type annotations to db/util.py and configured mypy

### LINKED ISSUE #3274: Add CI job to check and enforce python typing information to ensure quality
Type checking helps make the code base more maintainable, but given that there is a lot of existing code, it's less feasible to add types all at once.

- [ ] Choose a type checking tool
- [ ] configure the tool in such a manner that we can add types one file at a time, essentially ratcheting the checking over time. Once a file becomes typed, CI should ensure it stays that way.
- [ ] As an initial step, add types to at least one file to demonstrate feasibility of this approach. Over time, continue adding types

skools-here:
**Description**
This PR introduces type annotations to the augur\application\db\util.py file and adds mypy for static type checking.
This PR fixes #3274

Changes included:
1) Ran mypy locally on the project and fixed all type issues in this file and made mypy.ini so we can add more files in future for types checking.
2) Starting with smaller utility/helper files as suggested, focusing on the ORM/database access layer.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/db/util.py
+++ b/augur/application/db/util.py
@@ -29,31 +29,6 @@ def catch_operational_error(func: Callable[[], Any]) -> Any:
 
     raise Exception(error)
 
-
-def execute_session_query(query: Any, query_type: str = "all") -> Any:
-    """
-    Executes an SQLAlchemy ORM query with retries on OperationalError.
-    
-    Args:
-        query: SQLAlchemy ORM query object.
-        query_type: One of 'all', 'one', 'first'.
-    
-    Returns:
-        Query result according to query_type.
-    """
-    func: Optional[Callable[[], Any]] = None
-
-    if query_type == "all":
-        func = query.all
-    elif query_type == "one":
-        func = query.one
-    elif query_type == "first":
-        func = query.first
-    else:
-        raise Exception(f"ERROR: Unsupported query type '{query_type}'")
-
-    return catch_operational_error(func)
-
 def convert_orm_list_to_dict_list(result: List["Base"]) -> List[Dict[str, Any]]:
     """
     Converts a list of ORM model instances to a list of dictionaries.
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -21,7 +21,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .[dev] 
+          pip install .[dev]
+          pip install mypy
 
       - name: Run mypy
         run: |
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -4,7 +4,7 @@ import json
 from typing import List, Any, Optional
 import os
 from augur.application.db.models import Config 
-from augur.application.db.util import execute_session_query, convert_type_of_value
+from augur.application.db.helpers import execute_session_query, convert_type_of_value
 
 def get_development_flag_from_config():
     
--- a/augur/application/db/helpers.py
+++ b/augur/application/db/helpers.py
@@ -1,0 +1,26 @@
+from typing import Any, Callable, Optional
+from augur.application.db.util import catch_operational_error
+
+def execute_session_query(query: Any, query_type: str = "all") -> Any:
+    """
+    Executes an SQLAlchemy ORM query with retries on OperationalError.
+    
+    Args:
+        query: SQLAlchemy ORM query object.
+        query_type: One of 'all', 'one', 'first'.
+    
+    Returns:
+        Query result according to query_type.
+    """
+    func: Optional[Callable[[], Any]] = None
+
+    if query_type == "all":
+        func = query.all
+    elif query_type == "one":
+        func = query.one
+    elif query_type == "first":
+        func = query.first
+    else:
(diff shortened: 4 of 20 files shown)


augur/application/db/util.py
```diff
@@ -1,20 +1,20 @@
...
+    error: Optional[str] = None
     timeout = 240
 
     while attempts < 4:
-
-        # do the sleep here instead of instead of in the exception 
-        # so it doesn't sleep after the last failed time
         if attempts > 0:
-            #Do a 30% exponential backoff
+            # Do a 30% exponential backoff
```
MoralCode: this seems like a non-typing change that removed a helpful comment - was this removed on purpose?

augur/application/db/util.py
```diff
@@ -43,48 +53,47 @@ def execute_session_query(query, query_type="all"):
     return catch_operational_error(func)
 
 
-
-def convert_orm_list_to_dict_list(result):
-    new_list = []
+def convert_orm_list_to_dict_list(result: List[Any]) -> List[Dict[str, Any]]:
```
MoralCode: it might be nice to not use `Any` as the type here - does the ORM have a base/parent class that all the ORM classes inherit from? maybe thats a better thing to use?

augur/application/db/util.py
```diff
@@ -43,48 +53,47 @@ def execute_session_query(query, query_type="all"):
...
-    return new_list
 
+    return new_list
 
 
-def convert_type_of_value(config_dict, logger=None):
-        
-    data_type = config_dict["type"]
+def convert_type_of_value(
+    config_dict: Dict[str, Any], logger: Optional[Any] = None
```
MoralCode: does config_dict come from somewhere? could its type be specified more precisely than `Any`? same with  the logger

augur/application/db/util.py
```diff
@@ -43,48 +53,47 @@ def execute_session_query(query, query_type="all"):
...
+) -> Dict[str, Any]:
+    """
+    Converts the 'value' field in config_dict to the type specified in 'type'.
+    Supported types: str, int, bool, float.
+    """
+    data_type: Optional[str] = config_dict.get("type")
 
-    if data_type == "str" or data_type is None:
+    if data_type is None or data_type == "str":
```
MoralCode: why was the order of these changed? this could cause a crash if the data type is none because it is relying on the short curcuiting behavior of `or`

augur/application/db/util.py
```diff
@@ -43,48 +53,47 @@ def execute_session_query(query, query_type="all"):
...
 
     elif data_type == "bool":
-        value = config_dict["value"]
-        
-        if value.lower() == "false":
-            config_dict["value"] = False
-        else:
-            config_dict["value"] = True
+        value = str(config_dict["value"]).lower()
+        config_dict["value"] = value != "false"
```
MoralCode, line 66: if value is exactly `false` (string), this will evaluate to true.

MoralCode (changes requested):
seems like an okay first pass (i still need to test it locally)

Overall notes: 
- any changes to the actual logic are likely going to slow this down a fair bit as the logic needs testing to make sure it behaves the same
- also, try to avoid using `Any` as a type if you can at all avoid it - its not very helpful at enforcing typings since it accepts anything.

augur/application/db/util.py
skools-here: I’ll update convert_orm_list_to_dict_list to use the ORM base class, and refine config_dict + logger to use more specific types (Optional[logging.Logger], etc.) .

augur/application/db/util.py
skools-here: Yes i will fix this this .

skools-here:
Yes thanks for the review will be changing all the logic changes which may cause problem and will also try to reduce the use of any .

augur/application/db/util.py
MoralCode: wait did i get this wrong?

skools-here: the earlier function (Not type safe one) already had that anything not false is True behavior. So according to me this must work i am not sure though .

MoralCode:
we should also probably add a CI job to check the types so its clear if something breaks on a pull request

pyproject.toml
```diff
@@ -99,6 +99,7 @@ dev = [
     "pytest==6.2.5",
     "toml>=0.10.2",
     "ipdb==0.13.9",
+    "mypy>=1.8.0",
```
MoralCode, line 102: if you change this file, you also need to run `uv sync` and commit the changes to the `uv.lock` file. this will likely help some of the CI jobs pass

MoralCode:
@skools-here do you have any updates on this?

no worries if not - just wanted to check if you were still interested in merging this. If not, I can potentially find another maintainer can take over and make the necessary adjustments to get this merged

Part of me also wonders what would happen if mypy is run on this repo without any explicit typing added? :thinking:

skools-here:
@MoralCode After revisiting the , I realized this might be a bit more complex than I initially expected, especially around how mypy interacts with the existing code. I don’t think I’ll be able to give it the attention it needs right now, so feel free to reassign it or have someone else take it over.

MoralCode:
Superseeded by #3315

PR #3284, opened 2025-09-23
Load test : Ready for Release

sgoggins:
**Description**

This PR : 
1. Comments out an API endpoint that has been getting abused in production, and is no longer essential. 
2. Reverts a change to `augur/tasks/util/collection_util.py` that, for reasons that defy logic, disable secondary and facade tasks on some deployment environments. We have yet to determine WHY this causes issue, but know that it does so are reverting the change for release.


MoralCode:
I made #3289  so we can ship the part of this that we agree on and keep the discussion focused on figuring out the issues with the other piece

I think this PR should be closed in favor of just merging its constituent parts directly into main

PR #3289, opened 2025-09-23
commented out abused API endpoint

MoralCode:
**Description**
This takes the non-contested part of #3284 and merges it into main because it mitigates an issue with this endpoint being abused

This fix was done entirely by sean, I'm just making this PR so this change can land in main to simplify the review process for everything else (I strongly disagree that we should be proposing PRs containing merge commits from pull requests)

### DIFF
--- a/augur/api/routes/contributor_reports.py
+++ b/augur/api/routes/contributor_reports.py
@@ -305,8 +305,8 @@ def new_contributor_data_collection(repo_id, required_contributions):
     #     select count(*) from augur_data.repo; 
     #     """)
 
-    with current_app.engine.connect() as conn:
-        df = pd.read_sql(contributor_query2,  conn)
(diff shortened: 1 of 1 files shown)


augur/api/routes/contributor_reports.py
```diff
@@ -39,283 +39,295 @@ def new_contributor_data_collection(repo_id, required_contributions):
...
+
+    #     """)
+    # contributor_query2 = ("""
+
+    #     select count(*) from augur_data.repo; 
+    #     """)
 
     with current_app.engine.connect() as conn:
-        df = pd.read_sql(contributor_query,  conn)
+        df = pd.read_sql(contributor_query2,  conn)
```
MoralCode: where is `contributor_query2` defined?

PR #3297, opened 2025-10-01
refactor send_messages to remove a almost entirely duplicate code path

MoralCode:
**Description**

This change is identical to that proposed in #3271, except it contains a fix for the problem where other jobs were not being scheduled and thus had to be reverted (in #3292)

In reviewing the code, there was a spot in there that was intended to exit early and skip that iteration. I had used `return` (which exits the whole function) in the refactor rather than `continue` (which skips one loop iteration) because I was focused on deduplicating the if statement and didnt see that it was in a loop.

This is a refactor and doesnt fix an explicit issue, but helps generally reduce tech debt

**Notes for Reviewers**
When testing, ensure that all jobs are getting scheduled correctly.

**Signed commits**


MoralCode:
This doesnt really fix a bug so much as it improves the quality of the code (yes theres the detail that the original version of this introduced a bug but that was already reverted so for all intents and purposes this is a new refactoring change)

PR #3299, opened 2025-10-06
Fix examples of repos CSV files

giordano:
The [code which parses the repos files](https://github.com/chaoss/augur/blob/b0bb3b80402ee5fcd84bec7334e58a41f9f5ec8a/augur/application/cli/db.py#L61-L69) interprets the first column as URL and the second one as repo group ID, but all the examples were doing the opposite.

**Signed commits**


MoralCode:
This has two approvals so im going to merge it

PR #3300, opened 2025-10-06
docker compose: Make it crystal clear that all the GitHub/GitLab user/keys must be set

giordano:
The non-interactive Docker compose workflow requires all the variables `AUGUR_GIT{LA,HU}B_{USERNAME,API_KEY}` to be set even if they are going to be unused, otherwise you get stuck waiting for [the prompt](https://github.com/chaoss/augur/blob/b0bb3b80402ee5fcd84bec7334e58a41f9f5ec8a/scripts/install/config.sh#L18-L29)
```
You entered a blank line, are you sure?
```

**Signed commits**

### DIFF
--- a/docs/source/docker/docker-compose.rst
+++ b/docs/source/docker/docker-compose.rst
@@ -26,7 +26,7 @@ This section of the documentation details how to use Augur's Docker Compose conf
 
 .. warning::
 
-    Don't forget to provide your external database credentials in a file called ``.env`` file. Make sure the following environment variables are specified.
+    Don't forget to provide your external database credentials in a file called ``.env`` file. Make sure all the following environment variables are specified, keep placeholder values if you don't need some of them.
     Don't specify AUGUR_DB if you want the docker database to be used.
 
     Example .env:
--- a/docs/source/getting-started/using-docker.rst
+++ b/docs/source/getting-started/using-docker.rst
@@ -10,7 +10,7 @@ the following resources (or more).
 1. Clone the Augur repository https://github.com/chaoss/augur
 
 
-2. Create a .env file in the top level directory with the following fields:
+2. Create a ``.env`` file in the top level directory with the following fields (don't remove any variable, keep placeholder values if you don't need some of them):
 
 .. code:: python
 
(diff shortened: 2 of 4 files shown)


sgoggins (approved):
LGTM.

docs/source/getting-started/using-docker.rst
```diff
@@ -35,7 +35,7 @@ or
 
     podman compose up --build
 
-And augur should be up and running! Over time, you may decide that you want to download and run newer releases of Augur. It is critical that your `.env` file remains configured to use the same database name and password; though you can change the password if you understand how to connect to a database running inside a Docker container on your computer.
+And augur should be up and running! Over time, you may decide that you want to download and run newer releases of Augur. It is critical that your ``.env`` file remains configured to use the same database name and password; though you can change the password if you understand how to connect to a database running inside a Docker container on your computer.
```
MoralCode, line 38: the extra layer of backticks shouldnt be needed AFAIK

docs/source/docker/docker-compose.rst
```diff
@@ -26,7 +26,7 @@ This section of the documentation details how to use Augur's Docker Compose conf
 
 .. warning::
 
-    Don't forget to provide your external database credentials in a file called ``.env`` file. Make sure the following environment variables are specified.
+    Don't forget to provide your external database credentials in a file called ``.env`` file. Make sure all the following environment variables are specified, keep placeholder values if you don't need some of them.
```
MoralCode, line 29: Would it be possible to explain why the placeholders should be kept? (i.e. it will cause things not to start up if one is left unspecified?)

docs/source/getting-started/using-docker.rst
giordano: It is: `` `.env` `` is rendered as italic, ``` ``.env`` ``` is rendered as monospace.  See the RST docs: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

MoralCode: oh right, i forgot this was RST and not markdown

Issue #3308, opened 2025-10-14
megatask: Audit and unify logging throughout augur


MoralCode:
While tracking down https://github.com/chaoss/augur/issues/3306, I noticed that the logs (in facade) are both:
1. using a custom logging wrapper function
2. being written to a database table

while an unconventional practice, writing to the database can be done using existing output features of the logging package in python, allowing logging in facade to use standard calls like `logger.warn(message)` rather than custom functions like this.

Also there are many places throughout the code that are using print() statements rather than the python logging library

This suggests we should probably audit all the logging calls in augur to see if we can simplify everything under one system (and also potentially ask why we need logs in the database).

MoralCode:
This should also include a review of the logging docs to ensure they are accurate

Specifically this part of the docs are old: https://github.com/chaoss/augur/blob/3fb2bae9d779561a77715cb3af9f780725ba3e10/docs/source/development-guide/logging.rst?plain=1#L5

sgoggins:
@MoralCode : The logs written to the database are from the original Facade program written by @brianwarner ... https://github.com/brianwarner/facade .. Is the original source code we massively adapted for Augur ... 

So, I think they can go, no problem.

MoralCode:
See also https://github.com/chaoss/augur/issues/2848

MoralCode:
tagging as good first issue for newcomers to tackle parts of the child issues in small steps. none of these issues should be closed with a single large PR.

Im going to lock the comments on this parent issue to encourage anyone new to discuss things in the child issues and/or on slack (link available on the getting started page of https://chaoss.community)

PR #3309, opened 2025-10-14
Log facade messages based on the type requested

### LINKED ISSUE #3307: Print facade logs at the level the user specifies
While tracking down #3306, it was noted that this log message should be a WARNING, not INFO:

https://github.com/chaoss/augur/blob/7f58fbc8075d228c9d427f86e577af31e4a49520/augur/tasks/git/util/facade_worker/facade_worker/repofetch.py#L381

While attempting to quickly fix this, I discovered that Logs are all being output as INFO level logs, rather than the level the user selected

https://github.com/chaoss/augur/blob/7f58fbc8075d228c9d427f86e577af31e4a49520/augur/tasks/git/util/facade_worker/facade_worker/config.py#L170

MoralCode:
**Description**

This PR fixes #3307 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/git/util/facade_worker/facade_worker/config.py
+++ b/augur/tasks/git/util/facade_worker/facade_worker/config.py
@@ -167,7 +167,13 @@ class FacadeHelper():
         # Log an activity based upon urgency and user's preference.  If the log level is
         # "Debug", then just print it and don't save it in the database.
         log_options = ('Error','Quiet','Info','Verbose','Debug')
-        self.logger.info(f"* {status}\n")
+        logmsg = f"* {status}\n"
+        if level == "Error":
+            self.logger.error(logmsg)
+        elif level == "Debug" or level == "Verbose":
+            self.logger.debug(logmsg)
+        else:
+            self.logger.info(logmsg)
 
         #Return if only debug 
         if level == 'Debug':


augur/tasks/git/util/facade_worker/facade_worker/config.py
```diff
@@ -167,7 +167,13 @@ def log_activity(self, level, status):
         # Log an activity based upon urgency and user's preference.  If the log level is
         # "Debug", then just print it and don't save it in the database.
         log_options = ('Error','Quiet','Info','Verbose','Debug')
-        self.logger.info(f"* {status}\n")
+        logmsg = f"* {status}\n"
+        if level == "Error":
```
sgoggins, line 171: The message level is usually set in all caps like , `DEBUG` or `ERROR` .. does this logging checker do anything to account for differences in case? I think it needs to ...

MoralCode: that is added/managed by the `logging` package (i.e. when `self.logger.<whatevr>()` is called)

All we need to worry about here is the `level` parameter being passed to this function. All the existing calls to this logging helper that I see use Camel case and the existing check for whether Debug mode is requested does not account for casing, so i think its fine.

I will also be likely entirely rewriting all this logic if/when i fix #3308

PR #3312, opened 2025-10-15
Fix `git reset` command to include remote default branch

giordano:
**Description**

`git reset --hard <remote>` isn't a correct command, the argument of `git reset` must be a tree-ish, which a remote name alone isn't.

In some cases this command fails with
```
fatal: ambiguous argument 'origin': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

**Notes for Reviewers**

**Signed commits**


sgoggins (approved):
This is a GREAT catch! ... @MoralCode : I think this could be one small thing that fixes a few repos that appear stuck; Though my testing is showing mostly the large repos just take large time.

Issue #3302, opened 2025-10-07
Store config files in the same place


MoralCode:
I notice there are many config files referenced thoughout this repo.

- `db.config.json` in `__init__`, db, and engine within the cli module
- `augur.config.json` in the CLI
- `config.yml` in app/view/init, for which `CONFIG_LOCATION=/config/config.yml` in the dockerfile is defined

Can we standardize the concept of a config directory so all these files can be in one place and we can start reorganizing how config is done?

skools-here:
I think this issue can be closed or there is some work left in this ?

MoralCode:
its probably worth going through and auditing the code that accesses the other config files and adjust them to use the environment variable that was introduced in #3314. then we can eliminate the similar variable (that is present in the dockerfile and also does config) and generally clean up the handling of config files.

skools-here:
Hey i can update all locations that read configuration files (db.config.json, config.yml, augur.config.json) to use the CONFIG_DATADIR environment variable introduced in #3314 . Can also removes legacy variables like CONFIG_LOCATION .

Should i work on this ?

MoralCode:
that would be great!

PR #3314, opened 2025-10-15
Allow default config to be read from a file

MoralCode:
**Description**
The default config is currently hardcoded. When sysadmins, (especially those using docker where the db config is overwritten on startup) want to change something, it requires augur's container images to be rebuilt since the code has changed.

This change checks for the presence of a json file reads from that if present, instead of the hardcoded config.  

Per #3302 this PR also introduces a new environment variable containing a path to a config directory, which, if specified, is used as the base directory when looking for this augur config file.

**Notes for Reviewers**
Draft PR because I have yet to test this locally - but will do soon

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -123,7 +123,8 @@ class AugurConfig():
         self.logger = logger
 
         self.accepted_types = ["str", "bool", "int", "float", "NoneType"]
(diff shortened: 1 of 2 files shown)


MoralCode:
i havent personally run this

PR #3315, opened 2025-10-15
Alternative mypy setup

MoralCode:
**Description**
This sets up mypy in a slightly different way to #3280.

to run the type checker, run `uv run mypy`

This PR sets up for an eventual fix to #3274 

**Notes for Reviewers**
mostly just making this for reference and to see if it causes the CI jobs to freak out

if this ends up being used, i should probably go back and squash some of the fixup commits i made

**Signed commits**

### DIFF
--- a/augur/application/db/data_parse.py
+++ b/augur/application/db/data_parse.py
@@ -457,7 +457,7 @@ def extract_needed_gitlab_issue_label_data(labels: List[dict], repo_id: int, too
 
 
 
-def extract_needed_issue_message_ref_data(message: dict, issue_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> List[dict]:
+def extract_needed_issue_message_ref_data(message: dict, issue_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> dict:
     """
     Retrieve only the needed data for pr labels from the api response
 
@@ -487,7 +487,7 @@ def extract_needed_issue_message_ref_data(message: dict, issue_id: int, repo_id:
     return message_ref_dict
 
 # retrieve only the needed data for pr labels from the api response
-def extract_needed_pr_message_ref_data(comment: dict, pull_request_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> List[dict]:
+def extract_needed_pr_message_ref_data(comment: dict, pull_request_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> dict:
 
     message_ref_dict = {
             'pull_request_id': pull_request_id,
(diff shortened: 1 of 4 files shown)


MoralCode:
@skools-here what do you think of this? I made this branch mostly just to see what would happen if I ran mypy on augur out of the box (answer: a ton of type errors i didnt want to deal with)

so I ended up narrowing it to the db files and only had like 6 things to fix. It turns out mypy generally won't complain if things are untyped, which seems helpful for migrating.

The only thing left IMO is to figure out how to get the CI (i feel like one of the existing CI jobs should be able to handle it?) to run the equivalent of `uv run mypy` to ensure that things in this set of files dont regress.

skools-here:
@MoralCode  This looks great narrowing Mypy’s scope to the database files is a very practical approach.

Using uv run mypy fits nicely into our existing workflow. The main thing now is ensuring this check runs automatically in CI to prevent type regressions in future changes.

I think all that’s needed is to add a step in .github/workflows/checks.yml to run Mypy. Once in place, this will work seamlessly, and we can gradually expand the set of type-checked files over time.

MoralCode:
> I think all that’s needed is to add a step in .github/workflows/checks.yml to run Mypy. Once in place, this will work seamlessly, and we can gradually expand the set of type-checked files over time.

Is this something you want to do/think you would be able to do?

skools-here:
> Is this something you want to do/think you would be able to do?

I think I can try to work on this workflow addition. Would love to get your help in between though.

MoralCode:
Sounds good!
Do you want to try and build on this PR? Or should @sgoggins merge it and we can add the CI jobs in a new PR?

skools-here:
I’m happy to build on this PR so everything lands together. 

But if you’d prefer to merge this first and handle the CI setup separately, that works for me too — totally your call!

sgoggins:
> **Description** This sets up mypy in a slightly different way to #3280.
> 
> to run the type checker, run `uv run mypy`
> 
> This PR sets up for an eventual fix to #3274
> 
> **Notes for Reviewers** mostly just making this for reference and to see if it causes the CI jobs to freak out
> 
> if this ends up being used, i should probably go back and squash some of the fixup commits i made
> 
> **Signed commits**
> 
>     * [x]  Yes, I signed my commits.

You only made 5 commits. I don't think we need to squash those ... but I am kind of a pack rat.

augur/application/db/data_parse.py
```diff
@@ -457,7 +457,7 @@ def extract_needed_gitlab_issue_label_data(labels: List[dict], repo_id: int, too
 
 
 
-def extract_needed_issue_message_ref_data(message: dict, issue_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> List[dict]:
+def extract_needed_issue_message_ref_data(message: dict, issue_id: int, repo_id: int, tool_source: str, tool_version: str, data_source: str) -> dict:
```
sgoggins, line 460: Did you run Augur with these changes? On the face of it I am mildly curious if changing from `List[dict]` to a `dict` will alter expected return values for other parts of Augur.

MoralCode: Python typings are just hints, they aren't actually going to change the properties of the data thats returned or cause errors by themselves (unless you like make a syntax error or try and use a type thats not present or something)

MoralCode:
I can do a quick check tomorrow to make sure augur still boots up and appears to work. All the changes here should be logically identical to how the code was beforehand. This is just setting up to have a type checker be able to validate that the declared types match with how the python objects/data are actually being used

PR #3330, opened 2025-10-22
Fixing the API key Visibility issue in debug mode

### LINKED ISSUE #3328: Stop printing API keys to the log when in debug mode
https://github.com/chaoss/augur/blob/a632b048606b2b85dc540e47964c230e03051775/augur/tasks/util/random_key_auth.py#L36

This line prints a full API key to the log when in debug mode.

if we are going to be printing secrets in plain text, can we at LEAST pass it through a function to obfuscate the key somehow? (i.e. remove all but the first 6 and last 3 characters of the key, and replace the middle with 6 or so asterisks `******`, or even better, find a really small one-way cryptographic hash algorithm to pass the key though)

Separate, but related: The only place I think we should be printing API keys is in the output of the `augur github api-keys` CLI and maybe the web UI (and even then it should be censored by default without an EXPLICIT opt-in CLI arg or button click)

PredictiveManish:
**Description**
This PR Secures the API Key logging in ```RandomKeyAuth```, improves the security by ensuring the keys never logged in plaintext. Specifically:
1. Introduces `mask_key()`
 which obfuscate API keys in logs, by only showing first 6 and last 3 char and in between there're asterisks.
2. Added `key_fingerprint()` to log a hash which is short and non-reversible
3. Updated debug logging to use the masked value and fingerprint instead of the raw key.
4. Error logging updated to avoid printing the full list of keys when none are available.

This PR fixes #3328 

**Signed commits**

### DIFF
--- a/augur/tasks/util/random_key_auth.py
+++ b/augur/tasks/util/random_key_auth.py
@@ -7,12 +7,6 @@ import hashlib
 from augur.util.keys import mask_key
 
 
-def key_fingerprint(key: str, length: int = 12) -> str:
-    """Return a short non-reversible fingerprint of the key for correlation."""
-    h = hashlib.sha256(key.encode("utf-8")).hexdigest()
-    return h[:length]
-
-
 class RandomKeyAuth(Auth):
     """Custom Auth class for httpx that randomly assigns an api key to each request.
 
@@ -39,8 +33,7 @@ class RandomKeyAuth(Auth):
 
         # Log only masked or hashed form, never the full key
         masked = mask_key(key_value)
-        fingerprint = key_fingerprint(key_value)
-        self.logger.debug(f"Key used for request (masked): {masked} | fingerprint: {fingerprint}")
+        self.logger.debug(f"Key used for request (masked): {masked}")
 
         # Apply formatting if needed
         if self.key_format:


augur/tasks/util/random_key_auth.py
```diff
@@ -3,51 +3,59 @@
...
+        header_name (str): name of header to set the key into
+        key_format (str): optional format string with {0} placeholder for key
     """
-    
-    # pass a list of keys that are strings
-    # pass the name of the header that you would like to be set on the request
-    # Optionally pass the key_format. This is a string that contains a {} so the key can be added and applied to the header in the correct way.
-    # For example on github the keys are formatted like "token asdfasfdasf" where asdfasfdasf is the key. So for github 
-    # the key_format="token {0}"
+
```
MoralCode, line 18: not a fan of the extra/unrelated changes, but i can see how in some cases it marginally improves things (yet in others some detail is lost).

I'll see what everyone else says, but Im inclined to let this stay

augur/tasks/util/random_key_auth.py
```diff
@@ -3,51 +3,59 @@
...
+        key_value = choice(self.list_of_keys)
+
+        # Log only masked or hashed form, never the full key
+        masked = mask_key(key_value)
+        fingerprint = key_fingerprint(key_value)
+        self.logger.debug(f"Key used for request (masked): {masked} | fingerprint: {fingerprint}")
+
+        # Apply formatting if needed
+        if self.key_format:
+            key_string = self.key_format.format(key_value)
```
MoralCode, line 40: Was a little worried at the size of this diff at first, but i see you unwrapped the code from within an if and turned it into more of an early-fail/gate type pattern. Looks good!

augur/tasks/util/random_key_auth.py
```diff
@@ -3,51 +3,59 @@
...
+    """Mask key except for the first and last few characters."""
+    if not isinstance(key, str) or len(key) <= (first + last):
+        return "*" * stars
+    return f"{key[:first]}{'*' * stars}{key[-last:]}"
+
+
+def key_fingerprint(key: str, length: int = 12) -> str:
+    """Return a short non-reversible fingerprint of the key for correlation."""
+    h = hashlib.sha256(key.encode("utf-8")).hexdigest()
+    return h[:length]
```
MoralCode: i think in the original issue I was intending to propose *either* blanking out the middle or hashing. I think blanking out the middle is better overall since current users are probably more used to seeing their keys in this form (ie.. when they input them or use the CLI to check key expiry). I'll wait to see what the other maintainers think, but I suspect the value of the hashing isnt yet worth the additional dependency

augur/tasks/util/random_key_auth.py
```diff
@@ -3,51 +3,59 @@
...
 from random import choice
+import hashlib
 
 
+def mask_key(key: str, first: int = 6, last: int = 3, stars: int = 6) -> str:
+    """Mask key except for the first and last few characters."""
+    if not isinstance(key, str) or len(key) <= (first + last):
+        return "*" * stars
+    return f"{key[:first]}{'*' * stars}{key[-last:]}"
+
```
MoralCode: I suspect this function will be helpful in other parts of the codebase too.

Maintainers: where would be the best place to move this such that it could be used by the augur CLI too? im thinking the `augur.util` package maybe?

augur/tasks/util/random_key_auth.py
PredictiveManish: Thanks for your review and time, waiting for other reviews then will proceed further.

augur/tasks/util/random_key_auth.py
PredictiveManish: Thanks.

augur/tasks/util/random_key_auth.py
PredictiveManish: Yes I also think that it can be moved to somewhere else for making it helpful for all required place, just let me know where to move I will do it and import in this file!

augur/tasks/util/random_key_auth.py
PredictiveManish: That’s a fair point. I thought that having a short hash helps correlate which key was used without exposing any part of it (useful in environments where masking might still reveal too much).
But I agree the masking approach is simpler and more familiar.
Happy to revert to that if maintainers prefer consistency with existing CLI behavior.

augur/tasks/util/random_key_auth.py
```diff
@@ -3,51 +3,59 @@
 
 from httpx import Auth, Request, Response
 from random import choice
+import hashlib
```
sgoggins, line 6: does hashlib require a change to pyproject.toml? I think it likely does because I do not recall (its been 9 years so that recall isn't perfect) using hashlib previously on Augur.

augur/tasks/util/random_key_auth.py
sgoggins: I think all the key LOGGING occurs through this file; We have caused ourself confusion in the past by creating general function libraries where localized ones were adequate. I'm in favor of generalizability. I just cannot think of a specific case where we would need it ... CLI is a good point @MoralCode ... `augur.util` is a nice centralized place and one of the first we look at if the function isn't local. ... On @MoralCode 's recommendation/suggestion, that might make sense. Then we just import `augur.util` (look through the codebase for how we do that) into this file. That seems smart!

augur/tasks/util/random_key_auth.py
sgoggins: @MoralCode makes a really good point here. I like hashing from a security perspective ... if we leave the first 4 and last 4 in place, and hash the middle part it is WAY more secure when logging, and also allows us to debug key issues as they related to particular keys. 

Specifically, keys go bad for one reason or another. If we preserve no part of the key it is more secure, but then quite difficult to identify the bad keys. 

One possible compromise (and replacing the middle seems like a good one already) would be to include the `id` of the key from the `augur_operations.worker_oauth` table. Such as: 

```bash
oauth_id	name	consumer_key	consumer_secret	access_token	access_token_secret	repo_directory	platform
20004	x@x.com	0	0	ghp_NOTREAL	0		github
21000	doge@x.com	0	0	ghp_NOTREAL	0		github
20005	lessla@x.com	0	0	ghp_NOTREAL 0		github
```

so ... 20004, 21000, and 20005 would be included in the log ... 

But that "feels" like more work than removing the middle. I'm ok with either.

augur/tasks/util/random_key_auth.py
sgoggins: @MoralCode : I tend to agree that this change isn't necessary, and every change carries some small degree of risk.

augur/tasks/util/random_key_auth.py
sgoggins: LGTM

sgoggins (changes requested):
I made a few suggestions that may require some small changes. The pyproject.toml question is probably the most significant question.

augur/tasks/util/random_key_auth.py
MoralCode: https://docs.python.org/3/library/hashlib.html

I think its part of the python standard library

augur/tasks/util/random_key_auth.py
PredictiveManish: Reverted changes!

augur/tasks/util/random_key_auth.py
PredictiveManish: Done as you suggested sir!

augur/tasks/util/random_key_auth.py
PredictiveManish: removed `key_fingerprint()` as of now, if it requires will add it later as per advancement of repo!

PR #3332, opened 2025-10-23
Fix variables issue in #3183

PredictiveManish:
**Description**
This PR resolves the warnings given in changes in #3183. Removes the unused import `sphinx_rtd_theme` and import variables `__copyright__, __release__, __version__` correctly from `metadata.py` into `conf.py` as suggested by reviewbot.

This PR fixes small import issues.

**Signed commits**

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -19,7 +19,7 @@
 #
 import os
 import sys
-
+import sphinx_rtd_theme
 
 here = os.path.abspath(os.path.dirname(__file__))
 


docs/source/conf.py
```diff
@@ -19,13 +19,15 @@
 #
 import os
 import sys
-import sphinx_rtd_theme
+
```
MoralCode: this seems based on the name like it may be important - how do you know this isnt an issue with the linter incorrectly identifying an unused package that may actually be used?

MoralCode (commented):
Might also be worth looping in the author of the prior PR that this update is for @JohnStrunk.

Overall i dont really have much opinion on this other than these diffs seem reasonably small - will probably defer explicit approval to other maintainers who have more experience to better evaluate whether these changes are helpful

docs/source/conf.py
JohnStrunk: This is the read-the-docs theme. It's [referenced below](https://github.com/chaoss/augur/pull/3332/files#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698L119), though not in a form that will be picked up by a linter.
The preview seems to build, but I wonder whether local builds build correctly and have the proper theme applied.

PredictiveManish:
> Might also be worth looping in the author of the prior PR that this update is for @JohnStrunk.
> 
> Overall i dont really have much opinion on this other than these diffs seem reasonably small - will probably defer explicit approval to other maintainers who have more experience to better evaluate whether these changes are helpful

Yes, this issue is not that much big just a small warning! Main issue is the other part of unable to importing properly! it's just a solution for the warning!

Issue #3249, opened 2025-08-12
Update docs for running in production


MoralCode:
This is an issue to track changes that need to be made to the "how to run augur in production" [section](https://oss-augur.readthedocs.io/en/main/deployment/toc.html) of the docs

Changes to potentially add/mention/include (unless specified by a maintainer, do one of these per pull request):
- https://github.com/chaoss/augur/pull/3221
- https://github.com/chaoss/augur/issues/3231#issuecomment-3179446607
- Mention https://github.com/oss-aspen/infra-ansible/ ?
- mention https://github.com/chaoss/augur-utilities/ ?
- document all the environment variables in use (maybe in the README?)
- mention that the shm-size set in docker  should be increased and should correlate with whats set in the postgres config for shared memory size
- https://github.com/chaoss/augur/pull/3337 the worker counts need updating to values that better match the system that augur is installed on
- recommendations to tune postgres based on your server

niteshver:
Is this issue open??

MoralCode:
yeah i think it is still open - I just briefly checked the docs and it seems like they still need updating to mention many of the things above, and potentially also including testing to make sure the current instructions (which seem to use the makefile method) still work.

MoralCode:
i think @raisuraj2004 had a decent start on this in #3260  as well (although i havent fully reviewed it). maybe you can both collaborate on this for hacktoberfest?

AumOzaa:
hey!

I'd like to take up this issue, it'll be my first contribution.
I plan to add a new how_to_run_augur_in_production.md file in the docs/ folder and raise a PR documenting the AUGUR_RESET_LOGS environment variable.

MoralCode:
You're welcome to take this on! Make sure you take a look at the previous linked attempts at this to help you (both as an outline of what you may want to do, and also to see if you would do anything differently).

Feel free to reach out in the #wg-augur-8knot channel on slack (see [here](https://chaoss.community/kb-getting-started/) for join link) if you want to chat or run into issues!

AumOzaa:
hey,

Created a PR for the `how_to_run_augur_in_production.md` file in the `docs/` folder: #3311.  
Please let me know if any changes are required.

PR #3337, opened 2025-10-23
replace worker vmem config item with individual process counts

MoralCode:
**Description**
This pr removes the `worker_process_vmem_cap` config item and replaces it with a configurable count per worker:
- `core_worker_count`
- `secondary_worker_count`
- `facade_worker_count`

We default these values to 5 workers each because we anticipate most users getting started with docker will be likely running on laptops. We plan to document that these values should be changed in production as part of that issue ( #3249 )

This makes managing the tasks more intuitive for augur admins.

**Notes for Reviewers**
When running this, you will need to ensure that the database values are reset based on the hardcoded defaults in order for these values to exist in your config table. There is no migration for adding this and I do not plan to add one.

**Signed commits**


MoralCode:
given that this had to be updated in two places, I'm also curious if one of the functions can be entirely removed.


Also this probably still needs updating to adjust the place in the code where the tasks are actually added to the queues

sgoggins:
@MoralCode : ah .. looks like this one is breaking docker builds.

sgoggins:
@MoralCode : What are thinking re: docker build failure? Wait to merge until it passes, or solve it on another PR?

augur/application/cli/backend.py
```diff
@@ -87,7 +87,10 @@ def start(ctx, disable_collection, development, pidfile, port):
     if disable_collection:
         os.environ["AUGUR_DISABLE_COLLECTION"] = "1"
     
-    worker_vmem_cap = get_value("Celery", 'worker_process_vmem_cap')
+    core_worker_count = get_value("Celery", 'core_worker_count')
```
sgoggins, line 90: Do we need to add values to the database?

MoralCode: yes, but thats why i filed #3356 - the values and sensible defaults already exist in the config.py file, and I think we should have the application insert them if something tries to access them and it cannot find the value.  Then users dont need to run some manual command and we can keep the DB migrations limited to only structural fixes

MoralCode:
Works on my machine (and also with augur.json - although with some fiddling because podman permissions).

sgoggins:
Docker image build still failing, @MoralCode

MoralCode:
Rerunning. cant see an obvious error besides failing to clone chaoss/whitepaper.....

sgoggins (approved):
I think we have worked this one pretty close to the nub ... @MoralCode : Any reservations about merging?

MoralCode:
> I think we have worked this one pretty close to the nub ... @MoralCode : Any reservations about merging?

nope, good to merge!

PR #3339, opened 2025-10-24
remove unused functions for calculating recollection

MoralCode:
**Description**
These functions seem entirely unused

**Notes for Reviewers**

**Signed commits**


augur/tasks/util/collection_util.py
```diff
@@ -18,101 +18,6 @@
 from augur.tasks.util.collection_state import CollectionState
```
sgoggins, line 20: @ABrain7710 : I think this unused code is from our old job scheduler. We are not using it anymore, and I think we can get rid of this code now, but I'm a packrat and need another perspective.

Ulincsys (approved):

PR #3340, opened 2025-10-24
allow materialized view refresh to be disabled

### LINKED ISSUE #3331: Make materialize view refresh on/off
Turn materialized view refreshing into something that can easily be turned on or off

MoralCode:
**Description**
this change skips scheduling the materialized view refresh task if the config item for it (refresh_materialized_views_interval_in_days) is set to 0

This PR fixes #3331

**Notes for Reviewers**

**Signed commits**


sgoggins (approved):
LGTM, and kind of urgent for larger instances after talking with @cdolfi this week. 

There are some scripts the user can edit to do the rebuild manually, which is what I do. Basically every two weeks i am sure to restart augur, and while I am restarting, the last step BEFORE restarting is run those scripts. 

`scripts/mat_view_explore` ... we should probably provide some instructions as well. 

This is just FYI ... but, tl;dr on most systems we need to `export PGPASSWORD=whatevaYoPasswordIS` ... and edit connect info such as host and port.

PR #3346, opened 2025-10-28
Fix hardcoded collection intervals in multiple spots

### LINKED ISSUE #3338: Audit of how the collection dates are set
review on how this variable is set (I will populate this more in a bit, dont want to forget) 

https://github.com/chaoss/augur/blob/a632b048606b2b85dc540e47964c230e03051775/augur/tasks/start_tasks.py#L182

mohsinm-dev:
## Summary
- Moved hardcoded collection intervals from source code to configuration files
- Replaced hardcoded `days_until_collect_again = 1` parameter with configurable value
- Consolidated collection interval settings to eliminate code duplication across multiple spots

## Changes Made
- Updated `build_secondary_repo_collect_request` function to use configuration instead of hardcoded values
- Added collection interval settings to configuration files
- Removed unused hardcoded parameters as identified in the issue

## Test plan
- [ ] Verify configuration files are properly loaded
- [ ] Test that collection intervals respect new configuration values
- [ ] Confirm backward compatibility with existing deployments

Fixes #3338

### DIFF
--- a/docs/source/getting-started/collecting-data.rst
+++ b/docs/source/getting-started/collecting-data.rst
@@ -93,6 +93,9 @@ The celery monitor is responsible for generating the tasks that will tell the ot
 
 - ``refresh_materialized_views_interval_in_days``, number of days to wait between refreshes of materialized views.
 
+If you choose, you can also adjust the values in the ``Tasks`` block if you would like to control when tasks should be re-run on a given repository. 
+This is specified as a number of days since the last successful run.
+
 Adding repos for collection
 -----------------------------
 


augur/tasks/github/releases/core.py
```diff
@@ -23,7 +23,7 @@ def get_release_inf(repo_id, release, tag_only):
 
 
         release_inf = {
-            'release_id': release['id'],
+            'release_id': str(release['id']).strip(),
```
MoralCode: This change (and several others in this diff are not related to the issue being fixed here - is there a reason you made this change?

augur/tasks/start_tasks.py
```diff
@@ -193,13 +195,15 @@ def secondary_task_success_util_gen(repo_git, full_collection):
         return secondary_task_success_util.si(repo_git)
 
     secondary_enabled_phases.append(secondary_task_success_util_gen)
-    request = CollectionRequest("secondary",secondary_enabled_phases,max_repo=60, days_until_collect_again=10)
+    # Get collection intervals from config
+    intervals = get_collection_intervals_from_config()
+    request = CollectionRequest("secondary",secondary_enabled_phases,max_repo=60, days_until_collect_again=intervals['secondary'])
```
MoralCode, line 199: I disagree with doing things this way - i would rather keep the parameter in each function (while fixing the `CollectionRequest` instantiation to actually use it) and then read in the config values at the time each of these functions is called.

MoralCode: Doing it this way just means the function has more side-effects/prerequisites/things it expects that aren't clearly documented/visible in the function signature - it makes the functions less able to be refactored/reorganized as the codebase evolves without also moving around other things as well.

augur/tasks/util/collection_util.py
```diff
@@ -213,6 +213,25 @@ def get_enabled_phase_names_from_config():
...
+def get_collection_intervals_from_config():
+    """Get collection intervals from config with fallback to hardcoded defaults."""
+    try:
+        tasks_config = get_section("Tasks")
+        return {
+            'core': tasks_config.get('core_collection_interval', 15),
+            'secondary': tasks_config.get('secondary_collection_interval', 10), 
+            'facade': tasks_config.get('facade_collection_interval', 10),
+            'ml': tasks_config.get('ml_collection_interval', 40)
+        }
```
MoralCode: Im a little mixed on whether we need a dict for this overall since its one additional place that needs modifying if we add another worker for some reason.

Also this function seems to be injecting its own default values independently of the config classes that already exist. IMO the defaults should be handled in one place

MoralCode:
@mohsinm-dev Since this is fairly high priority for us, are you likely to have time to work on this PR, or would you be okay if one of the core maintainers takes it over and makes the fixes to make recollection intervals configurable?

augur/tasks/start_tasks.py
```diff
@@ -28,7 +28,7 @@
 from augur.tasks.util.collection_state import CollectionState
 from augur.tasks.util.collection_util import *
 from augur.tasks.git.util.facade_worker.facade_worker.utilitymethods import get_facade_weight_time_factor
-from augur.application.db.lib import execute_sql, get_session
+from augur.application.db.lib import execute_sql, get_session, get_section
```
MoralCode: this is importing from a different place that is separate from `application/config.py` and shouldnt be used

augur/tasks/start_tasks.py
```diff
@@ -175,7 +175,10 @@ def core_task_success_util_gen(repo_git, full_collection):
     primary_enabled_phases.append(core_task_success_util_gen)
     primary_gitlab_enabled_phases.append(core_task_success_util_gen)
 
-    primary_request = CollectionRequest("core",primary_enabled_phases,max_repo=40, days_until_collect_again=15, gitlab_phases=primary_gitlab_enabled_phases)
+    # Get core collection interval from config
+    tasks_config = get_section("Tasks")
+    core_interval = tasks_config.get('core_collection_interval', 15)
+    primary_request = CollectionRequest("core",primary_enabled_phases,max_repo=40, days_until_collect_again=core_interval, gitlab_phases=primary_gitlab_enabled_phases)
```
MoralCode: the days_until_collect_again parameter is unused. I think it should use the parameter and the config value should be injected at the point where build_primary_repo_collect_request is called (same for all these other similar functions)

MoralCode:
@mohsinm-dev looks good! I added a couple small changes (one to rename the variables to include the word `days` so its clearer what the values should be, and another one to add a brief mention of these changes to the ReadTheDocs). I also rebased these changes on top of the current main branch.


@sgoggins Can you set DCO to passing on this one?

augur/tasks/start_tasks.py
sgoggins: @MoralCode : This seems like a good time to fix an antipattern if you see win. Should we ask @mohsinm-dev to make these changes? I see the PR passed and sort of presume that this is a non-issue for you wrt merging .. ?

augur/tasks/start_tasks.py
```diff
@@ -215,7 +217,7 @@ def facade_task_update_weight_util_gen(repo_git, full_collection):
 
     facade_enabled_phases.append(facade_task_update_weight_util_gen)
 
-    request = CollectionRequest("facade",facade_enabled_phases,max_repo=30, days_until_collect_again=10)
+    request = CollectionRequest("facade",facade_enabled_phases,max_repo=30, days_until_collect_again=days_until_collect_again)
```
sgoggins, line 220: same pattern @MoralCode noted above .. down a few more for each of the task sections.

sgoggins (approved):
I'm going to say LGTM.

augur/tasks/start_tasks.py
MoralCode: im not sure what you are referring to here?

PR #3353, opened 2025-10-29
Add Toggle Option for Facade Commit Messages

IsaacMilarky:
**Description**
- Add an option to be able to toggle collection for commit messages

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -52,7 +52,8 @@ default_config = {
                 "rebuild_caches": 1,
                 "run_analysis": 1,
                 "run_facade_contributors": 1,
-                "facade_contributor_full_recollect": 0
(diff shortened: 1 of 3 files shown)


sgoggins:
@IsaacMilarky : I think you said this was ready for review when we met the other day. Checking back because its still in "draft mode".

MoralCode:
My $0.02: could we make the new option more descriptive? Like "facade_process_commit_messages"?

Other users may not know what this is intuitively and the config is going to be most peoples main interaction with augur internals

sgoggins:
@IsaacMilarky : I think @MoralCode makes a good suggestion.

sgoggins:
@IsaacMilarky : Can you let us know if this is finished up?

PR #3342, opened 2025-10-26
Fix null cntrb email 

razekmh:
**Description**
- Please include a summary of the change.

This PR fixes a portion of https://github.com/UCL/open-source/issues/26

**Notes for Reviewers**
During the data collection phase we were watching the [Contributor](https://github.com/chaoss/augur/blob/b192e4c043ea33e2fb2e06105f540a4c7d3b0096/augur/application/db/models/augur_data.py#L146) table. We focused on the columns `cntrb_email` and `cntrb_canonical`. We limited our test to one repo at the time but the same situation was observed in multi repo collection. We noticed that the count of the `cntrb_email` goes up and then slowly starts to go down with the collected emails replaced with Null values.  We assume this is not the expected behavior? 

Looking at PostgreSQL logs (we enabled the logs to investigate the situation) we could see that there was a upsert statement (insert and update on conflict) in the logs. 
for example (I modified the identifiers a bit but the structure is the same). The statement have many users at once. I limited it to one for the example)
```sql
INSERT INTO augur_data.contributors (
        cntrb_id,
        cntrb_login,
        cntrb_email,
        cntrb_full_name,
        cntrb_company,
        cntrb_created_at,
        cntrb_location,
        cntrb_canonical,
        cntrb_last_used,
        gh_user_id,
        gh_login,
        gh_url,
        gh_html_url,
        gh_node_id,
        gh_avatar_url,
        gh_gravatar_id,
        gh_followers_url,
        gh_following_url,
        gh_gists_url,
        gh_starred_url,
        gh_subscriptions_url,
        gh_organizations_url,
        gh_repos_url,
        gh_events_url,
        gh_received_events_url,
        gh_type,
        gh_site_admin,
        tool_source,
        tool_version,
        data_source
    )
VALUES (
        '09000099-9900-0000-0000-000000000000'::uuid::UUID,
        'example_user',
        NULL,
        NULL,
        NULL,
        NULL,
        NULL,
        NULL,
        NULL,
        555555,
        'example_user',
        'https://api.github.com/users/example_user',
        'https://github.com/example_user',
        'XXQxxXXlxxX1XXA5',
        'https://avatars.githubusercontent.com/u/555555?v=4',
        '',
        'https://api.github.com/users/example_user/followers',
        'https://api.github.com/users/example_user/following{/other_user}',
        'https://api.github.com/users/example_user/gists{/gist_id}',
        'https://api.github.com/users/example_user/starred{/owner}{/repo}',
[...]

Issue #3345, opened 2025-10-27
Epic: Fix large code duplication


MoralCode:
Several observations have bene made about ways in which major components/functions of the codebase are duplicated, requireing changes to be made in more than one place.

- #3338 
- #3357 (helps fix duplication in reading from config 
~~- The get_section in lib.py (more reading from config dupe not addressed by the above) #3357 ~~
- augur backend start cli function and augur api start cli function seem to be very similar...
- bulk_insert_dicts and session.insert_data code seems very similar (found since https://github.com/chaoss/augur/pull/3342/changes updated both of them)
- the get_value in lib.py is duplicated with config stuff
- facade has an entire config helper to itself... seems redundant

Is there a historical reason for this?

@sgoggins

sgoggins:
I think this is almost certainly technical debt that was introduced when we were refining contributor resolution.

razekmh:
There are a few files with suffix `_old` two of them are in the [data models module](https://github.com/chaoss/augur/blob/main/augur/application/db/models/augur_data_old.py)

MoralCode:
looking specifically at the two data insertion methods touched by #3342 - a codebase wide search indicates that

insert_data (2020, seemingly no specific reason to add) is used by the tables/models:
- CollectionStatus
- Config
- Repo
- UserGroup/UserRepo
- Contributor
- Message
- PR/IssueMessageRef
- PullRequest
- PRAssignee/reviewer/Meta
- PRReviewMessageRef
- PRFile/PRCommit


bulk_insert_dicts (created apr 2024 by andrew to remove db session dependence) is used by:
- Contributor
- ContributorRepo
- CommitMesage
- RepoDepsLibyear
- RepoDependency
- RepoDepsScorecard
- RepoLabor
- Issue/PREvent
- Issue Label/Assignee
- RepoClone
- Repo
- ContributorsAlias
- UnresolvedCommitEmail
- PRLabel/Assignee/Reviewer/Meta
- Message
- PRReviewMessageRef
- Release
- repoInfo
- Issue/PREvent
- Issue
- IssueMessageRed
- PullRequest
- and more.....


So this seems more complex than just duplicate code....

PR #3354, opened 2025-10-29
apply coalesce to values being bulk-added to handle nulls

### LINKED ISSUE #3317: Contributors information being removed from the `augur.contributors` table during data collection
## Symptom of the issue

We're collecting data with Augur on a bunch of repos, and we noticed something odd on https://github.com/UCL/rsd-engineeringcourse specifically, but this is probably happening in others as well, we're simply using UCL/rsd-engineeringcourse as a canary, but there's nothing special about it.

What we observe is that during the initial data collection, if we monitor the rows with non-`NULL` value for `cntrb_email` in the `augur.contributors` table (`SELECT * FROM augur.contributors WHERE cntrb_email IS NOT NULL;`), this list shrinks over time, indicating that some data is being lost, or more likely overwritten incorrectly.

Possibly related, in the `augur.commits` table we found that some commits don't have the `cmt_ght_author_id` field set correctly, just as an example commit https://github.com/UCL/rsd-engineeringcourse/commit/fffaf9ad163c9c6ebb4128c69f4ce9d40e465d18 isn't attributed to any GitHub user, when it should.

This behaviour is fully reproducible on UCL/rsd-engineeringcourse.  For the record, we're working on top of commit 60138d072849647c815d90b65f405d7b307e1109.

## Possible explanation

This line https://github.com/chaoss/augur/blob/60138d072849647c815d90b65f405d7b307e1109/augur/tasks/github/pull_requests/tasks.py#L115 is inserting data in the database with `on_conflict_update=True` https://github.com/chaoss/augur/blob/60138d072849647c815d90b65f405d7b307e1109/augur/application/db/session.py#L96 executing an SQL query which looks like
```sql
INSERT INTO ...
ON CONFLICT (cntrb_id) DO
UPDATE
SET cntrb_id = excluded.cntrb_id,
...
```
possibly `NULL`-ifying non-`NULL` fields.  Comparing emails coming from different sources (git commit vs public github profile fetched for PRs and issues) is suspiciously wrong, as they need not match, and a user can have several email addresses associated to their email account, none of them necessarily matching the one in the public profile, if any.

***Edit***: one small component of the issue is this typo https://github.com/chaoss/augur/blob/c5eeb072ae5c541e5fbb56295b3afab5dffaf4b0/augur/application/db/models/augur_data.py#L2888 (it should be `obj`, not `ojb`), but there's more to it, that's not the full story.

MoralCode:
**Description**
A collaborative change built with the other augur maintainers (@sgoggins and @ABrain7710) to add the `coalesce` postgres function to our bulk insert methods to ensure that values in the database cannot be overwritten with NULL values. This was inspired by @razekmh's solution in #3342

This was implemented without a parameter to toggle this functionality off because we couldnt think of a scenario where one might want to intentionally overwrite data with null in augur 

This PR fixes #3317

This supersedes #3342 

**Notes for Reviewers**
should be all set to merge once the OP of the linked issue confirms this still fixes the issue.

**Signed commits**

### DIFF
--- a/augur/application/db/session.py
+++ b/augur/application/db/session.py
@@ -143,8 +143,10 @@ class DatabaseSession(Session):
 
             # create a dict that the on_conflict_do_update method requires to be able to map updates whenever there is a conflict. See sqlalchemy docs for more explanation and examples: https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#updating-using-the-excluded-insert-values
             setDict = {}
+            base_table = getattr(table, "__table__", table)
             for key in data[0].keys():
-                setDict[key] = func.coalesce(getattr(stmnt.excluded, key), getattr(table.c, key))
+                existing_col = getattr(base_table.c, key)
+                setDict[key] = func.coalesce(getattr(stmnt.excluded, key), existing_col)
                 
             stmnt = stmnt.on_conflict_do_update(
                 #This might need to change
--- a/augur/application/db/lib.py
+++ b/augur/application/db/lib.py
@@ -354,8 +354,10 @@ def bulk_insert_dicts(logger, data_input: Union[List[dict], dict], table, natura
 
         # create a dict that the on_conflict_do_update method requires to be able to map updates whenever there is a conflict. See sqlalchemy docs for more explanation and examples: https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#updating-using-the-excluded-insert-values
         setDict = {}
+        base_table = getattr(table, "__table__", table)
         for key in data[0].keys():
-            setDict[key] = func.coalesce(getattr(stmnt.excluded, key), getattr(table.c, key))
+            existing_col = getattr(base_table.c, key)
+            setDict[key] = func.coalesce(getattr(stmnt.excluded, key), existing_col)
             
         stmnt = stmnt.on_conflict_do_update(
             #This might need to change


giordano:
I did a quick smoke test and it looks like this PR fixes our problem with some commits, like https://github.com/UCL/rsd-engineeringcourse/commit/fffaf9ad163c9c6ebb4128c69f4ce9d40e465d18, not being attributed to any user, as [it's currently happening](https://github.com/chaoss/augur/issues/3317#issue-3526950995).  `SELECT c.* FROM augur_data.contributors AS c WHERE cntrb_email IS NOT NULL` is quickly filled up and doesn't decrease over time anymore.    Thank you!

MoralCode:
Sweet! Since I made modifications since we wrote this and it impacts a critical part of the code, I'd like to get @ABrain7710's input on this before merging

augur/application/db/session.py
```diff
@@ -142,8 +143,10 @@ def insert_data(self, data_input: Union[List[dict], dict], table, natural_keys:
 
             # create a dict that the on_conflict_do_update method requires to be able to map updates whenever there is a conflict. See sqlalchemy docs for more explanation and examples: https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#updating-using-the-excluded-insert-values
             setDict = {}
+            base_table = getattr(table, "__table__", table)
             for key in data[0].keys():
-                    setDict[key] = getattr(stmnt.excluded, key)
+                existing_col = getattr(base_table.c, key)
+                setDict[key] = func.coalesce(getattr(stmnt.excluded, key), existing_col)
```
samcunliffe, line 149: Same lines of code in two places, should this be abstracted out?

samcunliffe: Also, should you tag/co-authored-by @razekmh somewhere? Maybe mention in the merge commit?

MoralCode: > Same lines of code in two places, should this be abstracted out?

if you are referring to session.py vs lib.py, then yes. In fact this fix is what sparked #3345, which I plan to fix quite soon, just maybe not in this release

> Also, should you tag/co-authored-by @razekmh somewhere? Maybe mention in the merge commit?

yep, I can do that! Also edited this PR description to explicitly mention it. Even though we only took a tiny piece of the original PR, the inspiration was definitely helpful

ABrain7710 (approved):

Issue #3356, opened 2025-10-29
Handle missing config entries in a self-healing way


MoralCode:
https://github.com/chaoss/augur/pull/3303 was merged to write a config value to the database using migrations.

Migrations are primarily intended for structural database changes.

Instead, I think that the config managment class we currently have should be able to detect a missing config value and insert it (with its default value), thereby automatically healing from missing config items.

MoralCode:
This need to add new config values to the DB affects both docker and non-docker installs, although its less impactful on docker because new values should be added to the hardcoded object of default values which should get picked up at startup by the init script we have set up for that environment.

MoralCode:
we want to do a tiered approach
- config table
- augur.json
- hardcoded defaults

sgoggins:
For bare metal instances I created these three rows until we get to self healing: 

```sql
INSERT INTO "augur_operations"."config" ( "section_name", "setting_name", "value", "type") VALUES ('Celery', 'core_worker_count', '79', 'int');
INSERT INTO "augur_operations"."config" ( "section_name", "setting_name", "value", "type") VALUES ('Celery', 'secondary_worker_count', '50', 'int');
INSERT INTO "augur_operations"."config" ("section_name", "setting_name", "value", "type") VALUES ('Celery', 'facade_worker_count', '80', 'int');
```

MoralCode:
Turns out a couple contributors who have recently been in slack have manually-configured setups.

PR #3357, opened 2025-10-29
Deduplicate/lib get_section - and some other config improvements

MoralCode:
**Description**
`lib.py` has a function called `get_section` that retrieves a value from the config database. This function duplicates functionality from the `augur.application.config.AugurConfig` class.

It is currently causing issues with new installs. Users are seeing python `KeyError` when trying to read config values that are newly added to the database, such as `facade_contributor_full_recollect` that was added in #3303.

While this PR doesnt fix the `KeyError`, it ensures that the config is not being read in multiple ways, allowing for a simpler implementation in #3356 

This PR migrates the remaining few uses of the old function over to the new implementation

**Notes for Reviewers**

**Signed commits**


MoralCode:
Runs on my machine. Found the missing import causing issues with the CI jobs

sgoggins (approved):
LGTM!

PR #3368, opened 2025-11-03
Fixed issue with static log time | First time contributor

Vatsal-Patel-09:
**Description**
- Fixed misleading log message for materialized views refresh task.
- The log claimed it runs "every night at 1am CDT" but actually runs every N days relative to Augur startup time using.
- Updated the message to display the actual configurable interval and removed the incorrect timezone reference.

This PR fixes #3360 

**Notes for Reviewers**
- I am first time contributor trying to get hands on with Open-Source
- feel free to point out error(s) or possible improvement(s) would love to hear and learn!

**Signed commits**
- [x] Yes, I signed my commits.

PR #3370, opened 2025-11-03
Fix: make refresh-materialized-view log dynamic with interval and time zone

### LINKED ISSUE #3360: Log message incorrectly referenced a set schedule
This log line claims that things are being refreshed at 1am CDT when the interval is configurable

https://github.com/chaoss/augur/blob/0c57454a27f298d67d8d61f5d78250d2127b540e/augur/tasks/init/celery_app.py#L245

- [ ] confirm what time, if any, this job runs, or if its likely to be relative to when augur starts up
- [ ] confirm whether augur is using the system timezone
- [ ] update the log message to specify the correct number of days and either remove or print  the correct timezone dynamically

YashSejani:
**Description**
This PR fixes an incorrect log message in the Celery initialization code.  
Previously, the log line hardcoded “Scheduling refresh materialized view every night at 1am CDT”,  
which was misleading because the refresh interval is configurable and not tied to a specific timezone.

This PR updates the message to:
- Dynamically show the configured refresh interval (`mat_views_interval` days)
- Display the system timezone
- Log the next expected refresh time relative to Augur startup

This improves clarity and accuracy for users running Augur in different environments.

Fixes #3360

**Notes for Reviewers**
- The job still runs based on `mat_views_interval` (default from config).
- Timezone is derived dynamically from the system settings.
- Tested locally to confirm the new log output prints correctly.

**Signed commits**

### DIFF
--- a/augur/tasks/init/celery_app.py
+++ b/augur/tasks/init/celery_app.py
@@ -7,6 +7,8 @@ import datetime
 import traceback
 import inspect
 import celery
+import time
+from datetime import datetime, timedelta
 from celery import Celery
 from celery import current_app 
 from celery.signals import after_setup_logger
@@ -242,8 +244,6 @@ def setup_periodic_tasks(sender, **kwargs):
 
         mat_views_interval = int(config.get_value('Celery', 'refresh_materialized_views_interval_in_days'))
         if mat_views_interval > 0: 
-            import time
-            from datetime import datetime, timedelta
             # system timezone
             timezone_name = time.tzname[0]
             # next expected refresh time


augur/tasks/init/celery_app.py
```diff
@@ -242,7 +242,18 @@ def setup_periodic_tasks(sender, **kwargs):
 
         mat_views_interval = int(config.get_value('Celery', 'refresh_materialized_views_interval_in_days'))
         if mat_views_interval > 0: 
-            logger.info(f"Scheduling refresh materialized view every night at 1am CDT")
+            import time
+            from datetime import datetime, timedelta
```
MoralCode: Can you move these imports to the top of the file and deduplicate them with existing imports if they are already present?

YashSejani: okk

augur/tasks/init/celery_app.py
```diff
@@ -7,6 +7,8 @@
 import traceback
 import inspect
 import celery
+import time
+from datetime import datetime, timedelta
```
MoralCode, line 11: This module for `datetime` (on the right side of the import) is using the same name as the package imported about 5 lines above this (just above `import traceback`). This is likely going to cause issues with calls in the code that call `datetime.datetime.something()`. 

Id suggest adjusting your calls to `datetime` and `timedelta` to be `datetime.datetime` and `datetime.timedelta` respectively

MoralCode:
Also be aware that this PR, #3368, and #3370 are all solving the same issue, and only one of them can be merged.

PR #3389, opened 2025-11-10
Ensure Docker logging facilities can see gunicorn errors

MoralCode:
**Description**
Currently, if gunicorn crashes due to an error, the docker logs will only show the attempt to start the gunicorn (frontend) web server processes followed by their immediate shutdown. The cause is only findable by looking in the configured log directory and reading the log files.

This PR uses the environment variable for docker deployments to change the gunicorn config to redirect errors to stdout so they get picked up by docker

**Notes for Reviewers**
So uuh, I fixed the issue that was causing gunicorn to crash before realizing i should use it to test this, so while im not certain it works, it should be a pretty basic change and lines up with the [gunicorn docs for this errorlog setting](https://docs.gunicorn.org/en/stable/settings.html#errorlog)

**Signed commits**

### DIFF
--- a/augur/api/gunicorn_conf.py
+++ b/augur/api/gunicorn_conf.py
@@ -39,8 +39,14 @@ del is_dev
 
 # set the log location for gunicorn    
 logs_directory = get_value('Logging', 'logs_directory')
+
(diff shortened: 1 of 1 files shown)


sgoggins (approved):
LGTM

PR #3412, opened 2025-11-17
Fix database connection leaks in DEI routes and user CLI commands

### LINKED ISSUE #3401: Database Connection Leaks in API Routes and CLI Commands
**Description:**
Three database connection leaks might cause "too many clients" errors. Sessions are created but never closed.

**Affected files:**
  1. `augur/api/routes/dei.py:112` - `dei_report()` endpoint never closes session
  2. `augur/api/routes/dei.py:39` - `dei_track_repo()` endpoint lacks exception handling for session cleanup
  3. `augur/application/cli/user.py:67` - `reset_password()` command never closes session

**Code locations:**

**Issue 1:** `augur/api/routes/dei.py:112-132`
```python
def dei_report(application: ClientApplication):
    session = DatabaseSession(logger, engine=current_app.engine)
    # ... operations ...
    return send_file(report.resolve())  # Missing session.close()
```
**Issue 2: augur/api/routes/dei.py:39-101**
```python
def dei_track_repo(application: ClientApplication):
    session = DatabaseSession(logger, engine=current_app.engine)  # No context manager
    # ... 60+ lines of operations ...
    session.close()  # Line 99 - not reached if exception occurs
```
**Issue 3: augur/application/cli/user.py:67-78**
```python
def reset_password(username, password):
    session = Session()
    # ... operations ...
    session.commit()
    return click.echo("Password updated")  # Missing session.close()
```
---
Note: Initial scan performed using Claude Code to identify potential database leaks across the codebase. I've manually verified these by reviewing the actual code.

shlokgilda:
**Description**
  - Fixed database connection leaks in DEI API endpoints and user CLI commands. Fixes #3401.
  - Three functions were creating database sessions without properly closing them in all code paths (early returns, exceptions)
  - Converted `dei_track_repo()` and `dei_report()` to use context managers for automatic session cleanup
  - Added try-finally blocks to `add_user()` and `reset_password()` CLI commands to guarantee session cleanup
  - Fixed missing import in `dei_track_repo()`
  - Removed non-functional `session.autocommit = True` (has no effect in SQLAlchemy 2.x) and added explicit `session.commit()` to persist BadgingDEI records

  **Notes for Reviewers**
  - `dei_report()` scoping is critical: session must stay open during template rendering (lazy-loads `project.repo`), then closes before file operations
  - All changes follow existing context manager patterns used elsewhere in the codebase (e.g., `augur/application/cli/db.py`)
  - Manually tested `augur user add` and `augur user password_reset` commands - sessions properly close in all paths
  - DEI endpoints (`dei_track_repo`, `dei_report`) should be manually tested (although I don't think these are even used?)
  
  **Signed commits**

### DIFF
--- a/augur/api/routes/dei.py
+++ b/augur/api/routes/dei.py
@@ -12,7 +12,7 @@ from augur.api.util import api_key_required, ssl_required
 from augur.application.db.models import ClientApplication, CollectionStatus, Repo, RepoGroup, BadgingDEI
 from augur.application.db.session import DatabaseSession
 
-from augur.tasks.util.collection_util import CollectionRequest,AugurTaskRoutine, get_enabled_phase_names_from_config, core_task_success_util
+from augur.tasks.util.collection_util import CollectionRequest,AugurTaskRoutine, get_enabled_phase_names_from_config_session, core_task_success_util
 from augur.tasks.start_tasks import prelim_phase, primary_repo_collect_phase
 from augur.tasks.github.util.util import get_repo_weight_by_issue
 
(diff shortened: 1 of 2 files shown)


MoralCode:
> * to persist BadgingDEI records
> DEI endpoints (`dei_track_repo`, `dei_report`) should be manually tested (although I don't think these are even used?)

i think that table is unused and i plan to drop it soon lol (just filed https://github.com/chaoss/augur/issues/3423)
I suspect you are right about the routes too.

> Fixed database connection leaks in DEI API endpoints and user CLI commands

funny you say that. I'm fairly sure that the DatabaseSession object itself is causing leaks, even when used with `with` based on chatting with other maintainers. Gen AI suggests that this is because the methods often returns raw `Result` objects due to db calls not using functions like `.one()` that specify what kind of results they want, meaning connections are held onto and brought outside the with scope.


Overall it seems like you are trying to do several things in this PR that may be better to split into separate, more focused PRs.

shlokgilda:
Okay, yeah! The "too many things in one PR" sounds like me. 

I'll try to be more vigilant of this. 

Re DatabaseSession: I didn't check that. If that's the case, that should fix a whole bunch of issues.

ABrain7710 (approved):

MoralCode:
@shlokgilda are there tests to confirm that the core of this change does in fact leak fewer connections?

shlokgilda:
Hmm, good question! I don't think so. Might have to figure out a way to calculate the number of connections before/after the fix.

Any suggestions?

sgoggins (approved):
LGTM

PR #3414, opened 2025-11-17
[keyman] Fix typo and improve documentation: Fixes #3413

### LINKED ISSUE #3413: keyman performance improvements
## Background

Working with the keyman orchestration system and found a few issues that I think should be addressed.

## Issues

**Key distribution isn't fair**
When you publish the same key to the same platform twice, it gets added to the pool multiple times and receives extra traffic. Not common but should be prevented.

```python
# Without duplicate check, key appears twice in pool
keypub.publish("ghp_abc123", "github_rest")
keypub.publish("ghp_abc123", "github_rest")  # duplicate
# Output: fresh_keys["github_rest"] = ["ghp_abc123", "ghp_abc123"]
```

**Typo in exception class**
`WaitKeyTimeout` has `tiemout_seconds` instead of `timeout_seconds`. Found this while debugging.

**Missing documentation**
- No clear docs on how many Redis connections keyman creates
- The directory could benefit from a directory-specific README. 

## What I'm planning to fix
1. Add a duplicate check in `publish_key()` - simple `if key not in list` before appending
2. Fix the typo
3. Document the connection pooling situation
4. Improve docstrings
5. Write a proper README for the keyman directory

## Testing
Should be straightforward. Publish the same key twice and make sure it only appears once in the pool

shlokgilda:
**Description**
Improves keyman code quality and documentation. Fixes #3413 
- Fixed typo: `tiemout_seconds` → `timeout_seconds` in `WaitKeyTimeout` exception
- Added type hints to all functions
- Standardized all docstrings with Args/Returns/Raises sections
- Created `keyman/README.md` with architecture overview and usage examples

**Notes for Reviewers**
- No API changes to `KeyClient` or `KeyPublisher`. 
- Renamed `min` to `min_timeout` in `Orchestrator.py` function `new_key()`.

**Note:** Docstrings and README.md were generated using Claude Code. Verified manually.

**Signed commits**


keyman/Orchestrator.py
```diff
@@ -43,35 +58,61 @@ def __init__(self) -> None:
...
+        """
         if platform not in self.fresh_keys:
             self.fresh_keys[platform] = [key]
             self.expired_keys[platform] = {}
             self.invalid_keys[platform] = set()
         else:
-            self.fresh_keys[platform].append(key)
+            # Prevent duplicate keys from increasing selection probability
+            if key not in self.fresh_keys[platform]:
+                self.fresh_keys[platform].append(key)
```
MoralCode: since this is a functional change, I suspect it would be best to move to a separate PR for more thorough review

MoralCode (changes requested):
The docs and typing improvements generally seem okay (make sure to add this file or the whole keyman directory to the list of things `mypy` will typecheck)

Before making substantive changes, I think we should really have some thorough unit testing for keyman to make sure we dont break some edge case. I'd recommend doing that as a separate PR, as well as making any substantive changes to keyman logic in another, different PR

keyman/Orchestrator.py
shlokgilda: Fair enough. Closing this PR. Will start with a new PR for unit-tests for keyman, followed by this functional bug fix, and then docs.

MoralCode: @shlokgilda we have unit tests (mostly a smoke test) i just learned

MoralCode: early versions of keyman had probabistic weighting for preferring certain keys, so this is somewhat legacy and just left this way because its not hard to change.

if we want keys to be unique in the key list we should just replace it with a set anyway because sets gaurantee uniqueness

MoralCode:
Why the heck is pylint complaining about the `self.REQUEST` thing not existing? even before this PR....

shlokgilda:
I think it's because REQUEST is set dynamically via `setattr` in `__init__` - pylint's static analysis can't trace that. It just sees there's no class-level REQUEST attribute defined anywhere. Not 100% sure but that's my guess. Could add type hints at class level to  fix it, or just disable the warning there

MoralCode:
I just checked and this pr looks like it does what it says - README, docstrings, and a typo fix (so consider the previous request to remove the fix for duplicate keys as resolved).

However, given 

> Note: Docstrings and README.md were generated using Claude Code. Verified manually.

and the fact that a large portion of this PR is docstrings and README, I think @Ulincsys (the author of keyman) should probably verify that the documentation matches the intentions of keyman before this is merged

keyman/KeyOrchestrationAPI.py
```diff
@@ -92,8 +91,21 @@
 }
 
 class WaitKeyTimeout(Exception):
-    def __init__(self, timeout_seconds) -> None:
-        self.tiemout_seconds = timeout_seconds
+    """Raised when waiting for a key exceeds the specified timeout.
```
Ulincsys: WaitKeyTimeout is raised when the Key Orchestrator returns a `wait` message, indicating that there are no fresh keys available for the requested platform.

The `timout_seconds` parameter indicates to the client how long it needs to sleep before a fresh key will become available.

shlokgilda: Addressed the comment @Ulincsys . Thanks for the keen eye!

MoralCode (approved):
Approving since this has been reviewed by @Ulincsys and the feedback has been addressed

MoralCode:
@shlokgilda rebasing should get rid of the else after return lint warnings (and may contribute to making reviewdog happy)

also @Ulincsys should re-review now that the feedback is addressed

MoralCode:
ok... what is it with this specific PR thats causing docker to now exhibit the same weird timeout issue as in #3527

shlokgilda:
I don't think this has anything to do with the podman issue in #3527. Only the linter is failing here and it's because pylint can't detect `KeyClient.REQUEST` and `KeyPublisher.ANNOUNCE` - they're set dynamically via `setattr` in `__init__`. Discussed this same thing back on Dec 19.

Quick fix: add class-level type annotations or just `# pylint: disable=no-member` where needed.

The podman test actually passes (all patterns matched) - it just fails during cleanup due to #3530, which is unrelated to this PR.

Should I go ahead and disable the pylint rule where needed since we decided to keep using `setattr`?

MoralCode:
> Should I go ahead and disable the pylint rule where needed since we decided to keep using `setattr`?

If you can add like a tag to the affected keyman class so it doesnt warn about missing properties for this class only that would be nice since i want to keep this check for the rest of the codebase.

if we have to individually tag each usage, thats a pain but maybe still worth doing so we dont learn to ignore the linter

shlokgilda:
Done. Added `# pylint: disable=no-member` at the class level for both `KeyClient` and `KeyPublisher`

shlokgilda:
Aah, I should go ahead and change my default merge strategy. Will do that.

MoralCode:
> Aah, I should go ahead and change my default merge strategy. Will do that.

i looked at these settings for the whole repo, but its frustratingly singular - like i cant say "i want to allow merge commits when merging PRs, but have rebase be the default when updating a PR before it merges"

MoralCode:
This is a docs only fix and seems to have 1.5 approving reviews (johns feedback seems to have been addressed)

MoralCode:
tests pass. merging with bypass.

Issue #3319, opened 2025-10-21
Augur "forgets" about a running task


MoralCode:
What Happens/symptoms:
1. a task begins running in augur
2. *something happens*
3. rabbitmq logs show the following error (could also be logged as a warning):
```
2025-10-14 15:37:05.560429+00:00 [error] <0.1187.0> Error on AMQP connection <0.1187.0> (172.18.0.5:48982 -> 172.18.0.3:5672, vhost: 'augur_vhost', user: 'augur', state: running), channel 0:
2025-10-14 15:37:05.560429+00:00 [error] <0.1187.0>  operation none caused a connection exception connection_forced: "broker forced connection closure with reason 'shutdown'"
```
4. the task remains in a "started" state in flower forever. counting flower tasks shows  processed = failed + succeeded - 1
5. according to Augur's DB the task is still running (i.e. 
`SELECT cs.* FROM augur_operations.collection_status AS cs
WHERE (core_status <> 'Success') or (secondary_status <> 'Success') or (facade_status <> 'Success')` shows atask "collecting"

This seems to affect facade

So far all instances of this have been on rabbitmq 3.12.14 since the fix for #3255  hasnt fully shipped yet

MoralCode:
[this](https://github.com/rabbitmq/rabbitmq-server/discussions/13627) discussion on rabbits repo seems to suggest that something about the application code may be preventing rabbit heartbeats for 2 minutes and causing this. unsure if that applies here because we see this anywhere from a few to a dozen times a month or so, so its probably somewhat intermittent. the logs also dont indicate missing heartbeats

MoralCode:
this google groups thread seems to also start drilling down into the "broker forced connection closure" part: https://groups.google.com/g/rabbitmq-users/c/a_zJuidbHic?pli=1

MoralCode:
@sgoggins posted this SQL query in the 10/28 augur meeting as a band-aid/way to fix (by marking as collected) facade tasks that may have been forgotten about if theres a rabbit connection issue (i.e. the tasks keep running and the commit counts match between the clone and what GH says they should be, but because rabbit forgets about it, they maybe dont get marked as completed)

Use with caution

```sql
WITH counts AS (
    SELECT c.repo_id, COUNT(DISTINCT c.cmt_commit_hash)::bigint AS commits_captured
    FROM augur_data.commits c
    GROUP BY c.repo_id
),
eligible AS (
    SELECT a.repo_id
    FROM augur_operations.collection_status a
[...]

Issue #3371, opened 2025-11-03
Proposal: Add generative AI disclosure checkbox next to the DCO checkbox in the PR template


MoralCode:
Given that many other projects (such as fedora) are adopting genAI policies, I think it may be helpful to at least give contributors a clear opportunity to disclose their use of Generative AI in the form of an easy checkbox alongside the existing DCO checkbox.

I think this should look something like:

- [ ] I used Generative AI (chatbots, cursor, etc) to assist me with this contribution

This would be followed by a comment offering an optional place to explain in what way it was used (i.e. planning vs writing code vs translating etc.

PredictiveManish:
One small consideration, does it want to clarify that this checkbox is purely informational and doesn’t affect acceptance criteria?

That might help prevent hesitation from new contributors who use AI minimally (e.g, for grammar or formatting suggestions).

MoralCode:
I'm not aware of CHAOSS having a formal policy prohibiting AI contributions.

My view is that all code contributions should be held to the same standards for quality, legality, etc. As long as the submitter understands and is willing to take responsibility for the code they are contributing (including engaging with maintainers to make changes to the code in order for it to pass code review), I don't really see much difference between a quality contribution built with whatever degree of AI use and a quality human-built contribution.

Some of this responsibility is already required in the form of the commit signoff, which is meant to signify that the developer whose name is ultimately attached to the work agrees that the terms of the [Developers Certificate of Origin](https://developercertificate.org/) have been met (i.e. they have the right to contribute the code, etc).

I mostly propose this partly out of my own curiosity, and also partly as a learning exercise. I suspect there will soon be a growing interest in other developers and other projects wanting to learn more about the proliferation of AI tools in the open source world. If we can experiment with possible ways to gather this information in a voluntary way on our own project, that information will end up in Augur (via PR comments and/or commit messages) and allow us to develop queries and other ways to analyze the information in a way that could help other projects better understand to what degree their own communities use AI

shlokgilda:
I think this is a great idea. I agree with the point that at the end of the day, the committer is responsible for any and all code they submit - whether it's written entirely by hand, with AI assistance, or anywhere in between.

[...]

PR #3420, opened 2025-11-19
Fix deadlock issues by implementing timeout handling for git operations

shlokgilda:
**Description**
Fixes subprocess pipe deadlock in facade git operations that caused tasks to hang indefinitely on large repositories.

**The Problem:**
Facade tasks were getting stuck for 12+ hours on `git checkout` operations when processing large repos (600+ MB pack files). The issue was a subprocess pipe buffer deadlock - git would write more than 64KB of output to stdout/stderr, fill the OS pipe buffer, and block waiting for someone to read it, while Python blocked waiting for git to finish. Neither could proceed.

Additional details in comments of issue #3319.

**The Fix:**
1. **Replaced `Popen().wait()` with `subprocess.run()`** - Modern API that handles process lifecycle correctly
2. **Added `stdout=DEVNULL, stderr=DEVNULL`** - Prevents pipe creation entirely by redirecting to `/dev/null`. Git output is immediately discarded by the kernel, so no buffer can fill up.
3. **Added conservative timeouts** - Protects against truly hung operations:
   - 2 hours for `git clone` (large repos)
   - 10 minutes for `git pull` and `git checkout`
   - 5 minutes for `git reset` and `git clean`
   - 1 minute for branch/remote queries
4. **Used `check=False`** - Manual error handling for retry logic and custom error messages

**Why `subprocess.run()` is sufficient:**
- Internally calls `.wait()` for us (blocking call)
- With `capture_output=True`, internally calls `.communicate()`
- With `stdout=DEVNULL`, no communication needed (no pipes to read)
- Cleaner, safer API that prevents the deadlock pattern entirely

**Testing:**
- Verified all subprocess calls now have appropriate timeouts
- Confirmed no orphaned exception handlers
- Pattern follows Python subprocess best practices

**Notes for Reviewers**
The key insight is that `subprocess.run()` with `stdout=DEVNULL` creates no pipes at all - output goes directly to the kernel's null device. This makes deadlock impossible regardless of output size. The timeouts are conservative to allow legitimate long operations while preventing infinite hangs.

**Signed commits**

**GenAI Disclosure:** Used Claude Code to draft this PR.


MoralCode:
holy git ops batman - i didnt think there were so many places we did git subprocesses lol

MoralCode:
> GenAI Disclosure: Used Claude Code to draft this PR.
Do you mean just the PR description? or the pr diff contents itself?

shlokgilda:
> > GenAI Disclosure: Used Claude Code to draft this PR.
> > Do you mean just the PR description? or the pr diff contents itself?

Aah, my bad! I meant just the PR description.

MoralCode:
> Aah, my bad! I meant just the PR description.

no worries, just checking! cant wait to find time to finish the policy part of #3371 so we can have a better template for people to use

sgoggins:
Outstanding contribution! Reviewing .

augur/tasks/git/util/facade_worker/facade_worker/repofetch.py
```diff
@@ -149,7 +149,18 @@ def git_repo_initialize(facade_helper, session, repo_git):
...
+                cmd, shell=True,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=7200,  # 2 hours for large repos
+                check=False
+            )
+            return_code = result.returncode
+        except subprocess.TimeoutExpired:
+            facade_helper.log_activity('Error', f'Git clone timed out: {cmd}')
+            return_code = -1  # Timeout error code
```
MoralCode: It seems like we reuse this pattern a lot - could we potentially refactor it into a single function (maybe that's what we can use FacadeHelper for) to deduplicate a lot of this process running wrapper code? especially one where the option for discarding output is explicit

shlokgilda: Great idea. Pushed the changes. 

GenAI Disclosure: Used Claude Code to write the new `run_git_command` in `FacadeHelper` class. Verified the accuracy of the code by manual testing and restarting augur.

augur/tasks/git/util/facade_worker/facade_worker/config.py
```diff
@@ -254,3 +255,48 @@ def insert_or_update_data(self, query, **bind_args)-> None:
...
+                )
+                return result.returncode, result.stdout.strip()
+            else:
+                result = subprocess.run(
+                    cmd, shell=True,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    timeout=timeout,
+                    check=False
+                )
```
MoralCode: Lots of these options seem duplicated - i wonder if it would be best to create a dict of the common options and then append the correct set of options depending on `capture_output` and then expand those as  kwargs in a single call to `subprocess.run`

shlokgilda: Addressed in latest commit @MoralCode

augur/tasks/git/util/facade_worker/facade_worker/config.py
```diff
@@ -254,3 +255,53 @@ def insert_or_update_data(self, query, **bind_args)-> None:
...
+                run_options['stdout'] = subprocess.DEVNULL
+                run_options['stderr'] = subprocess.DEVNULL
+
+            result = subprocess.run(cmd, **run_options)
+
+            # Return appropriate output based on capture_output flag
+            if capture_output:
+                return result.returncode, result.stdout.strip()
+            else:
+                return result.returncode, ''
```
MoralCode:
```suggestion
            return result.returncode, result.stdout.strip() if capture_output else ''
```

shlokgilda: In the same file, there are a few more places where we don't use ternary if else blocks. So, I skipped it here as well. Nonetheless, do you think we should use `return result.returncode, (result.stdout.strip() if capture_output else '')` with the parentheses for clarity? Or is it clear to folks that `result.returncode` will always return as the first element?

shlokgilda: Thoughts @MoralCode?

MoralCode: i think the parenthesis would help yeah

MoralCode:
```suggestion
            return result.returncode, (result.stdout.strip() if capture_output else '')
```

MoralCode:
@shlokgilda has been thoroughly testing this and has [confirmed](https://chaoss-workspace.slack.com/archives/C0226ELG6R4/p1764789515373849?thread_ts=1764789018.459519&cid=C0226ELG6R4) that the facade workers are flowing.

I trust this as far as testing goes and am going to mark this as ready

sgoggins:
@MoralCode : Re0running podman end to end, which failed.

MoralCode:
rerunning again. the test passed (all expected log lines seen in the run) but was marked failed because its shutdown took too long and it got killed

sgoggins:
> rerunning again. the test passed (all expected log lines seen in the run) but was marked failed because its shutdown took too long and it got killed

I think your prior comment and the ready tag signal its ready to merge, but since there were commits since your last review @MoralCode , I am double checking.

MoralCode:
I havent re-reviewed to mark it approved because I havent explicitly tested it personally, but i consider all my review comments to be resolved and I trust @shlokgilda's testing

Ulincsys (approved):
LGTM

IsaacMilarky (approved):
Looks good!

PR #3424, opened 2025-11-20
[tasks] Fix memory issues from large in-memory lists in collection tasks

### LINKED ISSUE #3404: Dont store potentially large in-memory lists of all items in `process_pull_request_commits` and `process_pull_request_files`
I'm observing a memory spike when starting up augur that is so large it is causing celery to get OOM killed (albeit by a userland script i have installed).

After analyzing the logs with the help of generative AI, the tool directed me towards several instances of tasks that showed up frequently in the logs and used coding patterns that stored what could be large volumes of data in logs, for example:

in `process_pull_request_commits`:
https://github.com/chaoss/augur/blob/main/augur/tasks/github/pull_requests/commits_model/core.py#L47

in `process_pull_request_files`:
https://github.com/chaoss/augur/blob/main/augur/tasks/github/pull_requests/files_model/core.py#L43

While im not certain that this is the cause of the memory issues, I think this could use refactoring to batch and inserted into the DB in groups of predefined size, rather than stored and inserted all at once.

I wonder if this might also help with some of the DB contention identified in #3343 

Another large in memory list causing actual problems (also theres one about a couple dozen lines below this in the same function):
https://github.com/chaoss/augur/blob/d704670fc80db2397e6f20912ad89f6376aaf372/augur/tasks/github/facade_github/tasks.py#L255

shlokgilda:
**Description**
This PR addresses memory issues when collecting data from repositories with large datasets (10,000+ issues/PRs/contributors). Fixes #3404.

**Key Changes:**
- **Generator pattern for issues**: Prevents loading all issues into memory at once
- **Batch processing**: Insert data in 1000-item batches across all collection tasks
- **`.clear()` over reassignment**: Reuses list objects instead of creating new ones, reducing GC pressure
- **Move inserts outside loops**: In PR reviews, contributors and reviews are already in memory, so batching the final insert is safe and more efficient

**Notes for Reviewers**
- All changes maintain existing logic—only optimization for memory efficiency
- Batch size of 1000 balances memory usage vs. database round trips
- PR reviews refactor moves inserts outside the loop: reduces N database operations to 1 bulk insert (safe since
  `all_pr_reviews` is already in memory)

**Signed commits**

**GenAI Disclosure:** Claude Code was used to generate this PR draft and review diff changes for logical correctness and
potential performance issues.

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -121,7 +121,7 @@ def trim_commits_post_analysis_facade_task(repo_git):
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc)
+    parent_commits = get_parent_commits_set(repo_loc, facade_helper)
 
     # Grab the existing commits from the database
     existing_commits = get_existing_commits_set(repo_id)
@@ -244,7 +244,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc)
+    parent_commits = get_parent_commits_set(repo_loc, facade_helper)
 
     # Grab the existing commits from the database
     existing_commits = get_existing_commits_set(repo_id)
@@ -255,7 +255,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     facade_helper.log_activity('Debug',f"Commits missing from repo {repo_id}: {len(missing_commits)}")
 
     
-    if missing_commits or repo_id is None:
+    if not missing_commits or repo_id is None:
         #session.log_activity('Info','Type of missing_commits: %s' % type(missing_commits))
         return
     
--- a/augur/tasks/github/pull_requests/tasks.py
+++ b/augur/tasks/github/pull_requests/tasks.py
@@ -375,11 +375,10 @@ def collect_pull_request_reviews(repo_git: str, full_collection: bool) -> None:
         for pull_request_id, reviews in all_pr_reviews.items():
 
             for review in reviews:
-
                 if "cntrb_id" in review:
                     pr_reviews.append(extract_needed_pr_review_data(review, pull_request_id, repo_id, platform_id, tool_source, tool_version))
 
-        logger.info(f"{owner}/{repo}: Inserting {len(pr_reviews)} pr reviews")
+        logger.info(f"{owner}/{repo}: Inserting pr reviews of length: {len(pr_reviews)}")
         pr_review_natural_keys = ["pr_review_src_id",]
         pr_review_string_fields = ["pr_review_body",]
         augur_db.insert_data(pr_reviews, PullRequestReview, pr_review_natural_keys, string_fields=pr_review_string_fields)
--- a/augur/tasks/git/util/facade_worker/facade_worker/config.py
+++ b/augur/tasks/git/util/facade_worker/facade_worker/config.py
@@ -29,6 +29,7 @@ import os
 import json
 import logging
 import random
+import subprocess
 from urllib.parse import urlparse
 import sqlalchemy as s
 from sqlalchemy.exc import OperationalError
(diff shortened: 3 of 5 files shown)


shlokgilda:
It's a slightly longer PR, but I think important and within scope since all these issues were possible causes of OOM exceptions.

augur/tasks/github/facade_github/tasks.py
```diff
@@ -262,7 +261,20 @@ def insert_facade_contributors(self, repo_git):
 
     key_auth = GithubRandomKeyAuth(logger)
 
-    process_commit_metadata(logger, key_auth, list(new_contribs), repo_id, platform_id)
+    # Process results in batches to reduce memory usage
+    batch = []
+    BATCH_SIZE = 1000
+
+    for row in result.mappings():
```
MoralCode: is this reliant on having a raw sqlalchemy Result object? I would love to also make sure we are explicit about calling something like `.fetchall()` after the query so that we dont unnecesserily hold on to database connections as we iterate results (that we then forget to close)

augur/tasks/github/facade_github/tasks.py
```diff
@@ -300,10 +312,21 @@ def insert_facade_contributors(self, repo_git):
 
 
     result = execute_sql(resolve_email_to_cntrb_id_sql)
-    existing_cntrb_emails = [dict(row) for row in result.mappings()]
 
-    print(existing_cntrb_emails)
-    link_commits_to_contributor(logger, facade_helper,list(existing_cntrb_emails))
+    # Process results in batches to reduce memory usage
+    batch = []
+    BATCH_SIZE = 1000
```
MoralCode, line 325: should we reuse BATCH_SIZE from above? is there a reason these are two different values/is there a reason we would want the batch sizes for the two halves to be different?

augur/tasks/github/pull_requests/tasks.py
```diff
@@ -389,32 +390,20 @@ def collect_pull_request_reviews(repo_git: str, full_collection: bool) -> None:
...
-
-
-
-
-
-
-
-
-
-
```
MoralCode, line 420: holy whitespace

augur/tasks/github/facade_github/tasks.py
shlokgilda: The idea was to let it be configurable for later use. Maybe we see that one of those queries is faster and so we can possibly increase the batch size for that one query. 

But I guess using just one is also fine.

MoralCode: its ok, this is a bit of a nitpick - i see how it would make sense to leave this in (functionally its identical and its probably not worth merging only to separate later if needed

augur/tasks/github/facade_github/tasks.py
shlokgilda: I didn't realize the original list comprehension was already materializing all results, effectively closing the cursor. I've updated the code to call `.fetchall()` immediately after both queries

MoralCode:
@shlokgilda has been thoroughly testing this and has [confirmed](https://chaoss-workspace.slack.com/archives/C0226ELG6R4/p1764789515373849?thread_ts=1764789018.459519&cid=C0226ELG6R4) that the facade workers are flowing and secondary is not memory bottlenecked anymore.

I trust this as far as testing goes and am going to mark this as ready

MoralCode:
I think I was the one who removed that import because pylint suggested it was unused. I haven't personally tested that change so prob worth reverting just in case.

augur/tasks/git/facade_tasks.py
```diff
@@ -255,7 +255,7 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
     facade_helper.log_activity('Debug',f"Commits missing from repo {repo_id}: {len(missing_commits)}")
 
     
-    if not len(missing_commits) or repo_id is None:
+    if missing_commits or repo_id is None:
```
MoralCode: Did this flip the truthiness of this?

It sounds like not len would be true if len is 0, but the new version is true if missing_commits has nonzero length

MoralCode: i introduced this bug so ill fix it

augur/tasks/github/facade_github/tasks.py
```diff
@@ -262,7 +264,20 @@ def insert_facade_contributors(self, repo_git):
 
     key_auth = GithubRandomKeyAuth(logger)
 
-    process_commit_metadata(logger, key_auth, list(new_contribs), repo_id, platform_id)
+    # Process results in batches to reduce memory usage
+    batch = []
+    BATCH_SIZE = 1000
```
Ulincsys, line 269: `BATCH_SIZE` should ideally be a configurable parameter instead of hardcoded here.

Ulincsys (commented):
Looks OK to me, with a few comments

ABrain7710 (approved):

augur/tasks/git/facade_tasks.py
shlokgilda: Gotcha!

MoralCode:
rebased, fixed the merge conflict with the string fields fix (#3434) and corrected my pylint bug

Ulincsys (approved):

PR #3431, opened 2025-11-25
fix: Update log message to show configurable mat_views_interval

### LINKED ISSUE #3360: Log message incorrectly referenced a set schedule

manojalluri:
The log message incorrectly stated that the materialized view refresh was scheduled "every night at 1am CDT" when the interval is actually configurable via the refresh_materialized_views_interval_in_days config option.

This change updates the log message to dynamically display the configured interval value (mat_views_interval) instead of the hardcoded time.

Fixes #3360

**Description**

This PR fixes #

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/init/celery_app.py
+++ b/augur/tasks/init/celery_app.py
@@ -242,7 +242,7 @@ def setup_periodic_tasks(sender, **kwargs):
 
         mat_views_interval = int(config.get_value('Celery', 'refresh_materialized_views_interval_in_days'))
         if mat_views_interval > 0: 
-            logger.info(f"Scheduling refresh materialized view every night at 1am CDT")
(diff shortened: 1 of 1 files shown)


MoralCode:
Closing as there are multiple parallel solutions and the one proposed in #3370 is more robust and complete in my opinion

PR #3254, opened 2025-08-28
feat: Complete Topic Modeling System with NMF and Advanced Features

xiaoha-cloud:
**Description**
- This PR implements a comprehensive topic modeling versioning system for Augur, replacing LDA with NMF (Non-negative Matrix Factorization) and adding complete model lifecycle management, optimization, comparison, and export capabilities.

This PR fixes #

**Notes for Reviewers**
1. Database Schema Changes:
- Migration 34: Creates topic_model_meta table with 21 fields (all NOT NULL except repo_id, visualization_data)
- Migration 35: Creates topic_model_event table for audit logging
- Comprehensive model versioning with parameters_hash and data_fingerprint for duplicate detection
image
2. Core Algorithm Changes:
- Replaced LDA with NMF (Non-negative Matrix Factorization) in clustering worker
- Added TF-IDF + CountVectorizer preprocessing pipeline
- Implemented NMF interpretability scoring and Gensim coherence calculation
- Smart parameter adjustment for small datasets with fallback logic
Pasted Graphic 2

3. API Endpoints:
- POST /topic-models/{repo_id}/train - Train new models
- GET /topic-models/{repo_id}/status - Check training status
- POST /topic-models/{repo_id}/optimize - Data-driven parameter optimization
- GET /topic-models/{repo_id}/compare - Compare model performance
- GET /topic-models/{repo_id}/visualization/{model_id} - Export visualization data
4. Web Interface:
- Complete topic models listing page with model metrics
- Detailed model view with NMF quality scores and topic distributions
- Interactive model comparison with coherence score analysis
- Model export functionality with JSON download
- ECharts visualizations for topic distributions and word clouds

**Signed commits**
- [x] Yes, I signed my commits.

PR #3432, opened 2025-11-26
feat: Add topic modeling worker with NMF and REPLACE strategy

xiaoha-cloud:
**Description**
- This PR implements a complete topic modeling system split into worker core and frontend components
- Implements message-level topic modeling using NMF with REPLACE strategy to prevent data growth
- Adds 2 frontend routes and 7 API endpoints with proper permission control

Worker Core:
- Implement message-level topic modeling using NMF (Non-negative Matrix Factorization) with CountVectorizer
- Add intelligent retraining system based on 4 dimensions (Age, Params, Quality, Data growth)
- Implement REPLACE strategy to prevent data growth in repo_topic and repo_cluster_messages tables (referenced from insight_worker pattern)
- Convert raw SQL inserts to ORM bulk_save_objects for better maintainability
- Store complete vocabulary in visualization_data JSONB field
- Add model versioning, comparison, and event logging modules

Frontend:
- Add 2 frontend routes and 7 API endpoints for topic modeling
- Implement @login_required for read operations and @admin_only for write operations (train/optimize)
- Add topic models list and detail page templates
- Fix dashboard_view to use AugurConfig instead of undefined requestJson

Configuration:
- Add config.json.example template with environment variable support
- Fallback to database configuration if config file not present

The REPLACE strategy deletes old records matching repo_id, tool_source, and tool_version before inserting new data, preventing accumulation of historical topic modeling results.

This PR fixes #

**Notes for Reviewers**

### REPLACE Strategy Implementation
To solve the data growth issue, I referenced the REPLACE strategy from `insight_worker` and applied it to:
- `repo_topic` table (tasks.py lines 1111-1119)
- `repo_cluster_messages` table (tasks.py lines 1149-1157)

The strategy deletes old records matching `repo_id`, `tool_source`, and `tool_version` before inserting new data.

**Signed commits**

### DIFF
--- a/augur/api/view/api.py
+++ b/augur/api/view/api.py
@@ -13,7 +13,7 @@ from augur.tasks.frontend import (
 )
 
 from ..server import app
-from .utils import *
+from .utils import render_module, renderRepos, render_message, toCacheFilename, getSetting
 from .init import report_requests
 
 logger = logging.getLogger(__name__)
--- a/augur/api/view/routes.py
+++ b/augur/api/view/routes.py
@@ -5,17 +5,19 @@ import logging
 import math
 from functools import wraps
 from flask import render_template, request, redirect, url_for, session, flash, jsonify
-from .utils import *
 from flask_login import login_user, logout_user, current_user, login_required
 
 from augur.application.db.models import User, Repo, ClientApplication
 from .server import LoginException
-from augur.application.util import *
+from augur.application.util import get_all_repos, get_all_repos_count
 from augur.application.db.lib import get_value
 from augur.application.config import AugurConfig
 from augur.application.db.session import DatabaseSession
 from ..server import app, db_session
 from datetime import datetime
+
+# Import utility functions from local utils module
+from .utils import getSetting, render_module, renderRepos, render_message
 
 logger = logging.getLogger(__name__)
 
@@ -78,7 +80,7 @@ def repo_table_view():
     query = request.args.get('q')
     try:
         page = int(request.args.get('p') or 0)
-    except:
+    except (ValueError, TypeError):
         page = 1
 
     sorting = request.args.get('s')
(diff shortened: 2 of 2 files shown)


MoralCode:
Is this essentially #3214 but rebased and without the schemas that were already merged?

xiaoha-cloud:
Actually, this is the worker/core part of #3254, rebased onto the latest main and without the schema changes that were already merged. I intentionally split the original big PR into smaller.My plan is to follow up with additional small PRs for the remaining pieces, so they can be reviewed and discussed step by step.

augur/api/view/api.py
```diff
@@ -1,11 +1,10 @@
 import logging
 import re
 
-from flask import flash, current_app, jsonify, redirect, request, url_for
+from flask import flash, jsonify, redirect, request, url_for
 from flask_login import current_user, login_required
 
-from augur.application.db.models import Repo, RepoGroup, UserGroup, UserRepo
-from augur.application.db.session import DatabaseSession
```
sgoggins, line 8: Does this really not require database connection? I question that. @MoralCode ?

sgoggins (approved):
I think this is ready. I assume you've tested it locally @xiaoha-cloud ?

augur/api/view/api.py
MoralCode: DatabaseSession is not the only place in our code that can access the database, however it is the one that (IMO) everything should be using if at all possible.

I havent looked through this code, but removing this import from api.py does make me think something has gone wrong too.

xiaoha-cloud: In main, `DatabaseSession` is also imported but unused: the routes never call with `DatabaseSession(...)`, and `add_existing_org_to_group` isn’t invoked. All DB access still flows through the `current_user.*` and `UserGroup/UserRepo` helpers, so removing the unused import was just a lint clean-up and doesn’t change behavior.

xiaoha-cloud: If we need to follow the “explicit outer session” convention, we can re-import from `augur.application.db.session import DatabaseSession` and wrap the relevant routes in `with DatabaseSession(logger, engine=app.engine) as db`: and pass that session into the calls. As of now, behavior matches `main`, just without the unused-import warning.

PR #3438, opened 2025-12-03
specify `pr_review_body` as a User generated content string field

MoralCode:
**Description**
Specifies that user generated content may be contained in the PR review body so it can be filtered for null values 

This PR fixes #3434 

**Notes for Reviewers**

**Signed commits**
- [X] Yes, I signed my commits.

PR #3439, opened 2025-12-04
[tasks/github] Batch processing for PR reviews collection

shlokgilda:
**Description**
- Adds batched processing to `collect_pull_request_reviews` to reduce memory usage
- Processes reviews in batches of 1000 instead of accumulating all in memory before insertion
- Combines contributor and review extraction into a single pass (was two separate loops)
- Removes redundant `repo_id` database lookup
- Fixes incorrect argument order in `extract_needed_pr_review_data` call (was storing wrong metadata values)

**Notes for Reviewers**
- The original code had a subtle bug: it passed `tool_source, tool_version` where `tool_version, data_source` was expected. This didn't crash because `tool_source` is hardcoded inside the function, but it stored `"pull_request_reviews_reviews"` as   `tool_version` and `"2.0"` as `data_source` instead of the correct values.
- Memory impact: For a repo with 5000 PRs averaging 3 reviews each, old code held ~15,000 review objects in memory. New code caps at ~1000.
- Follows the same batching pattern already used in `collect_pull_requests`.

**Testing**
- Testing this code with large repos with around 20-30K PRs each. I can see the logs and lower memory usage for workers working on  `collect_pull_request_reviews`. Will post testing related updates in comments  if any.

**Signed commits**

**AI Disclosure**: I used Claude Code to generate docstrings and write this PR draft.

### DIFF
--- a/augur/tasks/github/pull_requests/tasks.py
+++ b/augur/tasks/github/pull_requests/tasks.py
@@ -11,12 +11,12 @@ from augur.tasks.github.util.util import add_key_value_pair_to_dicts, get_owner_
 from augur.application.db.models import PullRequest, Message, PullRequestReview, PullRequestLabel, PullRequestReviewer, PullRequestMeta, PullRequestAssignee, PullRequestReviewMessageRef, Contributor, Repo
 from augur.tasks.github.util.github_task_session import GithubTaskManifest
 from augur.tasks.github.util.github_random_key_auth import GithubRandomKeyAuth
-from augur.application.db.lib import get_session, get_repo_by_repo_git, bulk_insert_dicts, get_pull_request_reviews_by_repo_id, batch_insert_contributors
+from augur.application.db.lib import get_repo_by_repo_git, bulk_insert_dicts, get_pull_request_reviews_by_repo_id, batch_insert_contributors
 from augur.application.db.util import execute_session_query
 from ..messages import process_github_comment_contributors
 from augur.application.db.lib import get_secondary_data_last_collected, get_updated_prs, get_core_data_last_collected
 
-from typing import Generator, List, Dict
+from typing import List
 
 
 platform_id = 1
@@ -52,15 +52,15 @@ def collect_pull_requests(repo_git: str, full_collection: bool) -> int:
                 total_count += len(all_data)
                 all_data.clear()
 
-        if len(all_data):
+        if all_data:
             process_pull_requests(all_data, f"{owner}/{repo}: Github Pr task", repo_id, logger, augur_db)
             total_count += len(all_data)
 
         if total_count > 0:
-            return total_count
-        else:
             logger.debug(f"{owner}/{repo} has no pull requests")
             return 0
+
+        return total_count
         
         
     
@@ -180,30 +180,6 @@ def process_pull_requests(pull_requests, task_name, repo_id, logger, augur_db):
     pr_metadata_string_fields = ["pr_src_meta_label"]
     augur_db.insert_data(pr_metadata_dicts, PullRequestMeta,
                         pr_metadata_natural_keys, string_fields=pr_metadata_string_fields)
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
 
 
 def process_pull_request_review_contributor(pr_review: dict, tool_source: str, tool_version: str, data_source: str):
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -121,7 +121,7 @@ def trim_commits_post_analysis_facade_task(repo_git):
     repo_loc = (f"{absolute_path}/.git")
     # Grab the parents of HEAD
 
-    parent_commits = get_parent_commits_set(repo_loc)
+    parent_commits = get_parent_commits_set(repo_loc, facade_helper)
 
     # Grab the existing commits from the database
     existing_commits = get_existing_commits_set(repo_id)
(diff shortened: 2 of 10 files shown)


augur/tasks/github/pull_requests/tasks.py
```diff
@@ -338,7 +389,6 @@ def collect_pull_request_reviews(repo_git: str, full_collection: bool) -> None:
     tool_source = "pull_request_reviews"
     data_source = "Github API"
 
-    repo_id = get_repo_by_repo_git(repo_git).repo_id
```
MoralCode, line 317: why was this removed?

augur/tasks/github/pull_requests/tasks.py
```diff
@@ -327,9 +327,60 @@ def collect_pull_request_review_comments(repo_git: str, full_collection: bool) -
...
+    if contributors:
+        # Remove duplicates within the batch before inserting
+        unique_contributors = remove_duplicate_dicts(contributors)
+        logger.info(f"{owner}/{repo} Pr reviews: Inserting {len(unique_contributors)} contributors")
+        augur_db.insert_data(unique_contributors, Contributor, ["cntrb_id"])
+
+    if pr_reviews:
+        logger.info(f"{owner}/{repo}: Inserting {len(pr_reviews)} pr reviews")
+        pr_review_natural_keys = ["pr_review_src_id"]
+        augur_db.insert_data(pr_reviews, PullRequestReview, pr_review_natural_keys)
```
MoralCode: This is probably going to need adjusting or merge-deconflicting to account for the fix in #3438 depending on which merges first

augur/tasks/github/pull_requests/tasks.py
shlokgilda: This was a duplicate call. repo_id is being reset on line 397 later.

augur/tasks/github/pull_requests/tasks.py
shlokgilda: Yeah, I agree.

shlokgilda: If we know the order of PRs in which they will be processed, I can rebase older PRs on this.

MoralCode: i suspect the other one is more *likely* to get in first, just because its smaller/more focused.

if youre able to replicate the underlying issue (i think the github.com/ClassclockAPI repo was the one that triggered that other issue, but theoretically anything with null characters in a PR review should do it) and help test the fix that would be super though!

shlokgilda: Okay! I can try to mock up a quick test for the other PR I guess.

shlokgilda:
Testing update: Did not encounter any errors. Table values correctly updated. Tested with `https://github.com/tensorflow/tensorflow` (large repo; ~75K PRs).

augur/tasks/github/pull_requests/tasks.py
sgoggins: I think that would be very helpful @shlokgilda

sgoggins:
@shlokgilda : I merged a PR that fixed a seperate issue on PR reviews from @MoralCode just now ... which created a conflict with your code. I think it looks pretty straightforward as both changes add safety. While I am nearly certain I know how to resolve the conflict while incorporating both changes, I would like @shlokgilda to change it in their fork/branch and update the PR ... So I don't mess something up editing online.

shlokgilda:
@sgoggins Fixed and pushed.

augur/tasks/github/pull_requests/tasks.py
```diff
@@ -327,9 +327,61 @@ def collect_pull_request_review_comments(repo_git: str, full_collection: bool) -
...
+        repo: Repository name (for log messages).
+
+    Returns:
+        None. Lists are NOT cleared by this function - caller must clear them.
+    """
+    if contributors:
+        # Remove duplicates within the batch before inserting
+        unique_contributors = remove_duplicate_dicts(contributors)
+        logger.info(f"{owner}/{repo} Pr reviews: Inserting {len(unique_contributors)} contributors")
+        augur_db.insert_data(unique_contributors, Contributor, ["cntrb_id"])
```
MoralCode, line 329: Does this move to procesing contributors within a batch, rather than globally, affect how contributors are processed?

i.e. if its resolving or deduplicating contributors only within a batch, would that mean its possible that something gets mis-resolved or a duplicate gets missed if the values are in separate batches?

shlokgilda: No, this won't cause any issues with contributors being missed across batches. The `remove_duplicate_dicts()` call is just an optimization to avoid sending redundant INSERT statements within a single batch. The actual cross-batch deduplication happens at the database level - `insert_data()` uses PostgreSQL's `ON CONFLICT DO UPDATE` (upsert) with `cntrb_id` as the natural key:

```python
augur_db.insert_data(unique_contributors, Contributor, ["cntrb_id"])
```
So if contributor "alice" shows up in batch 1, she gets inserted. If she appears again in batch 3, the `ON CONFLICT` clause updates the existing row instead of creating a duplicate. The batch-local dedup just saves us from sending multiple identical INSERTs in the same DB call.

sgoggins:
@shlokgilda : We will merge this as soon as the conflict is resolved! Thanks!

shlokgilda:
@sgoggins Merge conflict resolved.

MoralCode (approved):
My comments were addressed. I trust that this has been tested by the author and the code looks reasonable and well documented

PR #3444, opened 2025-12-06
Fix: Eliminate redundant full table scans in messages and events collection

### LINKED ISSUE #3440: Full table scans on every batch in messages and events collection
**Description:**
Both `messages.py` and `events.py` rebuild issue/PR URL mappings on every batch instead of once at the start. This causes redundant full table scans that scale with data volume.

**messages.py:156-164** - queries all issues and PRs per 20-message batch:
```python
issues = augur_db.session.query(Issue).filter(Issue.repo_id == repo_id).all()
prs = augur_db.session.query(PullRequest).filter(PullRequest.repo_id == repo_id).all()
```
**events.py:237-253** - rebuilds mappings per 500-event batch via `_get_map_from_*` methods.

Impact:
- 1000 messages -> 50 full scans of issues AND PRs tables
- 10000 events -> 40 full scans total

Expected behavior:
Build mappings once before the batch loop, pass as parameters. See #3439 for the expected pattern.

Suggested fix:
1. Move mapping queries outside batch loops
2. Build `issue_url_to_id_map` and `pr_url_to_id_map` once
3. Pass mappings to processing functions
4. Possibly also increase messages batch size from 20 to 1000 (unless this was an intentional design choice that wasn't documented in the code)

Files:
- augur/tasks/github/messages.py
- augur/tasks/github/events.py

PredictiveManish:
**Description**

Moved mapping queries outside batch loops and pass pre-built mappings as parameters to processing functions, following the pattern established in #3439. 
Solves #3440 
#### `augur/tasks/github/messages.py`
- Built `issue_url_to_id_map` and `pr_issue_url_to_id_map` once in `collect_github_messages()` before any batch processing
- Updated `process_messages()` to accept mappings as parameters instead of rebuilding them
- Updated `process_large_issue_and_pr_message_collection()` to accept and pass mappings
- Increased batch size from 20 to 1000 (reduces batch overhead)

#### `augur/tasks/github/events.py`
- Built `issue_url_to_id_map` and `pr_url_to_id_map` once in `BulkGithubEventCollection.collect()` before the batch loop
- Updated `_process_events()`, `_process_issue_events()`, and `_process_pr_events()` to accept mappings as parameters
- Removed redundant `_get_map_from_*()` calls from batch processing methods

## Performance Improvement

- Before: 1,000 messages -> 50 full scans of issues AND PRs tables
- After: 1,000 messages -> 1 full scan of each table (50x reduction)

- Before: 10,000 events -> 40 full scans total
- After: 10,000 events → 1 full scan of each table (40x reduction)

This PR fixes #3440 

**Notes for Reviewers**

**Signed commits**


augur/tasks/github/messages.py
```diff
@@ -124,17 +137,17 @@ def process_large_issue_and_pr_message_collection(repo_id, repo_git: str, logger
             logger.info(f"{task_name}: PR or issue comment url of {comment_url} returned 404. Skipping.")
             skipped_urls += 1
        
-        if len(all_data) >= 20:
-            process_messages(all_data, task_name, repo_id, logger, augur_db)
+        if len(all_data) >= 1000:
```
sgoggins: Changing this all the way to 1,000 is going to potentially generate more contention and locking; but not if teh full table scans are tamped down independently of this change.

MoralCode: if we are going to bump this, and it was set this low to begin with, id say raising it to 200 would be a good first step. thats 10x more records, so 10x less batch overhead without risking the batches getting too big

shlokgilda:
I just have non-blocking code quality suggestion: use a named constant like `MESSAGE_BATCH_SIZE = 500` instead of a magic number

augur/tasks/github/messages.py
PredictiveManish: updated it for 200.

MoralCode:
> I just have non-blocking code quality suggestion: use a named constant like `MESSAGE_BATCH_SIZE = 500` instead of a magic number

Yep, agreed

PredictiveManish:
> I just have non-blocking code quality suggestion: use a named constant like `MESSAGE_BATCH_SIZE = 500` instead of a magic number

What to give the size 500 or 200? as @MoralCode suggested for 200

sgoggins:
> > I just have non-blocking code quality suggestion: use a named constant like `MESSAGE_BATCH_SIZE = 500` instead of a magic number
> 
> What to give the size 500 or 200? as @MoralCode suggested for 200

200

sgoggins:
@MoralCode / @PredictiveManish : I'm rerunning the failed end to end test. Sometimes GitHub gets overwhelmed and they just timeout.

augur/tasks/github/messages.py
```diff
@@ -14,7 +14,7 @@
 from sqlalchemy.sql import text
 
 platform_id = 1
-
+MESSAGE_BATCH_SIZE = 200
```
sgoggins, line 19: @MoralCode / @shlokgilda : I think this amounts to all the requested changes I see.

shlokgilda: Sounds about right!

MoralCode:
> May we assume you ran this locally and collected data?

Lets not assume - would rather accidentally over-test than not test at all.

augur/tasks/github/messages.py
```diff
@@ -37,17 +37,30 @@ def collect_github_messages(repo_git: str, full_collection: bool) -> None:
...
+        issues = augur_db.session.query(Issue).filter(Issue.repo_id == repo_id).all()
+        for issue in issues:
+            issue_url_to_id_map[issue.issue_url] = issue.issue_id
+
+        # create mapping from pr url to pr id of current pull requests
+        pr_issue_url_to_id_map = {}
+        prs = augur_db.session.query(PullRequest).filter(PullRequest.repo_id == repo_id).all()
+        for pr in prs:
+            pr_issue_url_to_id_map[pr.pr_issue_url] = pr.pull_request_id
+
```
MoralCode, line 54: Above you rely on some functions for these mappings, Is this code just a duplicate of the code in those functions? If those functions are useful elsewhere and not tied in anywhere maybe they can be useful utility functions that can be imported here too?

Overall id recommend splitting this PR so that the changes to events.py can be merged without being held up by the larger question of refactoring ( ref #3345) that this file brings up

sgoggins: @PredictiveManish : I agree with @MoralCode that splitting the events.py from the other refactoring would make this more straightforward to merge and test.

PredictiveManish: Covered changes in `events.py` in #3479

MoralCode: sounds good, will focus on that PR first. This one is still on hold until we figure out how best to share the same function that was proposed in events.py

PredictiveManish: yeah, for events.py I have created new PR as suggested for split.

PR #3449, opened 2025-12-09
 [tasks/github] Batch processing for PR review comments collection

shlokgilda:
**Description**
- Adds batched processing to `collect_pull_request_review_comments` to reduce memory usage
- Processes comments in batches of 1000 instead of accumulating all in memory before insertion
- Combines contributor and comment extraction into a single pass (was two separate loops)
- Extracts shared `_flush_contributors` helper function for code reuse between PR reviews and PR review comments
- Adds defensive batch trigger that checks both `pr_review_comment_dicts` and `contributors` list sizes to prevent unbounded memory growth in edge cases

**Dependencies**
- **This PR should be merged after #3439** as it builds on that branch

**Notes for Reviewers**
- Memory impact: For a repo with many PR review comments, old code loaded all comments into memory via
`list(github_data_access.paginate_resource(...))`. New code streams from the generator and caps batches at ~1000.
- Follows the same batching pattern used in `collect_pull_request_reviews` from #3439
- The `_flush_contributors` helper is now shared between both PR reviews and PR review comments flush functions for consistency

**Testing**
- Tested this code with few larger repos (>50K PRs/issues). Works fine. 

**Signed commits**

**AI Disclosure**: I used Claude Code to write this PR draft and generate docstrings.


MoralCode:
converting to draft because of 
> This PR should be merged after https://github.com/chaoss/augur/pull/3439 as it builds on that branch

sgoggins:
@MoralCode  : I really appreciate the safety move of switching this to draft so somebody (me) doesn't merge these in the wrong order. :)

augur/tasks/github/pull_requests/tasks.py
```diff
@@ -231,76 +248,72 @@ def collect_pull_request_review_comments(repo_git: str, full_collection: bool) -
...
+    total_refs_inserted = 0
 
-    pr_review_comments_len = len(all_raw_pr_review_messages)
-    for comment in all_raw_pr_review_messages:
-
-        # pull_request_review_id is required to map it to the correct pr review
-        if not comment["pull_request_review_id"]:
-            continue
+    # Single-pass extraction: get both contributor and comment data together
+    for comment in github_data_access.paginate_resource(review_msg_url):
```
MoralCode, line 259: does gathering them together make things more likely to fail if there is an error in gathering one of them?

MoralCode (commented):
Would love to know how this has been tested. I can usually never get python dict comprehension to work right on the first try

shlokgilda:
I ran this against tensorflow/tensorflow (75K+ PRs) and a few other large repos. The few things I validated:
1. No errors during collection (workers completed successfully)
2. Database values matched expected counts (compared PR count vs reviews inserted)
3. Memory usage stayed stable (no gradual climb like before)

The dict comprehension part is actually pretty straightforward - it's just the two-pass loop collapsed into one. Instead of looping through reviews twice (once for contributors, once for review data), we do both extractions in the same pass.

augur/tasks/github/pull_requests/tasks.py
shlokgilda: If contributor extraction fails for a review, `process_pull_request_review_contributor()` returns `None`, so the `if contributor:` check prevents adding it. Then the review data extraction only happens if `"cntrb_id" in review`, so if the contributor failed to process, the review gets skipped too.

Isn't this the correct behavior? We shouldn't insert a review without its contributor anyway. In the old code, if contributor processing silently failed but review processing succeeded, you'd get orphaned review records. Or was that the expected behavior?

MoralCode: no, i don't think orphaned records were ever part of the intention of Augur - i think the idea is that if data is present, all of the linked records also are present.

PR #3435, opened 2025-12-02
Synchronize Database models with migrations

MoralCode:
**Description**
This PR resolves and reconciles the differences between the current, in-database data schema and the schema defined by the SQLAlchemy models. This is needed because, for the past few years, both the models and the migrations have had to be made manually, meaning a lack of perfect consistency between updates to the two sources can cause drift between them.

The recent changes in #3416 configured alembic to automatically compare the data models and the deployed schema and generate migrations that represent the differences. This PR reconciles the various differences that have accumulated over the years.

This PR is one step towards, but not yet a complete fix for #3388.

**Notes for Reviewers**
Most of the schema conflicts were resolved with relatively minor adjustments to the data models (such as specifying names for Foreign keys, adding fields to unique constraints, etc) to bring them in line with what actual deployed databases have.

There were a few places where I thought it was better to update the database instead:
- There was an index on `contributors.gh_login` that I dropped, since I believe gh_login has been superseded by another method of looking up contributors for a while now
- a unique constraint on the `pull_request_reviews` table was updated to add `tool_source` to its uniqueness check.
- the `augur_operations.users.email_verified column was dropped` since a recent maintainers call confirmed that email verification was a feature that was previously being explored as part of a password reset feature, but was ultimately dropped due to having limited utility (wasnt worth the the additional overhead of configuring sendgrid for email processing compared to a hypothetical future implementation that uses the Augur CLI to enable admins to perform password resets)

Please look over all the changes and let me know if you think any of these changes needs to be reconciled in a different way (i.e. a model change is incorrect and the database needs to change instead)

I have yet to run this on an actual augur install but plan to soon.

This change is a pretty severe blocker for other database changes I would like to make since I want to start from a relatively clean/consistent state when proposing additional migrations (including some that are important for the next release)

**Signed commits**
- [X] Yes, I signed my commits.

### Gen AI Disclosure
Generative AI was only used for the planning stages of this changes - for example, looking at the schemas or alembic-generated migrations and helping summarize what the differences were between the models and the database. All of these changes were understood/verified and implemented by me.

PR #3450, opened 2025-12-09
Consistent SA model setup

MoralCode:
**Description**
- This change is a simple stylistic refactoring change that adjusts how our SQLAlchemy models are defined so that it follows consistent rules

These rules are:
- additional metadata parameters in models specifying constraints, tablename, or schema) appear at the top of the class definition
- these metadata definitions should use the simplest syntax available given what they need to do (i.e. if a model only needs a schema set, it can be defined as a plain dict, but if its a schema setting and constraints, it needs to be a tuple with the dict at the end)

This PR is intended to pull this relatively simple improvement out from #3435 to make that PR more easy to review.

**Notes for Reviewers**

This should be a relatively easy merge

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -1359,7 +1359,8 @@ class Commit(Base):
 
 class CommitMessage(Base):
     __tablename__ = "commit_messages"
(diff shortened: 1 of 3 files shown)


sgoggins (approved):
LGTM.

PR #3464, opened 2025-12-12
truncate release_id values to 256 characters

paras-chinchalkar:
Added truncation logic to ensure release_id values fit within the database constraint:
Truncate release_id to 256 characters - Before inserting release data, truncate the release_id to exactly 256 characters to match the database constraint
Add warning logging - When truncation occurs, log a warning message with details about the original length and the release name/tag being processed
Handle both release types - Applied truncation logic to both regular GitHub releases and tag-only releases
Backward compatible - Added optional logger parameter to maintain backward compatibility
Changes Made
Modified augur/tasks/github/releases/core.py:
Updated get_release_inf() function signature to accept optional logger parameter
Added truncation logic for release_id in both regular and tag-only release paths
Added warning logging when truncation occurs
Updated insert_release() to pass logger to get_release_inf()

### DIFF
--- a/augur/tasks/github/releases/core.py
+++ b/augur/tasks/github/releases/core.py
@@ -7,7 +7,7 @@ from augur.application.db.util import execute_session_query
 from augur.application.db.lib import bulk_insert_dicts
 
 
-def get_release_inf(repo_id, release, tag_only):
+def get_release_inf(repo_id, release, tag_only, logger=None):
     if not tag_only:
 
         if release['author'] is None:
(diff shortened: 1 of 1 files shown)


MoralCode:
Thanks for your contribution! In order to properly prioritize and review this, please make sure you are submitting your contributions with descriptive titles that explain what the fix is. While you can mention the issue that you are fixing in the title, a better way would be to include text like "fixes#<number>" in the body of your PR message so that github can link the issue being solved automatically.

It would also be helpful to learn a little more about why you believe these changes are a good solution to the problem, or what led you to these solutions in the first place. It's okay to summarize your code changes in the PR description, but without context, its hard for other maintainers to see why this solution is better than other options that may be available.

MoralCode:
I also notice that this PR also contains the changes from https://github.com/chaoss/augur/pull/3463. If multiple sets of changes are included in a pull request, it makes the code a lot harder to review as the maintainers need to distinguish between which code is related to what changes and are less able to consider each change independently.

MoralCode:
Augur is a data storage/collection tool. I think intentionally losing data, such as by adding truncation logic to the app, especially when we could just as easily increase the character limit for release names, is not the correct path.

Closing this issue

PR #3470, opened 2025-12-15
Skip non-public GitHub events to avoid KeyError #3468

### LINKED ISSUE #3468: KeyError: 'url' in contributor_breadth_worker when processing GitHub events
While running Augur locally, the contributor_breadth_worker crashes with a KeyError: 'url' during GitHub event processing.

The error occurs when the worker tries to access event['repo']['url'] inside process_contributor_events. In some cases, the GitHub events payload does not include this key, which causes the worker to raise an exception

Image

repo url is not guaranteed to be present in all events. Treating it as mandatory in  contributor_breath_modeL is what causes the crash
ie   if we forked from a repo  later on the repo is made private by owner   and deleted by owner then we get this url keyerror

iGufrankhan:
**Description**
This PR fixes #3468 by defensively skipping non-public GitHub events in the
`contributor_breadth_worker`.

Some GitHub Events API responses (notably private fork events) do not include
repository metadata (e.g. `repo.url`), which causes the worker to raise a
`KeyError` and crash. Skipping non-public events avoids this failure and
aligns behavior with data generally available via the public GitHub API.

**Notes for Reviewers**
- The change is intentionally minimal and defensive.
- Only non-public events are skipped; no existing public event behavior
  is modified.
- This avoids relying on private repository data, which most Augur
  deployments cannot reliably access.

**Signed commits**

### DIFF
--- a/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
+++ b/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
@@ -129,6 +129,8 @@ def process_contributor_events(cntrb, cntrb_events, logger, tool_source, tool_ve
 
     cntrb_repos_insert = []
     for event_id_api in cntrb_events:
+        if not event_id_api.get('public', True):
+            continue    
 
         cntrb_repos_insert.append({
             "cntrb_id": cntrb['cntrb_id'],


MoralCode:
@iGufrankhan It seems like you are using a main branch from a few months ago as the base for your PR - can you update your fork's main branch to the current main, create/move your branch for this PR to the latest `main` commit, and then cherry-pick your one commit containing the fix (i.e. effectively rebasing your branch onto current main)? 

You're also going to want to make sure your commits are signed-off

iGufrankhan:
@MoralCode I’ve updated my fork’s main to the latest upstream and cherry-picked the fix into a clean branch. Is it okay if I close this PR and open a new one with the rebased commit?

MoralCode:
Thats one option, and probably the safest.

Alternatively, if you rename your old branch locally and use the same branch name for your new branch and force push it (careful with this), you can also update this existing PR. But thats probably the more complicated and risky way if you aren't used to doing that kind of thing in git

iGufrankhan:
I’m going to go with the first option and close this PR, then open a fresh one based on the latest upstream/main.
I want to keep the history clean and avoid any mistakes with git flow.
I’ll link the new PR here once it’s open.

PR #3471, opened 2025-12-15
Add actual SQLAlchemy Sequence objects so db can be initialized with SQLAlchemy

MoralCode:
**Description**
This was initially discovered while trying to fix the test fixtures to get the database-dependent tests back up and running. It was something that prevented databases created purely with SQLAlchemy (rather than replaying every migration) from successfully coming up.

Converting these values to use SQLAlchemy allows SQLAlchemy to emit the correct SQL to generate these sequences in the database again.

This was originally part of #3435 but is being factored out to keep the diff of that change small.

This change incorporates feedback from @shlokgilda in the above-mentioned PR that says that SQLAlchemy  sequences don't inherit the table's schema automatically

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -185,11 +185,7 @@ class Contributor(Base):
 
     cntrb_id = Column(
         UUID(as_uuid=True),
-        Sequence('contributors_cntrb_id_seq', start=25430, schema='augur_data'),
         primary_key=True,
-        server_default=text(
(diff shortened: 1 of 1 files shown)


sgoggins:
@MoralCode : Flag this one for me when you think its ready.

MoralCode:
rebased to address a minor merge conflict with #3435 (the release_id) changed in length.

Merging since this is functionally identical code to what was approved previously.

PR #3473, opened 2025-12-16
Skip non-public GitHub events in contributor_breadth_worker avoid keyerror #3468

iGufrankhan:
Description
This PR fixes https://github.com/chaoss/augur/issues/3468 by defensively skipping non-public GitHub events in the
contributor_breadth_worker.

Some GitHub Events API responses (notably private fork events) do not include
repository metadata (e.g. repo.url), which causes the worker to raise a
KeyError and crash. Skipping non-public events avoids this failure and
aligns behavior with data generally available via the public GitHub API.

Notes for Reviewers

The change is intentionally minimal and defensive.
Only non-public events are skipped; no existing public event behavior
is modified.
This avoids relying on private repository data, which most Augur
deployments cannot reliably access.
Signed commits

 Yes, I signed my commits.

### DIFF
--- a/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
+++ b/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
@@ -129,6 +129,8 @@ def process_contributor_events(cntrb, cntrb_events, logger, tool_source, tool_ve
 
     cntrb_repos_insert = []
     for event_id_api in cntrb_events:
+        if not event_id_api.get('public', True):
+            continue    
 
         cntrb_repos_insert.append({
             "cntrb_id": cntrb['cntrb_id'],


augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
```diff
@@ -129,6 +129,8 @@ def process_contributor_events(cntrb, cntrb_events, logger, tool_source, tool_ve
 
     cntrb_repos_insert = []
     for event_id_api in cntrb_events:
+        if not event_id_api.get('public', True):
```
MoralCode, line 132: within event_id_api, is the value stored as a string or boolean?

MoralCode (commented):
very small focused PR, nice!

augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
iGufrankhan: image


it is a boolean value  as you can see that public =true --> no quotes

MoralCode (commented):
LGTM! Could use more details on testing though.

Has this been running on anyones instance? can anyone intentionally reproduce the error that this fixes before using this version and not after?

iGufrankhan:
@MoralCode 


actually   
i think this is not gone fix the issue

beacuse this bug genarlly create when i forked   a repo it create a forkevents 

but same time origind repo owner made the repo priavte and also the i myslef my forked repo private then this  error is genaretd 


    
when i try to reproduce the error it occur 

   lets suppose 
   i forked some repo later the owner of the repo make it private or delete it , but event already create thoislead lead to repo empty   and issue is create for us ,
   
   this is not related or to private or public   or actually about mising of repo url in repo section 
   
   this is my understainding  what to think 
   
   i think of defensive line like repo or repo url is mising then it crash our model    as far i know that the github events is not stable 
   in term of events type and visiblitity 
   
   what do you think about this should i more investigate about it

iGufrankhan:
@MoralCode  
i check my self to reprodce the issue and try to fix it so that 
see that 



Screenshot 2025-12-21 115741



this genarte the issue not fixed

so  my solution with proof is like 


think this most accurate defensive line for this issue is so that it can help in future aspect also 

  repo =event_id_api.get('repo')
        if not repo or 'url' not in repo:
               continue
               


     this line of code is more accurate it fix if repo is not present (just in case) and if url is not present it doesnot crash our model



Screenshot 2025-12-21 120319






this is my understanding so far let me know what do you think

MoralCode:
I think the answers to these questions will narrow down the issue:

1. do you have a second github account you can test with? if so, can you try making the fork with a different account than the account that owns the API key augur is using?
2. Does the original base repo HAVE to be made private for this bug to be reproduced? or can it be done with any fork repo where the original is inaccessible (such as if it is deleted)?

iGufrankhan:
image


“this when i forked from my friend and friend made the repo private”


i think repo object is empty when the base repository becomes inaccessible after the event was created.


also it happen sometime but i think this is a issue so we have to add a defensive code like if repo is empty or url is not present the continue the process by skipping that  without any breaking of process

MoralCode:
Waiting for feedback from sean in the underlying issue about how to resolve this missing data

sgoggins:
@iGufrankhan : This issue seems reasonable: GH is protecting disclosure of private repository events, and there's an edge case where if I fork a public repo then make it private, there is a failure. I think I understand that the mode of failure is the non-inclusion of the repository URL ... Is it the API breaking (i.e., returning an error) for these events, or is Augur breaking because its not getting all the data it thinks it needs? 

I realize its a subtle distinction. If the API is generating an error, this fix makes good sense. I am a little more skeptical about the fix if Augur is choking on the data returned. That condition *smells* more like something we probably want to adjust in how we store data with that URL missing. 

What do you think?

iGufrankhan:
@sgoggins @MoralCode 
yes ,this is a the distinction really matters here
From my testing and attempts to reproduce the issue, the GitHub Events API is not returning an error in these cases. The API response is successful, but the payload is incomplete: the event object is present, yet the repo object is either empty or missing the url field entirely.

these are edges cases like 
1. a fork event created while the base repo was public, and  later became private or somehow deleted 

In that situation, GitHub appears to intentionally redact repository metadata, but still returns the event itself. Augur then fails because it assumes event["repo"]["url"] is always present, leading to the KeyError.


I agree with your point that this could also be handled at the data-model/storage level by allowing
them repository-less events to be stored   but 
i wanted to confirm one thing that   wihout url or id ,is this events valid ??

i think  for  the  Contributor_breadth_worker specifically, the repo URL is a core requirement for meaningful processing right??

that why my approach is like 
since the augur is not access for the private events so that  skipping events that lack repo.url seems consistent with the intent of this worker and avoids crashing the pipeline.

we use defensive line like this :

repo = event.get("repo")
if not repo or "url" not in repo:
    continue
    
This preserves existing behavior for all public events, prevents worker crashes, and keeps us aligned with the reality that GitHub event payloads are not fully stable across visibility changes

I may be missing some broader architectural context since I’m still getting familiar with the Augur codebase, but this is my current understanding of the issue. I’m happy to adjust the approach if storing and handling partial repository data is preferred longer-term.

iGufrankhan:
Closing in favor of a new PR with the updated solution.

MoralCode:
#3730 (linking the new PR for convenience)

PR #3491, opened 2025-12-22
docs: add WSL2 Docker-based installation guide

### LINKED ISSUE #3108: Lack of setup instructions for Augur on Windows using WSL
### Issue

There are no specific setup instructions for running Augur on Windows using WSL (Windows Subsystem for Linux). I attempted to set up Augur using WSL with Ubuntu and faced several issues due to missing or unclear steps in the documentation.

### What I Tried

- Used WSL2 with Ubuntu 22.04
- Followed the general setup guide from the README
- Faced issues during `augur backend start`, including dependency problems, unclear DB setup steps, and configuration issues.
- Had to manually troubleshoot a lot, which could’ve been avoided with a dedicated WSL section.

### Suggested Improvements

- Add a section in the documentation (README or a new file like `SETUP_WSL.md`) for WSL-based setup.
- Mention prerequisites like Docker Desktop with WSL integration, PostgreSQL handling inside/outside WSL, and `.env` config.
- Share common errors and solutions (e.g., how to handle port conflicts or PostgreSQL access).

### Environment

- OS: Windows 11
- WSL: Ubuntu 22.04 (WSL2)
- Docker Desktop (with WSL backend)
- Command that caused issues: `augur backend start`, DB connection setup

Happy to help contribute to this if needed!

HimasreeKolathur24:
**Description**
This PR adds a dedicated guide for installing and running Augur on Windows using WSL2 with Docker.
It documents the recommended Docker-based setup, required prerequisites, common issues, and practices to avoid manual installations that often lead to errors.

This PR fixes #3108

**Notes for Reviewers**
- This documentation focuses only on Docker-based setups in WSL2.
- Manual (non-Docker) installation steps are intentionally excluded.
- The guide aligns with the existing official Docker installation workflow.

**Signed commits**

### DIFF
--- a/docs/source/getting-started/wsl-setup.rst
+++ b/docs/source/getting-started/wsl-setup.rst
@@ -1,0 +1,78 @@
+Installing Augur on Windows Using WSL2 (Docker-Based)
+=====================================================
+
+Overview
+--------
+
+This guide explains how to run Augur on Windows using WSL2 with Docker.
+Manual (non-Docker) installation inside WSL is **not recommended**.
+
+Prerequisites
+-------------
+
+- Windows 10 or Windows 11
+- WSL2 enabled
+- Ubuntu 22.04
+- Docker Desktop with WSL integration enabled
+
+Recommended Setup (Docker)
+--------------------------
+
+1. Install Docker Desktop on Windows:
+
+   https://docs.docker.com/desktop/install/windows-install/
+
+2. Enable WSL integration for Ubuntu:
+
+   - Open Docker Desktop
+   - Go to **Settings → Resources → WSL Integration**
+   - Enable integration for Ubuntu
+
+3. Clone Augur inside WSL (not in the Windows filesystem):
+
+   ::
+
+      git clone https://github.com/chaoss/augur.git
+      cd augur
+
+Common Issues and Solutions
+---------------------------
+
+PostgreSQL Port Conflicts
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Stop PostgreSQL running on the Windows host.
+- Ensure only the Docker container uses port **5432**.
+
+File Permission Issues
+~~~~~~~~~~~~~~~~~~~~~~
+
+
+- Always clone the repository inside your WSL home directory.
+- Avoid using paths under ``/mnt/c``.
+
+Docker Resource Limits
+~~~~~~~~~~~~~~~~~~~~~~
+
+- Increase Docker memory allocation to at least **8 GB** in Docker Desktop.
+
+What Not To Do
+--------------
+
+- Do **not** run manual backend installations inside WSL.
+
+  Augur’s backend services are tightly coupled with the Docker environment and the
+  provided container configuration. Installing backend services manually inside WSL
+  bypasses Docker’s networking, dependency management, and service orchestration.
+  This commonly leads to version mismatches, broken service dependencies, and
+  unpredictable behavior when containers attempt to communicate with host-installed
+  services.
+
+- Do **not** run PostgreSQL separately outside Docker.
+
+  Running PostgreSQL outside Docker often causes port conflicts (especially on port
+  **5432**), data directory mismatches, and authentication issues. Augur’s Docker
+  configuration expects full control over the database lifecycle. Using an external
+  PostgreSQL instance can result in connection failures, corrupted database state, and
+  difficult-to-debug startup errors.
+
--- a/docs/source/getting-started/toc.rst
+++ b/docs/source/getting-started/toc.rst
@@ -26,6 +26,7 @@ Happy hacking!
 
    database
    installation
+   wsl-setup
    collecting-data
    frontend
    command-line-interface/toc


sgoggins:
Rerunning CI after failure that seems like a timeout issue on the GH end.

himasree9:
> Rerunning CI after failure that seems like a timeout issue on the GH end.

Thank you. Happy to help if anything is needed.

MoralCode:
Reformatting this to use the read the docs rst format would help this be seen by more people by adding it to the main docs page

himasree9:
> Reformatting this to use the read the docs rst format would help this be seen by more people by adding it to the main docs page

@MoralCode  
Thank you for the suggestion.
I’ve reformatted the WSL documentation using the Read the Docs RST format and updated the PR accordingly so it can be included in the main documentation.

himasree9:
Hi @sgoggins — I rebased the branch to clean up the commit history and address the review feedback.
All changes are now in a single commit. Thanks!

docs/install-wsl.md
MoralCode, line 1: We dont need both the RST and md versions. Can you remove the md one and make sure all its content is represented in the rst?

docs/source/getting-started/wsl-setup.rst
```diff
@@ -0,0 +1,62 @@
...
+Docker Resource Limits
+~~~~~~~~~~~~~~~~~~~~~
+
+- Increase Docker memory allocation to at least **8 GB** in Docker Desktop.
+
+What Not To Do
+--------------
+
+- Do **not** run manual backend installations inside WSL.
+- Do **not** run PostgreSQL separately outside Docker.
```
MoralCode, line 71: Are there reasons why these things are listed as things not to do? what would the consequences be if someone did them? are they just things you didnt test? or do they cause breakage?

Other people may not have the exact same situation as you, so explaining what happens in these cases can help others decide if this advice applies to them

docs/install-wsl.md
himasree9: Thank you for the clarification.
I will remove docs/install-wsl.md and ensure that all of its content is fully represented in docs/source/getting-started/wsl-setup.rst.
I’ll update the PR shortly

docs/source/getting-started/wsl-setup.rst
himasree9: These recommendations are based on observed breakage and conflicts.

Manual backend installation inside WSL:
When the backend is installed manually inside WSL while also using Docker, it commonly results in version mismatches, duplicated services, and port conflicts (especially with Python environments, PostgreSQL clients, and dependency versions). This can cause Augur services to fail to start correctly or behave inconsistently between Docker and non-Docker components.

Running PostgreSQL outside Docker:
Running PostgreSQL separately on Windows while also using the Docker-based stack can lead to connection failures and unexpected behavior because the containers are configured to communicate with the PostgreSQL service defined in docker-compose. Using an external instance breaks that expected network configuration and can prevent Augur from initializing or migrating the database correctly.

I will update the documentation to make these consequences explicit so users can better understand when and why these recommendations apply.

himasree9:
Thanks for the feedback!
I’ve updated the documentation with explanations describing the consequences of those setups and removed the redundant Markdown version as suggested.

docs/source/getting-started/wsl-setup.rst
MoralCode: It sounds like you are trying to essentially run two instances of augur at the same time (which would explain the port conflicts).

is there a reason you were trying to do that, rather than just running things in _one_ of the following ways:
- install docker directly to windows and use the docker stack (the method i would recommend, although that is out of scope for a WSL-specific guide)
- Install augur manually in WSL (i.e. you install and configure each service yourself rather than using the docker compose)
- install docker in WSL and use the docker stack there (this seems really inefficient and unlikely to work properly but could be worth a try)

himasree9: Thanks for the clarification.
That’s exactly the situation I encountered. The conflicts happened when parts of Augur were running via Docker while other components were installed manually inside WSL at the same time. That effectively created two partially overlapping environments (Docker-managed services and WSL-host services), which caused port conflicts and inconsistent service behavior.

I wasn’t intentionally trying to run two instances — it was more of an incremental setup attempt where manual installation inside WSL was combined with the Docker stack before realizing how tightly the Docker configuration assumes full control of the backend services.

I agree with your suggested approaches, and the documentation now reflects the recommendation to choose one clean setup path (Docker-based or fully manual) rather than mixing the two. The WSL guide is intentionally scoped to the Docker-based approach since that is the most stable and reproducible option for new users.

MoralCode: yeah docker and the manual install are essentially mutually exclusive processes. either you are in charge of managing all the various services that make up augur, or you leave it to the docker compose.

If you run a manual install and choose to run one service, like redis, in docker (without compose, or by making your own custom compose with only that service, custom port numbers, etc) thats fine, but at that point you are largely on your own since you are deviating from the standard docker setup that makes things easier to debug when there are issues.

himasree9: Yes, exactly—the docs now emphasize choosing a single setup path and avoiding mixed service management to prevent those conflicts. Thanks again for the guidance on this.

MoralCode:
As things are currently, we dont officially support windows. If we proceed with adding an install guide like this, i think it should either be on a third-party website (i.e. personal blog post or something that maybe can be linked to from the docs in a "this might work, but your mileage may vary" way, or otherwise clearly marked/separated  from official supported documentation.

Anyone have thoughts on which path makes more sense?

himasree9:
@MoralCode,
Thanks for raising this.
I’m happy to clearly mark this as community-supported / not officially supported, or move it to a separate or external reference if preferred.
My goal is simply to document a commonly attempted setup while being explicit about its limitations.
Happy to follow whatever direction makes the most sense for the project.

MoralCode:
> I’m happy to clearly mark this as community-supported / not officially supported, or move it to a separate or external reference if preferred.

If you weren't able to put this document in the augur repo at all, do you have an idea of where you would put it instead?

I think it may be helpful to maybe have something like a personal blog and/or dedicated Repository for Augur in WSL/Windows users in general. That seems like something that could also potentially be a better home for something like #3546 (i.e. use a whole issue tracker to do one issue-per problem rather than what we do here with one issue for all the problems)


Would you be willing to maintain/lead/manage a repo for Augur windows users like that? Or do you think it may be worth starting a conversation with other windows users of augur to see who wants to help co-lead something?


I'd be happy to chat about this in more detail over on the [CHAOSS Slack](https://chaoss.community/kb-getting-started/) in the #wg-augur-8knot channel if you'd like to maybe start a group discussion too

himasree9:
@MoralCode ,
That sounds like a good option to me. If everyone’s okay with it, I’d be interested in helping start something like a community-maintained repo for Augur on Windows / WSL.
Since I’m still a relatively new contributor, I think it would work best as a co-led effort rather than something I maintain alone. I’d be happy to collaborate with others and help drive it forward together.
I’m also open to continuing this discussion on Slack if that’s helpful.

MoralCode:
> I’m also open to continuing this discussion on Slack if that’s helpful.

i started a slack thread already and one other person has already joined it. I think this could have some momentum and be a useful resource that the augur maintainers can point to for people on WSL, but without having to commit to maintaining it just yet (since its an unsupported platform)

himasree9:
Thanks for starting the discussion — I agree this could be a useful resource.
Happy to help however I can as things move forward.

MoralCode:
Im going to close this PR because WSL is not officially supported but i have made a note of it so that if you or others decide to make a separate resource for supporting WSL people, you have this content to build upon

Thanks for your contribution and I look forward to chatting on slack!

PR #3494, opened 2025-12-23
docs: Document unused and legacy tables with retention and removal guidance

### LINKED ISSUE #3425: Update unused table list - consider moving to a `legacy` schema
There appears to be a [docs page](https://oss-augur.readthedocs.io/en/main/schema/unused_data.html) listing unused tables. 

I strongly suspect this is out of date. 

I think its worth going through and auditing which tables are in use and sorting them into three buckets:
1. in use tables
2. unused tables we should keep for data preservation on older instances
3. unused tables that should be dropped

category 2 is what we should document in this page of the docs and potentially move over to a new schema

See also #3423

PredictiveManish:
**Description**
Updated the unused-tables documentation with clear categories and recommendations.
- Added categorization of tables into in-use, legacy (preserve/move to legacy schema), and candidates for removal, with notes on DEI badging and sentiment tables that remain active.
- Documented recommended handling steps (preserve legacy data or drop after verification).
- Solves #3425 

This PR fixes #3425 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docs/source/schema/unused_data.rst
+++ b/docs/source/schema/unused_data.rst
@@ -1,5 +1,5 @@
 List of Unused and Legacy Data Tables
-=============================
+=====================================
 
 This page tracks tables that are not part of the current collection/analysis
 pipelines, and how we plan to handle them. Tables are grouped into three


MoralCode:
This PR is something that requires input from a maintainer before starting since even I am too new to this project to know which tables are still in use or not.

I'm going to close this pr because the underlying issue still needs input from a maintainer to specify how to act on each table (and therefore which ones need documenting as legacy because we chose to keep them in for a while longer and not use them)

That said, thanks for your contribution! I've seen your name around on several things now and I hope you'll continue finding ways to contribute that are interesting for you!

PredictiveManish:
> This PR is something that requires input from a maintainer before starting since even I am too new to this project to know which tables are still in use or not.
> 
> I'm going to close this pr because the underlying issue still needs input from a maintainer to specify how to act on each table (and therefore which ones need documenting as legacy because we chose to keep them in for a while longer and not use them)
> 
> That said, thanks for your contribution! I've seen your name around on several things now and I hope you'll continue finding ways to contribute that are interesting for you!

feeling sad, I am unable to solve this issue, but will surely work for it and also for other things which I found solvable with my capabilities.

PR #3483, opened 2025-12-19
docs: restructure install flow, promote Docker as default

shivansh31414:
Title:  
docs: restructure install flow, promote Docker as default
Changes Made:

toc.rst: Reordered installation methods, placing Docker first; transformed separate Docker section into “Advanced Docker Deployment.”

quick-start.rst: Promoted Docker as the primary recommendation; added a side‑by‑side comparison matrix of Docker vs. manual install.

index.rst: Adjusted navigation to consolidate Docker under the Getting Started flow.

README.md: Updated entry point guidance to highlight Docker as the recommended default path
Summary:  
This PR reorganizes the installation documentation to make Docker the default/primary installation method, while keeping manual installation paths available for advanced users. It addresses Issue #3477.

PR #3503, opened 2025-12-29
My fix branch

### LINKED ISSUE #3477: Docker quickstart setup instructions should be up near the manual setup instructions
Right now the docker quick start is in its own section further down the navigation than the manual install method (the primary one the docs direct people to from the homepage).

These two instructions should be side by side, ideally with docker being the default method.

This task may be a little big for a first-timer or AI to handle, so not tagging as good first anything. Strongly suggest that the ideal person to work on this is someone with at least a few docs contributions already done.

shivansh31414:
**Description**
as the feedback suggested added and removed lines that created misconceptions like removing the line in documentation that suggested augur does  not support some python versions and some deletions to make the pr more suitable

This PR fixes #3477.

### DIFF
--- a/docs/source/getting-started/toc.rst
+++ b/docs/source/getting-started/toc.rst
@@ -9,13 +9,13 @@ Choose Your Installation Method
 
 Augur offers two primary installation approaches:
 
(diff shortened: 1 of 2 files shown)


MoralCode:
Does this PR replace #3483?

shivansh31414:
yes

PR #3507, opened 2026-01-04
Fix cross-contamination between GitHub and GitLab columns in contribution

zeba-source:
…tors table

**Description**

This PR fixes #

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -46,9 +46,9 @@ The ``message_insights_worker`` uses a system-level package called OpenMP. You w
 
 Caching System (Redis)
 ----------------------
-* `Linux Installation <https://redis.io/docs/getting-started/installation/install-redis-on-linux/>`__
-* `Mac Installation <https://redis.io/docs/getting-started/installation/install-redis-on-mac-os/>`__
-* `Windows Installation <https://redis.io/docs/getting-started/installation/install-redis-on-windows/>`__
+* `Linux Installation <https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/install-redis-on-linux/>`__
+* `Mac Installation <https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/install-redis-on-mac-os/>`__
+* `Windows Installation <https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/install-redis-on-windows/>`__
 
 Message Broker (RabbitMQ)
 -------------------------


sgoggins:
Right now the docker builds are failing. I will rerun them, however, given the substantial scale of this PR it will require further triage once CI is working.

augur/application/schema/alembic/versions/38_restore_contributor_unique_constraints.py
MoralCode, line 1: Database migration files should be for migrating the database schema. It is best for them to be auto-generated through alembic. This file contains custom logic and raw SQL queries, which does not fit the pattern of the most recent set of database migrations.

Also since a prior migration removed the uniqueness constraints, it is highly likely that there is a reason that was done and I would want to understand why that was done and why it is now being undone. I dont currently have bandwidth to do such a deep dive.

augur/application/db/data_parse.py
```diff
@@ -689,48 +689,178 @@ def extract_needed_contributor_data(contributor, tool_source, tool_version, data
...
+        "cntrb_full_name": contributor.get('name'),  # GitLab 'name' field
+        
+        # -------------------------------------------------------------------------
+        # GITLAB-SPECIFIC COLUMNS (gl_*) - CORRECT MAPPING
+        # -------------------------------------------------------------------------
+        # These columns store GitLab-specific data. The GL-UNIQUE-B and GL-UNIQUE-C
+        # constraints ensure gl_id and gl_username are unique for GitLab users.
+        
+        "gl_id": contributor['id'],                        # GitLab user ID (unique constraint: GL-UNIQUE-B)
+        "gl_username": str(contributor['username']),       # GitLab username (unique constraint: GL-UNIQUE-C)
```
MoralCode: Seems like this core fix is useful (i see in the diff that we were assigning to `gh_` fields in a function meant for gitlab.

That said, its hard to review a PR with so many changes, especially this diff where large parts of this function were changed (with the only changes to many lines are simply comments that re-explain the code or add emojis)

MoralCode (changes requested):
Overall i think this PR is fundamentally on the right track, but needs a lot more review and editing to clean up unnecessary comments and lines of code to make it easy to review.

It also looks like you also wrote code for testing and migrating existing users data over to the correct columns. I'd be curious to learn more about your proposed strategies for handling these cases since I think making sure the approach is correct first, before reviewing the implementation is also a good idea.


Thanks for submitting this PR, I'd love to continue working with you to refine this and turn it into something that other maintainers can more easily review and merge

PR #3513, opened 2026-01-06
add CI job for running the unit tests

MoralCode:
**Description**
This adds a github CI job that runs the unit tests. This job is in a new file called functional_test because the unit tests that are currently enable test functionality that is critical to augurs functioning, so if they fail, the PR is **substantially** likely to contain broken code.

This Job is configured to run on each of the following minor python releases:
- 3.10
- 3.11
- ~~3.12~~ (disabled. planning to merge https://github.com/chaoss/augur/pull/3534 afterwards to get this working)
- ~~3.13~~ (will enable in future PR)
- ~~3.14~~ (will enable in future)

These tests are also repeated on macOS and Ubuntu runners.

This is - as far as im aware - all the versions of python Augur currently supports.

This behavior was originally part of #3422 but has been moved out so that the failures can be dealt with in smaller groups

Currently the only behavior being unit tested here is for the Hierarchical config system, but I anticipate adding more in the future

**Notes for Reviewers**

This job will run when this PR is proposed. if the CI is passing, this PR should be considered "tested" and working

**Signed commits**

### DIFF
--- a/.github/workflows/functional_test.yml
+++ b/.github/workflows/functional_test.yml
@@ -11,6 +11,7 @@ jobs:
   test:
     name: test with ${{ matrix.env }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 15
     strategy:
       fail-fast: false
       matrix:


MoralCode:
Having automated CI testing in place is helpful for many of my future plans, including writing unit tests to confirm database leaks, getting more of the tests working, adding new tests, etc.

.github/workflows/functional_test.yml
```diff
@@ -0,0 +1,42 @@
...
+          - "3.14"
+        os:
+          - ubuntu-latest
+          - macos-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v7
+      - name: Install tox
+        run: uv tool install --python-preference only-managed --python 3.11 tox --with tox-uv --with tox-gh
```
MoralCode: note to self: figure out the python versions - doing it in tox probably makes the most sense

.github/workflows/functional_test.yml
shlokgilda, line 1: Worth adding a timeout here? Something like `timeout-minutes: 10` would prevent jobs from hanging indefinitely if something goes wrong.

augur/application/config.py
```diff
@@ -546,6 +546,8 @@ class JsonConfig(ConfigStore):
 
     def __init__(self, json_data, logger: logging.Logger):
         super().__init__(logger)
+        if not self.writable:
```
shlokgilda, line 549: Since `writable` is hardcoded to `False` for `JsonConfig` (line 554), this condition is always true. Not wrong, but could be simplified to just always deepcopy.

Up to you though - the current form is more defensive if writable ever becomes dynamic.

MoralCode: Yeah i think im anticipating the possibility of something writable since I really want to move away from having config in a database in the long term. I just think we maybe need a better backend for it (maybe a config oriented version of keyman?) that isnt just an in-memory dict (too volatile) or file on disk (no way to deal with conflicts between what the user wants and what augur is trying to set)

.github/workflows/functional_test.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    branches: [main, release]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: test with ${{ matrix.env }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
```
MoralCode, line 13:
```suggestion
    runs-on: ${{ matrix.os }}
    timeout-minutes: 15
```

.github/workflows/functional_test.yml
MoralCode: added (the lazy way because im not at my work computer rn

shlokgilda (approved):
LGTM

MoralCode:
waiting for the docker smoke test and will probably merge without a second review given the low impact (i.e. this is new stuff, not touching old code) of this PR and the relatively high priority (other stuff is waiting on this)

PR #3514, opened 2026-01-06
fix(util): replace string parsing with AST in inspect_without_import #3504

### LINKED ISSUE #3504: Consider replacing string parsing with ast
inspect_without_import.py currently uses basic string parsing to get the functions from start_tasks.py
This might cause problems with indenting and formatting with small things like whitespaces. 

I'd suggest using Python's ast to inspect the file, as this makes it more robust and not susceptible to the parsing errors previously mentioned. 

If there isn't anyone working on this already, I'd be glad to give this a go. Waiting for a maintainer to review this.

guptapratykshh:
**Description**
- Replaced the string parsing logic in `augur/util/inspect_without_import.py` with the `ast` module to robustly extract phase function names from `start_tasks.py`.
- Added a new test suite in `tests/test_util/test_inspect_without_import.py` to verify the fix and ensure it handles various edge cases (indentation, decorators, multi-line definitions).
- Updated comments and docstrings to be more human-readable and concise.

This PR fixes #3504

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/util/inspect_without_import.py
+++ b/augur/util/inspect_without_import.py
@@ -55,7 +55,7 @@ def get_phase_names_without_import() -> List[str]:
     for node in ast.walk(tree):
         if isinstance(node, ast.FunctionDef):
             # We want any function that identifies itself as a 'phase'
-            if '_phase' in node.name:
+            if node.name.endswith('_phase'):
                 phase_names.append(node.name)
     
     return phase_names
--- a/tests/test_util/test_inspect_without_import.py
+++ b/tests/test_util/test_inspect_without_import.py
@@ -25,16 +25,19 @@ class TestGetPhaseNamesWithoutImport:
         for name in phase_names:
             assert '_phase' in name, f"Function {name} doesn't seem to be a phase function."
         
+        # These are the outputs of the original version of the function as of 7a46497aaed6c6dbecce9a1f8dda58282f9dd9fe
         expected_phases = [
             'prelim_phase',
-            'prelim_phase_secondary',
             'primary_repo_collect_phase',
-            'primary_repo_collect_phase_gitlab',
             'secondary_repo_collect_phase'
         ]
         
         for expected in expected_phases:
             assert expected in phase_names, f"We missed a required phase: '{expected}'"
+
+        for actual in phase_names:
+            assert actual in expected_phases, f"We have an extra phase: '{actual}'"
+
     
     def test_handles_different_indentation(self):
         """Ensures that messed up (but valid) indentation doesn't break anything."""


augur/util/inspect_without_import.py
```diff
@@ -1,19 +1,59 @@
...
+    try:
+        tree = ast.parse(source_code, filename=start_tasks_path)
+    except SyntaxError as e:
+        raise SyntaxError(
+            f"The file exists, but it contains invalid Python syntax: {e}"
+        ) from e
+    
+    phase_names = []
+    
+    # We walk through every node in the abstract syntax tree
```
shlokgilda: `ast.walk()` finds nested functions too. If someone writes:
```python
def primary_phase():
    def _phase_helper():  # Would be collected
        pass
```
Old code only caught top-level functions. If that's the intent:
```python
for node in tree.body:  # Only top-level
    if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
```
Not blocking, just worth considering.

shlokgilda:
Great refactor. I've added some inline comments. Other than that, could you add a note about whether you used any AI tools (Claude, Copilot, etc.) for this PR? If so, what were they used for? You can see examples from other closed PRs. But otherwise looks like good PR.

guptapratykshh:
Thanks for the review. No, I did not use any AI tools for the implementation in this PR.

augur/util/inspect_without_import.py
```diff
@@ -1,19 +1,59 @@
-#needed as a workaround since python executes imported files
-#This presents a problem since importing the phase functions themselves needs information from the config
-#while the config needs the phase names before population
-#The solution is to either make the user define the phase names seperate or to do this
-#Which is to import the .py as text and parse the function names.
```
MoralCode, line 5: This seems to be removing important documentation that explains why this hacky code is here in the first place. Can you make sure that information remains/isnt deleted by your PR?

tests/test_util/test_inspect_without_import.py
MoralCode, line 1: Nice work creating a unit test for this!

augur/util/inspect_without_import.py
```diff
@@ -1,19 +1,59 @@
...
+    Raises:
+        FileNotFoundError: If we can't locate the start_tasks.py file.
+        SyntaxError: If start_tasks.py has broken Python syntax.
+    """
+    tasks_dir = os.path.join(
+        os.path.dirname(os.path.dirname(__file__)),
+        "tasks"
+    )
+    start_tasks_path = os.path.join(tasks_dir, "start_tasks.py")
+
```
MoralCode: Personally id prefer if we used `pathlib` here (python built in object oriented path traversal) rather than `os.` calls, but maybe thats just my stylistic preference

@shlokgilda what do you think?

tests/test_util/test_inspect_without_import.py
MoralCode: This should probably get added to the pytest code in #3513 depending on which merges first

guptapratykshh: Thanks for the update. That makes sense. I will watch #3513 closely. If it merges before this PR, I will rebase and move these tests into the new structure.

augur/util/inspect_without_import.py
guptapratykshh: replaced and used pathlib over here

augur/util/inspect_without_import.py
guptapratykshh: reverted these comments

tests/test_util/test_inspect_without_import.py
MoralCode: i dont think they need to move just yet, you'd jus tneed to add a line to the config so that these tests get run along with the others that pytest is set up for

MoralCode:
id like to test this (run the unit tests) and add them to the CI, but overall looks good!

augur/util/inspect_without_import.py
```diff
@@ -3,17 +3,59 @@
...
+        ) from e
+    
+    phase_names = []
+    
+    # We walk through every node in the abstract syntax tree
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef):
+            # We want any function that identifies itself as a 'phase'
+            if '_phase' in node.name:
+                phase_names.append(node.name)
```
MoralCode, line 59: does this filter for functions? or could someone create a top level variable matching this pattern and cause it to break too?

guptapratykshh: No, this will not match top level variables. The check if isinstance(node, ast.FunctionDef): filters for function definitions only. Variables show up as ast.Assign or ast.AnnAssign nodes in AST tree, so this logic completely skips them. 

I just added specific unit test, test_ignores_variables_with_matching_name, to verify this behavior and ensure safety.

tests/test_util/test_inspect_without_import.py
MoralCode: heads up, #3513 is probably going to merge very soon

guptapratykshh: have also rebased the recent chanegs from the main

guptapratykshh: i think it is ready , as #3513 is merged

guptapratykshh:
the changes have done , pleae have a look

tests/test_util/test_inspect_without_import.py
MoralCode: Can you include this directory in the list of directories that will automatically be tested (i believe its near the bottom of pyproject.toml

MoralCode: has this been done?

MoralCode:
These new unit tests are not added to pyproject.toml such that they will be automatically tested. once thats done i will do a final review pass and merge this

MoralCode:

Thanks!

MoralCode:
Taking this over.

rebased
applied my feedback from earlier (added to the unit test file)
updated the logic to validate the output of this new function is an identical match for the current output.

MoralCode:
i have tested this insofar as i was running the unit tests both before and after this change to validate both versions of the function are passing. 

Happy to split this PR to merge the unit test first if that would help de-risk this or demonstrate these assertions publicly

tests/test_util/test_inspect_without_import.py
```diff
@@ -0,0 +1,382 @@
...
+def test_phase(repo_git, full_collection):
+    pass
+
+class SomeClass:
+    def another_phase(self, repo_git):
+        pass
+
+def    weirdly_spaced_phase   (repo_git):
+    pass
+'''
```
shlokgilda, line 54: These inline tests use `'_phase' in node.name` (contains check), but the actual implementation uses `node.name.endswith('_phase').` These are different matching rules. For eg. `build_primary_phase_request` contains `_phase` but doesn't end with it.

This matters because `test_ignores_non_phase_functions` explicitly asserts `build_primary_phase_request` IS found, which is correct under `in` but wrong under `endswith.` So the tests pass, but they're testing different logic than what ships.

shlokgilda: Maybe we refactor `get_phase_names_without_import()` to accept an optional `source_path` parameter (defaulting to the current hardcoded path). Then these tests can call the real function with a temp file instead of reimplementing the AST walk inline. That way any change to the matching logic is automatically covered.

tests/test_util/test_inspect_without_import.py
```diff
@@ -0,0 +1,382 @@
...
+        phase_names = []
+        
+        # AsyncFunctionDef is a different node type than FunctionDef
+        for node in ast.walk(tree):
+            if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and '_phase' in node.name:
+                phase_names.append(node.name)
+        
+        assert 'async_phase' in phase_names
+        assert 'sync_phase' in phase_names
+        assert len(phase_names) == 2
```
shlokgilda, line 237: This test checks `isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))` but the implementation only checks `ast.FunctionDef.` Since this test does its own inline AST walk rather than calling the real function, it passes. But it's testing behavior the function doesn't actually have. If an async phase function were added to `start_tasks.py,` this test would still pass while the function silently misses it.

augur/util/inspect_without_import.py
```diff
@@ -3,17 +3,59 @@
...
+            f"The file exists, but it contains invalid Python syntax: {e}"
+        ) from e
+    
+    phase_names = []
+    
+    # We walk through every node in the abstract syntax tree
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef):
+            # We want any function that identifies itself as a 'phase'
+            if node.name.endswith('_phase'):
```
shlokgilda, line 58: Most of the edge case tests use `'_phase' in node.name` (contains) instead of `endswith('_phase').` That means the tests are validating a broader matching rule than what this actually ships. More on that in the test file comments.

Also, this only checks `ast.FunctionDef`. If someone ever adds an `async def phase` function, it'd be silently skipped. Probably fine for now since there aren't any, but worth a comment or including `ast.AsyncFunctionDef` in the check.

tests/test_util/test_inspect_without_import.py
MoralCode: ah so this is probably because i changed it

MoralCode: the endswith logic is a closer match to the behavior existing non AST version of the function i think. i probably didnt go through all the test cases in very much detail

MoralCode:
yeah i agree with shlok, the tests need to not re-implement things. they should exercise the actual code paths being used.

original contributor is no longer active to provide details on whether these tests were written with AI or not so we can only guess.

Might make sense to have one of us, as maintainers reimplement them

That said, this is more a tech debt reduction and generally quite low priority fix, so maybe not worth doing that for now.

shlokgilda:
Hey, I've got the fixes ready on my fork but don't have push access to this branch. @MoralCode what's the best way to get these changes onto this branch? Should I open a separate PR instead?

PR #3516, opened 2026-01-07
Revert database url retrieval so manual install works

MoralCode:
**Description**
This PR fixes the database url handling in Alembic's config so that it accounts for the `db.config.json` file used by manual installs.

This issue was introduced in #3416 

The solution is essentially a revert back to calling get_database_string(). It also removes the code that fetches the string from alembic.ini, since thats a file committed to the repo and thus would never be a good place to store the database connection string/credentials.

**Notes for Reviewers**

**Signed commits**


MoralCode:
@sgoggins could you test this on your instance to confirm it works?

Issue #3277, opened 2025-09-17
Add option to use Github Enterprise API for api key checks


MoralCode:
Augur is difficult to use for people using github's enterprise api (`[api.SUBDOMAIN.ghe.com](http://api.subdomain.ghe.com/)`) because keyman seems to validate credentials with a [hardcoded API path](https://github.com/chaoss/augur/blob/9050232e/augur/tasks/github/util/github_api_key_handler.py#L12).

Ulincsys:
Related: #3087

SuyashJain17:
This looks like a well-scoped blocker for GitHub Enterprise users.
I’m going to trace where API key validation happens (keyman / auth flow) and see how the GitHub API base URL is currently hardcoded. I’ll aim for a minimal change that makes the API domain configurable while keeping api.github.com as the default. I’ll follow up with a proposal or PR once I understand the current flow.

Issue #3236, opened 2025-07-21
Epic Tracking: New Integrations


MoralCode:
Figured I'd make a tracking issue to keep track of the various new ideas for different platforms that may be worth eventually adding.

Inclusion in this issue doesn't mean it will get added, or will happen anytime soon - just that they area ideas that have been mentioned and may have enough useful repos using them to warrant adding if the circumstances are right. 

- **Template**
  - Contains: [repo, issues, PRs]
  - Usecase: 
  - Examples: 
  - Platform technical docs: 

---

- **[Forgejo](https://forgejo.org/)**
  - Contains: repo, issues, PRs
  - Usecase: Fedora (upcoming), SFConservancy projects
  - Example repo/instance: https://f.sfconservancy.org/explore/repos
  - Platform technical docs: TODO

- **Cgit**: #3127 
  - Contains: Git repo
  - Examples:
    - https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git
    - https://cgit.git.savannah.gnu.org/cgit/gnuzilla.git
    - 
  - Technical Info on cgit: https://git.zx2c4.com/cgit/about/

- **Gitiles**
  - Contains: repo
  - Usecase: Chromium, CEF, and related google projects
  - Examples: https://gerrit.googlesource.com/gitiles/
  - Platform technical docs: https://gerrit.googlesource.com/gitiles/
 

- **[Gerrit](https://gerrit.googlesource.com/gerrit)**
  - Contains: PRs/code review
  - Usecase: Chromium/google source
  - Examples: 
    - https://chromium-review.googlesource.com/c/angle/angle/+/6164683
  - Technical info on the platform: NOTE that a lot of their API docs are liable to be wrong/not up to date with current behavior

- **[bugzilla](https://www.bugzilla.org/)**
  - Contains: issues
  - Usecase: [Rappel](https://github.com/oss-aspen/Rappel/issues/102)
  - Example: glibc https://www.gnu.org/software/libc/sources.html
  - Technical info on the platform: TODO

- **[mantis bug tracker](https://www.mantisbt.org/)**
  - Contains: issues
  - Usecase: https://gramps-project.org/ uses this as their bug tracker
  - Technical info on the platform: TODO

- **GNU [Savannah](https://savannah.gnu.org/)**
  - Contains: Issues
  - Usecase: TODO
  - Example repo: TODO
  - Platform technical docs: TODO
- **Independent gitlab instances** (i.e. mainly setting up different keys for different instances*)
[...]

PR #3517, opened 2026-01-07
Fix GitLab data incorrectly stored in GitHub columns

### LINKED ISSUE #3469: Cross-contamination between github and gitlab columns in contributors table
im seeing lots of gh_* prefixed columns being used for gitlab data too...

_Originally posted by @MoralCode in https://github.com/chaoss/augur/issues/3435#issuecomment-3656839767_

While looking into this migration issue, I noticed that there are duplicate values in our database for the `gh_login` and `cntrb_login` fields in our db. Looking at a full row from one dupicate user, its clear that this data came from two different forges (one github, one gitlab).

There seems to be several issues here:
- cntrb_login and gh_login may both be populated together by both the github events task and the gitlab issue comments task.
- for the gitlab user, gl_username is NULL
- gh_avatar_url has a non-null value for the gitlab user (seems incorrect given the table name) while the gl_avatar_url field is NULL
- same thing for gh_url
- gh_user_id has a value for the gitlab user, while gl_id is NULL

Further, it seems like the gitlab username and other contributor data is largely being populated in the columns that have `gh_*` prefixes. This means that our Database models, which are seemingly written to enforce username uniqueness for gh and GL usernames separately, are not able to actually do so (and may not even be currently being constrained)

zeba-source:
## Summary
Fixes GitLab contributor data being incorrectly stored in GitHub-prefixed columns (`gh_*`) instead of GitLab columns (`gl_*`).

## Changes
- **augur/application/db/data_parse.py**: Map GitLab data to `gl_*` columns instead of `gh_*`
- **augur/tasks/util/contributor_utils.py**: Add forge detection utility
- **tests/**: Add integration tests to verify correct column usage

## Scope (Simplified per maintainer feedback)
This PR focuses only on the core code fix:
- ✅ Core column mapping fix in `data_parse.py`
- ✅ Essential tests for validation
- ⏭️ Database migrations → Separate PR (per maintainer request)
- ⏭️ Data cleanup scripts → Separate PR (per maintainer request)

## Previous Feedback Addressed
Per @MoralCode's review:
- ✅ Removed excessive documentation files (~140K lines)
- ✅ Removed database migration file
- ✅ Removed data cleanup scripts  
- ✅ Cleaned up excessive comments and emojis
- ✅ Reduced PR scope from 200+ files to 12 files

## Testing
- Unit tests verify GitLab→gl_* and GitHub→gh_* mapping
- Integration tests confirm no cross-contamination
- All tests pass locally

Fixes #3469

---

**Signed commits:** ✅ Yes, commits are signed
```

### DIFF
--- a/augur/api/gunicorn_conf.py
+++ b/augur/api/gunicorn_conf.py
@@ -40,7 +40,7 @@ del is_dev
 # set the log location for gunicorn    
 logs_directory = get_value('Logging', 'logs_directory')
 
-is_docker = os.getenv("AUGUR_DOCKER_DEPLOY").lower() in ('true', '1', 't', 'y', 'yes')
+is_docker = os.getenv("AUGUR_DOCKER_DEPLOY", 'False').lower() in ('true', '1', 't', 'y', 'yes')
 accesslog = f"{logs_directory}/gunicorn.log"
 errorlog = f"{logs_directory}/gunicorn.log"
 
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -1,6 +1,7 @@
 import sqlalchemy as s
 from sqlalchemy import and_, update
 import json
+import copy
 from typing import List, Any, Optional
 import os
 from augur.application.db.models import Config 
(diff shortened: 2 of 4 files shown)


MoralCode:
While this PR may solve the underlying issue (per my [prior comments](https://github.com/chaoss/augur/pull/3507#discussion_r2663237905)), I think it may actually be a better direction for the project to resolve the issue a different way.

Ultimately, if we correct this issue by movign gitlab data to the `gl_` columns, we paint ourselves into an awkward corner when it comes time to add support for more platforms (i.e. forgejo, see #3236) in the future because adding a new platform may mean duplicating our existing tables.

Id prefer to solve this issue by instead allowing the columns that we are currently trying to separate to coexist. by that I mean instead of `gl_username` and `gh_username` we should just have `platform_username` and a `platform` column indicating which platform the data came from.

I'd also like to integrate this system with other platform-oriented improvements, such as enabling the domains and keys for a platform to be changed independently of the underlying API implementation (i.e. to support users who want to use github enterprise #3277)


Ultimately a lot of these changes are going to be very big, very complex, and probably out of scope for newcomers, even (especially) with advanced AI tools.

Thanks for your contribution to this issue though! I'd be happy to chat about this in more detail over on the [CHAOSS Slack](https://chaoss.community/kb-getting-started/) in the #wg-augur-8knot channel if you'd like the team to help you find something to work on.

zeba-source:
Thanks for the detailed feedback and for taking the time to explain the
architectural direction! I completely understand your concern about
scalability - the platform column approach makes much more sense for
supporting multiple platforms in the future rather than creating new column
sets for each one.

I appreciate you thinking about the long-term maintainability of the
codebase. You're right that a more comprehensive refactoring to support
multi-platform architecture is probably beyond the scope of what I can
tackle right now, especially as someone still getting familiar with the
project.

I'd love to continue contributing to Augur in ways that are more suitable
for newcomers. I'll join the CHAOSS Slack #wg-augur-8knot channel - it
would be great to chat with the team about other issues that might be a
better fit for my current experience level.

Thanks again for the guidance and for being so welcoming to new
contributors!

PR #3518, opened 2026-01-07
implement timestamp correction utilities and tests for git commit data

### LINKED ISSUE #3472: facade_bulk_insert_commits should also correct `cmt_committer_timestamp`
> The existing fix in `lib.py:facade_bulk_insert_commits`) only validates `cmt_author_timestamp`, but in this case the corruption is in `cmt_committer_timestamp`:
> 
> ```python
> # lib.py lines 204-224
> placeholder_date = commit_record['cmt_author_timestamp']  # Only checks AUTHOR
> # ...
> if int(tzdata) not in postgres_valid_timezones:
>     tzdata = "+0000"  # Fix applied
> else:
>     raise e  # Author's timezone is valid, so re-raise!
> ```
> 
> Looking at the actual data for this commit from the stack trace:
> ```
> 'cmt_author_timestamp__1': '2025-11-03 16:28:43 -0500',  # VALID
> 'cmt_committer_timestamp__1': '2106-02-07 06:28:23 -13068837',  # INVALID
> 'cmt_filename__1': 'lib/event.c',
> 'cmt_commit_hash__1': '5de262a839b04ff4649d4e04784c98c08322d7b4',
> 'cmt_committer_name__1': 'Mark Stapp',
> 'cmt_added__1': 252,
> ```
> The author timestamp (-0500) is valid, so the fix logic hits `else: raise e` and never fixes the committer timestamp.
> 
> **Problematic Commit:** This is the commit with the corrupted timestamp: https://github.com/FRRouting/frr/commit/5de262a839b04ff4649d4e04784c98c08322d7b4. If you open the link, GitHub interprets the corrupted timestamp as "Jan 4, 2121". PostgreSQL sees it as 2106-02-07 06:28:23 -13068837 and rejects the impossible timezone offset.
> 
> **Fix:** The binary reduction logic needs to validate both `cmt_author_timestamp` and `cmt_committer_timestamp` when isolating bad records. 

 _Originally posted by @shlokgilda in [#3458](https://github.com/chaoss/augur/issues/3458#issuecomment-3656497467)_

shlokgilda:
**Description**

Fixes #3472 - PostgreSQL insertion failures due to invalid committer timestamps.

**The Bug:**
`facade_bulk_insert_commits()` only validated `cmt_author_timestamp` when handling DB insertion failures. If the author timestamp was valid but the committer timestamp had an invalid timezone (e.g., `-13068837` from git corruption), the insert would fail permanently.

**The Fix:**
- Created `augur/tasks/git/correction.py` with timezone validation functions (simple string operations)
- Modified `lib.py` to validate **both** author and committer timestamps when the binary search isolates a problematic commit
- Uses author timestamp as fallback for invalid committer timestamp to minimize data loss
- Only validates on insertion failure (keeps the existing performance optimization intact)

**Changes:**
1. **NEW:** `augur/tasks/git/correction.py` - Timestamp validation utilities
2. **MODIFIED:** `augur/application/db/lib.py` - Now validates both timestamps on failure
3. **NEW:** `tests/test_tasks/test_git/test_correction.py` - 12 unit tests (all passing)

**Notes for Reviewers**
- The fix maintains the existing binary search optimization - we only validate timestamps when PostgreSQL rejects an insert, not proactively
- `POSTGRES_VALID_TIMEZONES` includes all real-world timezone offsets (-12:00 to +14:00). Some of the timezone offsets were missing. Reference: https://docs.oracle.com/cd/E19563-01/819-4437/anovd/index.html
- Fallback chain: invalid committer → use author timestamp → UTC (minimizes data loss per issue discussion)

**Signed commits**

---

**AI Disclosure:** I used Claude Code to assist with writing docstrings, code review, writing test cases, and drafting this PR description. All test cases were manually verified and are passing locally.


augur/application/db/timestamp_utils.py
MoralCode, line 1: Can you make sure this file is added to the list of test files that get run in CI in `pyproject.toml`

sgoggins:
@shlokgilda / @MoralCode : The errors are now that we've exceeded our rate limit ... so, need to come back to this one.

MoralCode:
Rebased and force pushed., also removed seans unnecessary commit that was there to kick the CI

and addressed my own feedback about adding the new tests to the CI

augur/application/db/lib.py
```diff
@@ -1,17 +1,16 @@
...
-from sqlalchemy.exc import DataError
+from sqlalchemy import func
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.exc import OperationalError
 from psycopg2.errors import DeadlockDetected
 from typing import List, Any, Optional, Union
 
 from augur.application.db.models import Config, Repo, Commit, WorkerOauth, Issue, PullRequest, PullRequestReview, ContributorsAlias,UnresolvedCommitEmail, Contributor, CollectionStatus, UserGroup, RepoGroup
 from augur.tasks.util.collection_state import CollectionState
+from augur.tasks.git.correction import correct_timestamp
```
Ulincsys: The db module must not import from any task. The db module is intended to be isolated from the rest of Augur, and should be treated as though it was installed as a package via pip (as that is the eventual goal) 

Please ensure any new code needed by the db module resides within the module scope

shlokgilda:
@Ulincsys  Moved the timestamp correction logic out of the tasks layer and into `augur/application/db/timestamp_utils.py` so it lives within the `db` module scope. `lib.py` now imports `correct_timestamp` from there instead of reaching into tasks. Moved the tests as well. They're now at `tests/test_application/test_db/test_timestamp_utils.py.`                              
                                              
The one remaining cross-layer import is `CollectionState` from `augur.tasks.util.collection_state`. That predates this PR and is flagged with a TODO. Happy to move it into the `db` module here or in a follow-up, whichever you prefer.

MoralCode:
> Moved the timestamp correction logic out of the tasks layer and into `augur/application/db/timestamp_utils.py` so it lives within the `db` module scope.

These utils feel like they may be a better fit for augur.application.utils?

shlokgilda:
I think `db` is the right home for this. The correction logic is specifically about sanitizing timestamps for PostgreSQL insertion. It's called from `lib.py` when Postgres rejects an insert. Moving it to `augur.application.util` would mean the db module imports from outside its own package, which goes against the isolation goal @Ulincsys outlined?

MoralCode:
i think the goal is mainly that tasks/workers remain independent, although it would be nice to have the various "segments" of this codebase (i.e. the boundaries we shouldnt import across) documented as well.

To me it seems like these functions are re-usable date/timestamp cleanup functions. They seem maybe generally reusable, even if we only happen to import them in this database portion of the code.

shlokgilda:
@MoralCode Moved `timestamp_utils.py` from `augur/application/db/` to `augur/application/` so it's a reusable utility rather than tied to the db layer. Import in `lib.py` updated accordingly. Tests moved to `tests/test_application/test_timestamp_utils.py` and added to CI in `pyproject.toml`.

PR #3524, opened 2026-01-07
[lint] Disable pylint rule suggesting removal of explicit else blocks

### LINKED ISSUE #3521: Change linter config to NOT suggest removing "dangling else" blocks
Lots of prs recently have been removing else blocks (especially after a raise/return) because the linter suggested it.

This makes the code a little less intuitive to read, especially for first timers because its less exolicit about the developers intent (i.e. was this intended as an alternative case, or was it intended as just a general fall-through, as one might use in a function for validation, where falling through and missing all the if blocks is the  "happy path"). This is the loose consensus of the maintainers from yesterdays call

This issue will be complete when the linter config is updated to exclude enforcement of this rule

Upanshi-Mittal:
**Description**
This PR disables the Pylint rule that suggests removing explicit `else` blocks after `return`  statements. Keeping explicit `else` branches improves readability and makes developer intent clearer, especially for new contributors.

This PR fixes #3521

**Notes for Reviewers**
- Config-only change
- Disables Pylint rule R1705 (no-else-return)
- No runtime or behavior impact

**Signed commits**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -193,3 +193,8 @@ ignore_missing_imports = true
 follow_imports = "skip"
 disallow_untyped_defs = false
 exclude_gitignore = true
+
+[tool.pylint.messages_control]
+disable = [
+    "R1705",  
+]


MoralCode:
This change introduces new config rather than changing the existing config.

It also only addresses one of the two linter codes that are mentioned as being problematic, and the underlying issue is already solved in https://github.com/chaoss/augur/pull/3533/ Closing

PR #3525, opened 2026-01-07
(fix): remove no else raise and no else return rules from .pylintrc

### LINKED ISSUE #3521: Change linter config to NOT suggest removing "dangling else" blocks

pushpitkamboj:
Fix #3521 
Removed rules [R1720](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-raise.html) and [R1705](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html) from `.pylintrc`, disabling suggestions to remove `raise` or `return` blocks.

### DIFF
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -46,8 +46,8 @@ The ``message_insights_worker`` uses a system-level package called OpenMP. You w
 
 Caching System (Redis)
 ----------------------
-* `Linux Installation <https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/install-redis-on-linux/>`__
-* `Mac Installation <https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/install-redis-on-mac-os/>`__
+* `Linux Installation <https://redis.io/docs/getting-started/installation/install-redis-on-linux/>`__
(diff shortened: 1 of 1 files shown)


docs/source/getting-started/installation.rst
MoralCode, line 52: This change is not related - can you update your PR to only include the changes to the pylintrc?

pushpitkamboj:
sorry it was my mistake, thought to raise a new clean PR which will be better to have clean history.
#3533 
will close this PR

guptapratykshh:
Hey @pushpitkamboj, for future reference, you did not need to open a new PR. It is perfectly fine to address feedback in the same PR. This keeps the conversation thread intact.

MoralCode:
Closing as superseded by #3533

Issue #3527, opened 2026-01-07
stop the smoke tests from being flaky


MoralCode:
Right now the smoke tests (labeled as end to end tests, one for docker one for podman) are designed to run augur from docker for 3 minutes and parse the output looking for specific lines in the log to validate that it starts correctly.

This generally works, but sometimes the github action fails for no application-based reason (no exception etc) and a subsequent rerun usually passes.

Based on the logs, i suspect that, sometimes, the docker container just ends up taking a little too long to close down or something after it is told to stop, causing it to exceed a timeout (probably githubs side but could be our 3 min timeout), causing a job failure

We should find why this happens and fix it so the smoke tests no longer need occasional reruns to do their job (i.e. no phantom failures)

MoralCode:
I think @pushpitkamboj 's analysis in #3530 Sounds very plausible (podman handles shutdown timeouts differently)

MoralCode:
looking at that PR (initial version didnt seem to fix the issue) the chain of events seems to be:

1. CI job starts and runs
2. job succeeds (all log lines are present that should be) and exits early from `await_all.py` with code 0 (success)
3. the job command (podman compose -f docker-compose.yml up --no-build 2>&1 | (./scripts/ci/await_all.py /tmp/regex_matches.txt && podman compose -f docker-compose.yml down)` then moves on to the compose down command, which tries to stop the container stack
4. the default timeout is short (10 seconds) and podman will return with a failure code (1) if any container takes longer than this to shut down.

this is kinda where i lose the plot though

MoralCode:
example of this happening to docker: https://github.com/chaoss/augur/actions/runs/20789115638/job/59706650405?pr=3518

shlokgilda:
I've been looking at the logs for #3518, and I'm pretty sure the root cause is that `augur backend start` runs as PID 1 in the container.

PID 1 ignores `SIGTERM` by default unless you explicitly handle it. Since `backend.py` only seems to catch `KeyboardInterrupt` right now, when podman sends the stop signal, Augur ignores it until the daemon loses patience and kills it (hence the "given PID did not die within timeout" error).

I think the cleaner fix here would be to add a `SIGTERM` handler to backend.py (rather than just bumping the timeout) that triggers the same graceful shutdown logic as `KeyboardInterrupt`.

MoralCode:
oh yeah, 100% agree.

[...]

PR #3528, opened 2026-01-08
Fix flaky smoke tests by adding SIGTERM signal handling

antcybersec:
**[backend] Fix flaky smoke tests by adding SIGTERM signal handling**

## Description

This PR fixes issue #3527 by implementing proper signal handling for graceful shutdown of the backend server. The smoke tests were flaky because containers would hang during shutdown, sometimes exceeding the 3-minute test timeout.

## Summary of Changes

1. **Added SIGTERM and SIGINT signal handlers** (`augur/application/cli/backend.py`):
   - Implemented a `shutdown_handler` function that properly terminates all subprocesses (gunicorn server, celery workers, celery beat)
   - Added timeout-based termination (5s for server, 3s for workers) before forcefully killing processes
   - Ensures proper cleanup of collection resources and pidfile removal
   - Prevents double shutdown with a `shutting_down` flag

2. **Added `stop_grace_period` to docker-compose.yml**:
   - Set to `10s` for the backend service to give Docker a clear timeout expectation
   - Helps Docker know how long to wait for graceful shutdown before killing

## Root Cause

Docker sends `SIGTERM` to containers during shutdown (`docker compose down`), but the backend only handled `SIGINT` (KeyboardInterrupt). This caused containers to hang until Docker's default 10-second kill timeout, which sometimes exceeded the 3-minute test timeout in CI.

## Testing

- The signal handlers are registered early in the startup process
- Processes are terminated gracefully with timeouts before being killed
- All cleanup operations (collection status, rabbitmq, redis, pidfile) are properly executed

## Notes for Reviewers

- The signal handler uses a `shutdown_context` dictionary to store process references that need to be cleaned up
- Timeout values (5s for server, 3s for workers) were chosen to balance quick shutdown with allowing processes to finish gracefully
- The handler is idempotent - it checks `shutting_down` flag to prevent multiple executions

**Signed commits**

### DIFF
--- a/augur/application/service_manager.py
+++ b/augur/application/service_manager.py
@@ -12,6 +12,9 @@ logger = AugurLogger("augur_servicemanager").get_logger()
 
 
 class AugurServiceManager:
+    """ Provides a storage space for references to the various components of augur
+    This enables them to all be properly shut down in the event a shutdown signal (SIGINT - AKA ctrl-c, or SIGTERM) is received 
+    """
     def __init__(self, ctx, pidfile, disable_collection):
         self.ctx = ctx
         self.pidfile = pidfile


MoralCode:
The "Testing" section of your writeup only lists things to test

Can you provide more details on what your testing process is and what the results were?

antcybersec:
> The "Testing" section of your writeup only lists things to test
> 
> Can you provide more details on what your testing process is and what the results were?

Testing
Ran `augur server start --port 5000`, then sent SIGTERM to the main PID. Logs showed the new“shutting down gracefully” message; all Gunicorn and Celery processes exited cleanly and `main.pid` was removed.

Ran `docker compose -f docker-compose.yml up --build` followed by `docker compose -f docker-compose.yml down`. The main Augur container received SIGTERM, executed the new handler, and shut down without hanging. Redis received SIGTERM, performed a clean shutdown, and exited with code 0. The `augur-keyman` helper container exited with code 137 (SIGKILL after the grace period), but this did not block shutdown; Docker completed `compose down` successfully and removed all containers and networks.
image

Verified normal startup logs (Gunicorn webserver start, worker startup messages) and observed no new errors or tracebacks introduced by the signal-handling changes.

MoralCode:
> Ran `augur server start --port 5000`, then sent SIGTERM to the main PID. Logs showed the new“shutting down gracefully” message; all Gunicorn and Celery processes exited cleanly and `main.pid` was removed.

Does this not work with SIGINT currently? AFAIK docker/podman will initially try a more graceful shutdown signal first

although now that i think about it, maybe that more graceful signal is SIGTERM and the aggressive one is SIGKILL...

Is there a spot in the code where we already handle SIGINT? Maybe we can just add a little code there to also catch and handle SIGTERM the same way?

antcybersec:
> > Ran `augur server start --port 5000`, then sent SIGTERM to the main PID. Logs showed the new“shutting down gracefully” message; all Gunicorn and Celery processes exited cleanly and `main.pid` was removed.
> 
> Does this not work with SIGINT currently? AFAIK docker/podman will initially try a more graceful shutdown signal first
> 
> although now that i think about it, maybe that more graceful signal is SIGTERM and the aggressive one is SIGKILL...
> 
> Is there a spot in the code where we already handle SIGINT? Maybe we can just add a little code there to also catch and handle SIGTERM the same way?

You're exactly right—Docker sends SIGTERM by default (and only sends SIGKILL if the process refuses to stop after the grace period).

Previously, the code only handled SIGINT (via a try/except KeyboardInterrupt block around server.wait()). The issue is that SIGTERM doesn't raise a KeyboardInterrupt exception in Python—it just terminates the process immediately if no handler is registered.

That's why I couldn't just "add a little code" to the existing except block. I had to register a proper signal.signal handler. And since I was adding that, I pointed both SIGTERM and SIGINT to the same handler to ensure that Ctrl+C (local dev) and docker stop (CI/Deployment) trigger the same robust cleanup logic (gracefully terminating subprocesses like Gunicorn and Celery so they don't become zombies

augur/application/cli/backend.py
```diff
@@ -55,17 +56,93 @@ def start(ctx, disable_collection, development, pidfile, port):
...
+        'keypub': None,
+        'disable_collection': disable_collection,
+        'engine': ctx.obj.engine,
+        'pidfile': pidfile,
+        'shutting_down': False
+    }
+
+    def shutdown_handler(signum, frame):
+        if shutdown_context['shutting_down']:
+            return
```
MoralCode: What process does the `augur backend stop` CLI command use to shut things down? is it possible to reuse that as a shutdown handler to avoid the possibility of duplicating code/processes for shutting augur down (or creating redundant processes)?

antcybersec: > What process does the `augur backend stop` CLI command use to shut things down? is it possible to reuse that as a shutdown handler to avoid the possibility of duplicating code/processes for shutting augur down (or creating redundant processes)?

I looked into reusing stop / augur_stop, but there's a key difference in context:

augur backend stop is an external command. It implicitly has to "hunt" for processes (via psutil) because it doesn't own them.
shutdown_handler runs inside the parent process, so it holds the direct subprocess.Popen handles to its children.
By using the specific handles in shutdown_handler, we get much better reliability: we can call .wait(timeout=...) on the specific child objects and escalate to .kill() if they hang. If we reused augur_stop, we'd be relying on psutil scanning (which can be flaky) and would lose that direct parent-child process control, which is exactly what was causing these flakes in the first place (docker killing the parent before children were done)

shlokgilda:
I see the discussion about reusing code vs a new handler. Since we already have a `try/except KeyboardInterrupt` block that handles graceful shutdown (stopping server, workers, beat, cleanup), could we just reuse that logic entirely?

Instead of defining a new `shutdown_handler` function, we could just register a handler that raises `KeyboardInterrupt`:
```python
def sigterm_handler(_signo, _stack_frame):
    logger.info("Received SIGTERM, shutting down...")
    raise KeyboardInterrupt
signal.signal(signal.SIGTERM, sigterm_handler)
```

antcybersec:
> I see the discussion about reusing code vs a new handler. Since we already have a `try/except KeyboardInterrupt` block that handles graceful shutdown (stopping server, workers, beat, cleanup), could we just reuse that logic entirely?
> 
> Instead of defining a new `shutdown_handler` function, we could just register a handler that raises `KeyboardInterrupt`:
> 
> ```python
> def sigterm_handler(_signo, _stack_frame):
>     logger.info("Received SIGTERM, shutting down...")
>     raise KeyboardInterrupt
> signal.signal(signal.SIGTERM, sigterm_handler)
> ```

I actually considered that, but I realized it introduces a race condition during startup. The main try/except KeyboardInterrupt block only wraps server.wait(). If Docker sends SIGTERM during the startup phase (e.g. while checking Gunicorn health or starting Celery), the KeyboardInterrupt would be raised outside that try block, causing the script to crash without running the cleanup logic (leaving zombie processes).

By using a dedicated shutdown_handler that has access to the shutdown_context, we ensure that cleanup happens safely regardless of when the signal arrives (even during startup).

augur/application/cli/backend.py
MoralCode: Ah ok. Id like to eventually have all of them be the same if possible (i.e. have the CLI command somehow also trigger the proper shutdown process in the main backend) - thats probably a subject for a new issue though.

does the stop command just send SIGINT/SIGTERM/SIGKILL like most other things?

augur/application/cli/backend.py
```diff
@@ -55,17 +56,93 @@ def start(ctx, disable_collection, development, pidfile, port):
...
+        'celery_beat_process': None,
+        'keypub': None,
+        'disable_collection': disable_collection,
+        'engine': ctx.obj.engine,
+        'pidfile': pidfile,
+        'shutting_down': False
+    }
+
+    def shutdown_handler(signum, frame):
+        if shutdown_context['shutting_down']:
```
MoralCode: im not a huge fan of this function-inside-a-function thing, is there a way? maybe we maybe move it to the top level and pass the shtdown context object into it as a parameter? not sure though

augur/application/cli/backend.py
```diff
@@ -55,17 +56,93 @@ def start(ctx, disable_collection, development, pidfile, port):
...
+                os.unlink(shutdown_context['pidfile'])
+            except OSError:
+                pass
+
+        sys.exit(0)
+
+    # Register signal handlers for graceful shutdown
+    signal.signal(signal.SIGTERM, shutdown_handler)
+    signal.signal(signal.SIGINT, shutdown_handler)
+
```
MoralCode: it feels a little weird having these just exist in the CLI  - even though i suspect that the CLI is whats used for the actual startup too, not just user commands.

Is all of this likely to handle the backend start being called multiple times (i.e accidentally)?

antcybersec: hey @MoralCode thanks for the feedback! I've refactored the logic into a top-level AugurServiceManager
class to avoid the nested function and clean up state management.

I stuck with the explicit signal handler (instead of raising KeyboardInterrupt) because the try/except block in the main loop isn't active during startup. If Docker sends SIGTERM while Gunicorn is still initializing (which happens often in CI), raising KeyboardInterrupt would crash the script without cleanup.

also confirmed augur runs as PID 1 via init.sh, so the explicit signal.signal registration is definitely required!

augur/application/cli/backend.py
```diff
@@ -180,29 +263,15 @@ def start(ctx, disable_collection, development, pidfile, port):
...
+    finally:
+        # Ensure pidfile is cleaned up if we exit normally
+        if os.path.exists(pidfile):
             try:
-                keypub.shutdown()
-                cleanup_collection_status_and_rabbit(logger, ctx.obj.engine)
-            except RedisConnectionError:
+                os.unlink(pidfile)
+            except OSError:
                 pass
```
MoralCode: It may be nice to log an error if this happens

docker-compose.yml
```diff
@@ -57,6 +57,7 @@ services:
       - facade:/facade:rw
       - logs:/logs:rw
     restart: unless-stopped
+    stop_grace_period: 10s
```
MoralCode: This is the default value right? IMO we should only specify it if it is different

augur/application/cli/backend.py
antcybersec: I’ll add a log there (e.g. logger.warning or logger.debug) when the pidfile can’t be removed so we don’t swallow it silently.

docker-compose.yml
antcybersec: > This is the default value right? IMO we should only specify it if it is different

Yes, 10s is the default. I’ll remove the line and we can add stop_grace_period later only if we want a different value.

augur/application/cli/backend.py
MoralCode: id say do logger.error so its pretty obvious

augur/application/cli/backend.py
sgoggins: I'll try this one out locally after I do a fresh Ubuntu install on a laptop. This one looks like an improvement, but I need to test bare metal and docker cases.

augur/application/cli/backend.py
```diff
@@ -104,8 +104,8 @@ def shutdown_signal_handler(self, signum, frame):
         if os.path.exists(self.pidfile):
             try:
                 os.unlink(self.pidfile)
-            except OSError:
-                pass
+            except OSError as e:
+                logger.warning("Could not remove pidfile %s: %s", self.pidfile, e)
```
MoralCode: pylint is complaining about too many arguments. Please use a proper python f-string for this message and have it be logged at the `error` level, not `warning`

augur/application/cli/backend.py
```diff
@@ -270,8 +270,8 @@ def start(ctx, disable_collection, development, pidfile, port):
         if os.path.exists(pidfile):
             try:
                 os.unlink(pidfile)
-            except OSError:
-                pass
+            except OSError as e:
+                logger.warning("Could not remove pidfile %s: %s", pidfile, e)
```
MoralCode: same thing here, use an f-string and log this at the error level

MoralCode (changes requested):

MoralCode:
Rebased
Moved the new class to its own file under augur/application
brought over a small collection of interrelated shutdown-related functions into this file too (due to dependencies from the class) and re-imported them back into backend.py as needed

If the podman test fails to stop properly after this pr, im gonna be so annoyed lol

MoralCode:
i believe the pylint failure on this PR is benign/ignore-able and can be dealt with as we generally clean up our code style. I don't want to bloat the scope of a PR like this that affects some core-ish functionality (also given our pylint settings are overly broad #3657)

MoralCode:
that said, given the pylint scoping PR was merged, we can rebase this and that should cause pylint to settle down. at the cost of using a lot more API credits to rerun every job

PR #3531, opened 2026-01-08
fix(ci): resolve Podman E2E cleanup timeout failure

### LINKED ISSUE #3530: Podman E2E cleanup timeout failure
Podman Compose uses a default 10-second timeout for podman compose down, which is insufficient for Augur's multi-process architecture to shut down gracefully. The actual tests pass successfully, all 12 log patterns are matched but the CI job fails during container cleanup.
See #3522 

check line 280-288 inside `workflows/build_docker.yml`

Image

pushpitkamboj:
Podman Compose uses a default 10-second shutdown timeout, which is insufficient for Augur's multi process architecture to terminate gracefully. Unlike Docker, podman exits with an error instead of force-killing containers when the timeout is exceeded. see [docs](https://docs.podman.io/en/latest/markdown/podman-stop.1.html)

resolves #3530 

Solution
- Added -t 60 flag to podman compose down to allow 60 seconds for graceful shutdown
- Increased timeout minutes from 3 to 5 to accommodate the extended shutdown window

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -285,8 +285,8 @@ jobs:
         run: |
           podman compose -f docker-compose.yml up --no-build 2>&1 \
             | (./scripts/ci/await_all.py /tmp/regex_matches.txt \
-                && podman compose -f docker-compose.yml down -t 60)
-        timeout-minutes: 5
+                && podman compose -f docker-compose.yml down -t 120)
+        timeout-minutes: 7
         env:
           AUGUR_GITHUB_API_KEY: ${{ secrets.GITHUB_TOKEN }}
           AUGUR_GITHUB_USERNAME: ${{ github.repository_owner }}


MoralCode:
> Unlike Docker, Podman exits with an error when containers exceed the default 10-second shutdown timeout instead of force-killing them.

This comment from copilot seems... wrong given the docs say

> --time, -t=seconds
>
> Seconds to wait before forcibly stopping the container. Use -1 for infinite wait.

MoralCode:
Im not sure this is quite the right solution because the CI is still failing on this PR with very similar symptoms:

```
 ✅✅✅ MATCH pattern 11: 'Done generating scc data for repo'
✅✅✅✅✅ All patterns matched. ✅✅✅✅✅
>>>> Executing external compose provider "/usr/libexec/docker/cli-plugins/docker-compose". Please refer to the documentation for details. <<<<

time="2026-01-08T11:02:37Z" level=warning msg="The \"AUGUR_FLAGS\" variable is not set. Defaulting to a blank string."
 Container augur-augur-1  Stopping
 Container augur-augur-1  Error while Stopping
Error response from daemon: given PID did not die within timeout
Error: executing /usr/libexec/docker/cli-plugins/docker-compose -f docker-compose.yml down -t 60: exit status 1
Error: Process completed with exit code 1.
```

pushpitkamboj:
No actually if you do not provide the flag, the default is 10 seconds, 
checkout the description
`The --time option specifies the number of seconds to wait before forcibly stopping the container after the stop command is issued to the container. The default is 10 seconds.`
used copilot for structuring my thoughts :)

https://github.com/containers/podman-compose/issues/154
this maybe useful for this conversation

pushpitkamboj:
that said, I agree that this may not be the right solution for the problem, I will figure out more on this.

MoralCode:
Closing - increasing the timeout is not likely to solve the underlying issue

PR #3523, opened 2026-01-07
Add configurable batch sizes for GitHub data collection tasks

shlokgilda:
**Description:**
Normalizes hardcoded batch size magic numbers to named constants at the top of each file. This is a pure refactoring PR with no functional changes - all values remain the same.

**Changes:**
- Added named constants like `ISSUE_BATCH_SIZE`, `EVENT_BATCH_SIZE`, `PR_BATCH_SIZE`, etc.

**Note:**
Based on feedback in #3515, doing this in two PRs:
1. This PR: Normalize constants (easy to review, no risk)
2. Follow-up PR: Add hybrid config with 3 values instead of 9

**Signed commits**
- [x] Yes, I signed my commits.

**AI Disclosure Warning:** This PR description draft was generated using Claude Code.

PR #3482, opened 2025-12-19
Fix repo_deps_libyear column typo: current_verion -> current_version

shlokgilda:
**Description**                                                                                                                     
- Renames misspelled `current_verion` column to `current_version` in `repo_deps_libyear` table                                      
- Adds Alembic migration (rev 38) that drops/recreates the `explorer_libyear_detail` materialized view                              
- Updates ORM model, task code, and all schema files for consistency                                                                
- Follow-up to #2716 which was closed because it lacked the migration                                                               
                                                                                                                                    
This PR fixes #2662                                                                                                                 
                                                                                                                                    
**Notes for Reviewers**                                                                                               
@sgoggins mentioned in #2716 that renaming might require temp tables and data copying. This isn't necessary for PostgreSQL - column renames are simple metadata operations (`ALTER TABLE ... RENAME COLUMN`). The only complexity is the materialized view that references the column, which we drop and recreate with the corrected name. No data copying involved.                                
                                                                                                                                    
**Signed commits**                                                                                                                  
- [x] Yes, I signed my commits. 

**AI Disclosure:** Used Claude Code to write this PR draft, comments in the alembic migration and verify the migration.

PR #3532, opened 2026-01-08
Added accepted change requests count metric

### LINKED ISSUE #2627: Change Requests Accepted metric API
The canonical definition is here: https://chaoss.community/?p=3589

Khusi2005:
Added an implementation for the "Accepted Change Requests" metric based on the CHAOSS definition.The metric aligns with existing PR metrics and supports filtering by time period, repository, submitter, and merger.

**Description**
- Implements a new registered metric that counts accepted (merged) pull requests using Augur’s ingested GitHub data.
- Supports optional filters for date range, repository, repository group, PR submitter, and PR merger.
- Follows existing query patterns used in `pull_request.py` to ensure consistency with current metrics.
This PR fixes #2627

**Notes for Reviewers**
- This change has been tested locally.
- Feedback and suggestions are welcome.

**Signed commits**

AI Disclosure Warning: This PR description draft was generated using ChatGPT.


augur/api/metrics/pull_request.py
```diff
@@ -1226,7 +1226,88 @@ def pull_request_merged_status_counts(repo_group_id, repo_id=None, begin_date='1
...
+    
+    # Merger filter (who merged PR)
+    
+    if merger_id:
+        sql += """
+            AND pr.pull_request_id IN (
+                SELECT ie.issue_id
+                FROM issue_events ie
+                WHERE ie.action = 'merged'
+                  AND ie.cntrb_id = :merger_id
```
guptapratykshh: Hi @Khusi2005, I found a small bug in the `Merger Filter`. In Augur, pull_request_id and issue_id are different. So, using issue_events here won't work. We should use the `pull_request_events` table instead.

Here is the fix - 
```
      SELECT pre.pull_request_id
      FROM pull_request_events pre
      WHERE pre.action = 'merged'
        AND pre.cntrb_id = :merger_id
```

Khusi2005: Thank you for the feedback. I've updated the merger filter to use the pull_request_events table instead of issue_events. The fix is now pushed.

shlokgilda:
The overall approach is solid and aligns with the metric definition. However, there's one critical issue that needs to be fixed before this can be merged.

**Critical Issue:** The column `pr_src_author_id` doesn't exist in the `pull_requests` table. This will cause a SQL error at runtime.

```diff
- sql += " AND pr.pr_src_author_id = :submitter_id"
+ sql += " AND pr.pr_augur_contributor_id = :submitter_id"
```
The correct column is `pr_augur_contributor_id` (see augur/application/db/models/augur_data.py:1734).

**Minor Issues:** 
1. Missing spaces after commas in function signature:
`def accepted_change_requests_count(repo_group_id, repo_id=None, begin_date=None, ...)`
2. Some lines have trailing whitespace that should be cleaned up
3. Docstring clarity - The parameters should clarify that `submitter_id` and `merger_id` expect Augur's internal `cntrb_id` values, not GitHub user IDs

**Testing Required:** I noticed in the PR description that this was not tested locally. Please test this against an actual Augur instance with real data. Please:
1. Set up a local Augur instance using docker
2. Test the endpoint against repos that have collected PR data
3. Share the results/output

Example test commands:
```
curl "http://localhost:5000/api/unstable/repos/<repo_id>/accepted-change-requests-count"
curl "http://localhost:5000/api/unstable/repo-groups/1/accepted-change-requests-count?begin_date=2020-01-01"
```

Questions
1. Performance: Could you benchmark how long the query takes on a reasonably sized dataset? The `merger_id` subquery against `pull_request_events` could be slow on large repos.
2. AI Disclosure: Was any AI tooling (GitHub Copilot, ChatGPT, Claude, etc.) used to write this code? If so, please add an AI Disclosure to the PR description - see #3482 and #3523 for examples of how we document this.

Khusi2005:
All minor issues have been fixed. The metric has also been tested locally against a real Augur instance with collected data and verified using both the API and PostgreSQL. The API results match the database counts exactly.
**Performance**
I benchmarked the query using EXPLAIN ANALYZE on the local dataset. Since the dataset is small, PostgreSQL correctly prefers a sequential scan and executes the query in a few milliseconds. I also added a partial index on (repo_id, pr_merged_at), which enables an index-only scan and will scale efficiently on large Augur deployments.

**Testing Results**
https://docs.google.com/document/d/1JuDfhPdEbB6XHmNQZUCpSOxy6D3CZ7KsOvT2OQrASLk/edit?usp=sharing

AI Disclosure
This PR description draft was generated using ChatGPT.

PR #3336, opened 2025-10-23
remove unused dependencies

MoralCode:
**Description**
This change:
1. removes several dependencies that were seemingly never used
2. relocates a couple of dependencies to either the devdependencies area, or removes them since they are transitive dependencies

This PR is a major step towards resolving #3334 (and in future, updating dependencies)

**Notes for Reviewers**

**Signed commits**
- [x] Yes, I signed my commits.

I should probably squash these commits

PR #3534, opened 2026-01-09
fix(#3519): update tensorflow and dependencies to support python 3.12+

### LINKED ISSUE #3519: Dependencies wont install on Python 3.12+ due to tensorflow version
The tensorflor dependency needs updating or something to support Python 3.12+ as we currently have breakage (discovered in #3513 because of the unit tests) due to our pinned tensorflow verison likely not supporting these python versions.

Its been long enough since the releases of Python 3.12 and 3.13 (at least), that I would assume there are newer versions available that we could update to - we just need to do it and test it

guptapratykshh:
**Description**
Installation fails on Python 3.12 and above because the pinned version of TensorFlow (2.15.0) does not provide pre-built wheels for Python 3.12. Additionally, scikit-image (0.19.1) fails to build because it depends on distutils, which was removed in Python 3.12.

Updated project dependencies to support Python 3.12 and above, has been done locally:

- TensorFlow: Bumped from 2.15.0 to >=2.16.1 to access Python 3.12 compatible wheels.
- Scikit-Image: Bumped from 0.19.1 to >=0.22.0 to fix ModuleNotFoundError: No module named 'distutils'.
- Protobuf: Removed the <3.22 constraint to allow newer versions needed by the updated TensorFlow.
- Pandas: Relaxed the strict pin (==1.5.3 to >=1.5.3) to resolve dependency conflicts.

This PR fixes #3519

**Notes for Reviewers**

**Signed commits**


pyproject.toml
```diff
@@ -57,9 +57,9 @@ dependencies = [
     "mistune",
     "nltk==3.6.6",
     "numpy==1.26.0",
-    "pandas==1.5.3",
+    "pandas>=1.5.3",
     "partd>=0.3.10",
-    "protobuf<3.22",
+    "protobuf",
```
MoralCode, line 53: whats the current latest version of this? is this going to cause us to get put on a new major version?

Id also be curious why we were maxing out at 3.22 before.

guptapratykshh: The latest protobuf version is 6.33.2. However, removing the <3.22 cap allows resolver to choose version 4.25.8, which is compatible with tensorflow 2.17.1. The old cap was probably set because tensorflow 2.15.0 used protobuf 3.20.3. The <3.22 constraint aimed to avoid breaking changes from protobuf 4.x. But newer tensorflow 2.17.1 requires protobuf 4.x+.

MoralCode: oh wow, thats a lot of details on the version history

MoralCode:
in this PR, can you also re-enable the 3.12 unit tests so we can make sure they are passing too?

pyproject.toml
MoralCode: I suspect this may need to be done together with and/or after #3336 - a lot of our dependencies are probably transitive, like this protobuf one

MoralCode:
Going to mark as waiting until #3336 merges

MoralCode:
The PR has merged. @guptapratykshh please rebase this and fix the conflicts so it can be merged

also make sure 3.12 is enabled in the unit tests (i.e. the github actions matrix that runs the tests on different python versions

MoralCode:

Thanks!

MoralCode:
The unit tests do not run on python 3.12.

MoralCode:
I have started experimenting with this in a separate branch

MoralCode:
Closing due to previous comment and unresponsive contributor (per their other PRs)

PR #3549, opened 2026-01-10
Refactor CONFIG_LOCATION to use CONFIG_DATADIR #3547

### LINKED ISSUE #3547: Refactor the conflg location variable to use the new config directory value instead
https://github.com/chaoss/augur/blob/f895f0742c124743fc036bf69bf87ae4a21df210/docker-compose.yml#L75-L76

#3314 introduced a variable that allows augur config to be read from an `augur.json` file in a configured directory

other parts of augurs codebase have used other config files for other things previously

this issue is about refactoring those older uses to work with the new "config folder" setting and avoid unnecessarily duplicated variables

guptapratykshh:
**Description**
I have changed how the configuration file path is resolved to match the new config directory standards. Instead of using separate CONFIG_LOCATION variable, the system now builds the path using CONFIG_DATADIR. This simplifies docker-compose.yml file by getting rid of unnecessary variables while maintaining backward compatibility.

This PR fixes #3547.

**Notes for Reviewers**

**Signed commits**


augur/api/view/init.py
```diff
@@ -7,7 +7,13 @@
 env = Environment()
 
 # load configuration files and initialize globals
-configFile = Path(env.setdefault("CONFIG_LOCATION", "config.yml"))
+config_location = env["CONFIG_LOCATION"]
```
MoralCode: i dont like the square bracket syntax on env. It seems like this is liable to crash if CONFIG_LOCATION isnt set. Also CONFIG_LOCATION was removed in this PR, so its more likely to be unset

augur/api/view/init.py
```diff
@@ -7,7 +7,13 @@
...
+if config_location:
+    configFile = Path(config_location)
+elif env["CONFIG_DATADIR"]:
+    configFile = Path(env["CONFIG_DATADIR"]) / "config.yml"
+else:
+    configFile = Path("config.yml")
```
MoralCode, line 15: Did you choose to have three branches of if/else for backwards compatibility?

augur/api/view/run.sh
```diff
@@ -1,4 +1,4 @@
-export CONFIG_LOCATION="config.yml"
+export CONFIG_DATADIR="."
```
MoralCode, line 1: im unsure about defaulting this to the current directory. would want to understand where this file is run first.

@sgoggins is this part of a manual install?

augur/api/view/init.py
guptapratykshh: I have updated code to use os.getenv('CONFIG_LOCATION') which returns None safely if variable is unset.

augur/api/view/init.py
guptapratykshh: Yes,exactly. The first branch supports legacy CONFIG_LOCATION variable if it is still present, second handles new CONFIG_DATADIR standard, and third provides safe default fallback.

augur/api/view/run.sh
guptapratykshh: Yes, this preserves the exact previous behavior. Since old script exported config.yml (implicitly in the current directory), setting CONFIG_DATADIR='.' ensures new resolution logic still finds ./config.yml in that same location."

MoralCode:

Thanks!

PR #3554, opened 2026-01-11
Add .gitattributes to enforce LF line endings for source files

1steve78:
This PR adds a .gitattributes file to explicitly enforce LF line endings
for Python, shell, and other source files.

This prevents Windows-based Git configurations from automatically
rewriting line endings, which has caused CI and runtime issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,17 @@
+# Enforce LF line endings for source files
+*.py text eol=lf
+*.sh text eol=lf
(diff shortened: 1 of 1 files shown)


1steve78:
The failure appears to be in the Docker build step while downloading Geckodriver and seems unrelated to this change, which only affects Git line-ending handling. Happy to re-run or adjust if needed.

PR #3556, opened 2026-01-11

### LINKED ISSUE #3545: Add .gitattributes file to define how line endings should be handled

Sukuna0007Abhi:
Fixes #3545

### Summary
This PR adds a .gitattributes file to the repository to prevent Windows CRLF line ending issues that can break Python scripts and shell script entrypoints.

### Changes

- Added .gitattributes file with rules to enforce LF line endings for:
- Python files (*.py)
- Shell scripts (*.sh)
- All text files (default * text=auto)
- Marked binary files appropriately to prevent line ending conversion

### Testing
The .gitattributes file follows GitHub's recommended format for handling line endings. This will prevent Git from automatically converting LF to CRLF on Windows systems, which has been causing issues in WSL and other Windows-based environments.

### Related Issues
Resolves #3545

Pls review @sgoggins @MoralCode @cdolfi

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,99 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+# Force LF line endings for all text files
+*.py text eol=lf
+*.sh text eol=lf
+*.bash text eol=lf
+*.fish text eol=lf
+*.zsh text eol=lf
+
+# Configuration files
+*.ini text eol=lf
+*.cfg text eol=lf
+*.conf text eol=lf
+*.config text eol=lf
+*.toml text eol=lf
+*.yaml text eol=lf
+*.yml text eol=lf
+*.json text eol=lf
+
+# Documentation
+*.md text eol=lf
+*.rst text eol=lf
+*.txt text eol=lf
+
+# SQL files
+*.sql text eol=lf
+
+# Web files
+*.html text eol=lf
+*.htm text eol=lf
+*.css text eol=lf
+*.scss text eol=lf
+*.js text eol=lf
+*.jsx text eol=lf
+*.ts text eol=lf
+*.tsx text eol=lf
+*.vue text eol=lf
+
+# Template files
+*.j2 text eol=lf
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,99 @@
...
+*.rar binary
+*.exe binary
+*.dll binary
+*.so binary
+*.dylib binary
+*.pyc binary
+*.pyo binary
+*.pyd binary
+*.whl binary
+*.egg binary
```
MoralCode, line 99: a lot of these (from .exe down) are unlikely to be in the repo since many of them are compiled program outputs and wouldnt be checked into git

.gitattributes
```diff
@@ -0,0 +1,99 @@
...
+*.env text eol=lf
+*.sample text eol=lf
+LICENSE* text eol=lf
+CONTRIBUTING* text eol=lf
+CHANGELOG* text eol=lf
+README* text eol=lf
+
+# Binary files - ensure they're not modified
+*.png binary
+*.jpg binary
```
MoralCode, line 80: does the automatic behavior account for this already? whats the default behavior?

.gitattributes
```diff
@@ -0,0 +1,99 @@
...
+# Web files
+*.html text eol=lf
+*.htm text eol=lf
+*.css text eol=lf
+*.scss text eol=lf
+*.js text eol=lf
+*.jsx text eol=lf
+*.ts text eol=lf
+*.tsx text eol=lf
+*.vue text eol=lf
```
MoralCode, line 38: do we need to specify every single file type like this given that we assume things are ttext files in the first line?

shlokgilda:
@Sukuna0007Abhi There are a couple of other PRs targeting the same issue (#3561, #3562). Could you all please discuss the best way forward and come up with just one PR and close the rest?  You can discuss the optimal way forward in the linked issue #3545.

MoralCode:
Closing in favor of another duplicate PR that solves this issue with a smaller, more targeted change

Issue #3485, opened 2025-12-19
Add an augur system verification/diagnostic/self test script


MoralCode:
I think it would help for augur to provide a verification script for people to optionally run on their installations to automatically help detect and flag potential problems before they arise.

I think this should take the form of a new CLI command and should check both things that should already be true about an augur install, and also maybe pre-emptively check and warn for things that may affect future/planned migrations.

**Basic informational checks:**
_Basic information about your augur instance/setup_
- report current DB  alembic version, postgres version, checked out commit of augur and whether docker is in use or not
- are all migrations applied? 

**basic operational checks:**
_are important baseline expectations for operation working okay?_
 - are filesystem permissions correct for things like facade and logs (possibly with an attempt to automatically fix in leiu of a post-start hook, see #3212 )
 - Check that all git repos in facade have no changes made to them (clean working trees)
 - check that the ratio of commits with no linked contributor is within acceptable parameters.

**Benign Data Cleanliness:**
_not issues, just things that could use cleaning up_
  - check for duplicate aliases for a contributor (same email present many times)

**Bug Mitigation/regression prevention**:
_issues from previous bugs that may still affect older instances if cleanup was not performed_
  - are there any duplicate repos (i.e. multiple entries in the repos table with the same repo_src_id?
  - [ ] detect duplicate issue urls (#3457, see below query)
  - [ ] detect bad commits table data (from the empty name bug)
  - detect unreasonable numbers of aliases per contributor (resolution bugs as a result of the prior one)
      - special spot checks for “dave” “andy” etc common names that were also being mis-connected
  -  count contributors with null created_at dates (from the resolution bugs of late 2025 early 2026 - there shouldnt be many as theres an automatic task to correct these). if this doesnt go down within a few hours for no good reason, there may be an issue

**Future-thinking checks:**
_planned schema or other major changes we could proactively check for:_
- Check that all dates in the columns affected by #2662 are in an expected format (for future migration)
- check that all `gl_` prefixed columns in the contributors table are null (for future removal/merge into generic platform tables
- check for differences between contributors table cntrb_email and cntrb_canonical for future merge of those columns
- check for cntrb_canonical values in the email aliases table that are different/not present as actual aliases (prepare for column deletion)
- print warnings about misspelled column names potentially changing in future (#2662)

**Consistency checks:**
_sanity checks not born from any particular issue, but intended to act as canaries when stuff goes wrong_
 - make sure collection dates are all after author/committer date in commits table to the self check script 

[...]

PR #3557, opened 2026-01-11
Remove recursive flag from chmod in docker-compose issue 3306

### LINKED ISSUE #3306: facade cloned repositories in dirty state
When restarting augur, it is observed that all the permissions in cloned facade repos get set to `777`. This was observed via `git status` while investigating a hang with facade. See the [Slack thread](https://chaoss-workspace.slack.com/archives/C0226ELG6R4/p1760464207697009).

The line responsible appears to be
https://github.com/chaoss/augur/blob/7f58fbc8075d228c9d427f86e577af31e4a49520/docker-compose.yml#L89

this post_start section was only added for improving openshift support.

to reslove this issue:
- [ ] create a PR that removes these lines.
- [ ] create a new issue that documents the need to re-introduce these lines in a way that fixes the original permissions issue with running in openshift, but also doesnt cause this issue with non-openshift installs.

CC @JohnStrunk

Sukuna0007Abhi:
Fixes #3306
Related to #3212

### Problem
When Augur restarts, the chmod -R 777 command recursively changes permissions on all files in /facade, /logs, /config, and /cache directories. This causes all files in cloned git repositories under /facade to have their permissions changed to 777, putting them in a dirty state where git sees all files as modified.

### Solution ( as sir @JohnStrunk proposed)
Removed the -R (recursive) flag from the chmod command. Now it only changes permissions on the directories themselves, not all files within them. This is sufficient for the container user to access these directories while preserving the original permissions of files inside them.

### Changes
Changed chmod -R 777 /facade /logs /config /cache to chmod 777 /facade /logs /config /cache
Testing
As noted by sir @JohnStrunk in the issue, this should work fine as long as the directories are empty on the first run of the container, which is the normal case.

Next Steps
I will work on #3212  next to address the broader podman compose compatibility issues.

Fixes #3306 

Pls review sir @sgoggins @MoralCode

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,7 +87,7 @@ services:
     user: 2345:2345  # Run as an arbitrary non-root user
     post_start:
       # Make sure the user has access to the volumes
-      - command: chmod -R 777 /facade /logs /config /cache
+      - command: chmod 777 /facade /logs /config /cache
         user: root
   
   # Flower is a UI that helps more easily monitor running tasks for celery workers.


MoralCode:
This recursive flag is needed to adjust permissions inside some of the other directories. I can see why this could be an issue for Facade, but I don't think this is the best solution to the issue. I shared some other ideas on slack about how we can maybe use #3485 for a similar thing

MoralCode:
After reading the discussion in #3306 i see that i am mistaken in that assessment. In light of the new info, this is actually a really good, tightly focused PR!

Sukuna0007Abhi:
Haha @MoralCode very much thanks , I actually started crying 😭, thought a first day raised pr in this great organisation and it closed in same day.. thank you very much...

Sukuna0007Abhi:
Pls review @sgoggins

shlokgilda (approved):
LGTM

PR #3559, opened 2026-01-11
Set HOME variable in containers to avoid a permissions error when git persists configuration

guptapratykshh:
**Description**
this solves docker startup race conditions by adding a database health check and setting HOME=/tmp for handling permissions.

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -79,6 +79,7 @@ services:
       - CACHE_DATADIR=/cache
       - CACHE_LOCKDIR=/cache
       - CELERYBEAT_SCHEDULE_DB=/tmp/celerybeat-schedule.db
+      - HOME=/tmp
     depends_on:
       - augur-db
       - redis


docker-compose.yml
```diff
@@ -14,6 +14,11 @@ services:
       - augurpostgres:/var/lib/postgresql/data
     networks:
       - augur
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${AUGUR_DB_USER:-augur} -d augur"]
```
MoralCode: is there an environment variable that allows the value of the database augur uses to be configured? if so we should use that for the -d argument

docker-compose.yml
```diff
@@ -77,8 +82,10 @@ services:
...
       - CACHE_LOCKDIR=/cache
       - CELERYBEAT_SCHEDULE_DB=/tmp/celerybeat-schedule.db
+      - HOME=/tmp
     depends_on:
-      - augur-db
+      augur-db:
+        condition: service_healthy
       - redis
       - augur-keyman
       - rabbitmq
```
MoralCode, line 87: I think one of the other PRs for this also had changes to these lines to specifiy that they could start as long as the DB was up (before it becomes healthy).

is that the default behavior?

docker-compose.yml
guptapratykshh: this is the one have added there also- AUGUR_DB_NAME:-augur

docker-compose.yml
guptapratykshh: Yes, service_started is default behavior when no condition is specified. I have updated docker-compose.yml
 to explicitly state condition: service_started for Redis, Keyman, and RabbitMQ to avoid any ambiguity, while keeping the strict service_healthy check for the database to prevent race condition.

MoralCode:
Superseded by #3613

MoralCode:
This doesnt solve the linked issue by itself, but still may be important to merge

MoralCode:

Thanks!

MoralCode:
Im taking over and repurposing this issue to just contain the HOME fix because the rest of the fix is already present in https://github.com/chaoss/augur/pull/3613

MoralCode:
@ABrain7710 Could this complement your work on facade credentials? I think this particular issue relates to the creation of .gitconfig, which i think git defaults to using the homedir for. maybe we should redirect this to the config dir to persist it?

MoralCode:
https://stackoverflow.com/a/6172338 (symlinking) may also be a potential option. Part of the issue is that our custom user in docker land doesnt have their own homedir.

PR #3560, opened 2026-01-11

### LINKED ISSUE #3545: Add .gitattributes file to define how line endings should be handled

arsingla786:
**Description**
Adds a `.gitattributes` file to define consistent line ending behavior
and prevent CRLF-related issues on Windows + docker  systems while setup.

This PR fixes #3545 

**Notes for Reviewers**

Users running Augur on Windows (especially with Docker) can encounter
errors when files are rewritten with CRLF instead of LF. This change
ensures Python and shell files keep Unix-style line endings.

**Signed commits**


guptapratykshh:
i think you pushed all the changes that were there on the local regarding the fix of the docker

arsingla786:
Oh yes i forgot to save those and pushed together , but fixed that now

PR #3562, opened 2026-01-11

### LINKED ISSUE #3545: Add .gitattributes file to define how line endings should be handled

arsingla786:
**Description**
Adds a `.gitattributes` file to define consistent LF line endings and
prevent CRLF-related issues on Windows systems.

This PR fixes #3545

**Notes for Reviewers**
CRLF line endings on Windows caused issues when running Augur setup in
windows + Docker environments. This change ensures consistent LF endings (tested successful setup) .

**Signed commits**


.gitattributes
MoralCode, line 1: I dont think this filename is correct for this to be detected as a gitattributes file

arsingla786: i changed the file name

shlokgilda:
@arsingla786  There are a couple of other PRs targeting the same issue (#3556, #3562). Could you all please discuss the best way forward and come up with just one PR and close the rest?  You can discuss the optimal way forward in the linked issue #3545.

PR #3580, opened 2026-01-15
Replace subprocess.call() with check_call() in db.py to prevent silent failures

### LINKED ISSUE #3579: Augur DB CLI: create schema should use check_call and fail if there is an error.
> 2. Silent failure in `augur/application/cli/db.py`:
> ```python
> def create_schema():
>     call(["alembic", "upgrade", "head"])  # uses call(), not check_call()
> ```
> `subprocess.call()` returns the exit code but doesn't raise on failure - so the script happily continues even when Alembic fails.

 _Originally posted by @shlokgilda in [#3451](https://github.com/chaoss/augur/issues/3451#issuecomment-3635604818)_

guptapratykshh:
**Description**
I replaced subprocess.call with check_call in the database cli, ensuring that operations stop immediately and report errors if something goes wrong.

This PR fixes #3579

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/cli/db.py
+++ b/augur/application/cli/db.py
@@ -298,7 +298,7 @@ def upgrade_db_version():
 @test_db_connection
 def check_for_upgrade():
     """
-    Upgrade the configured database to the latest version
+    Show available database migration history
     """
     check_call(["alembic", "history", "-i"])
 


augur/application/cli/db.py
```diff
@@ -300,7 +300,7 @@ def check_for_upgrade():
     """
     Upgrade the configured database to the latest version
```
shlokgilda: Could you fix this docstring here as well: `Show available database migration history`

shlokgilda (dismissed):
LGTM. Love the short, targeted PR. Added one suggestion to fix a docstring that was incorrect even before this change.

augur/application/cli/db.py
guptapratykshh: done

sgoggins:
Waiting on Podman end to end test. The fundamental difference here is really the provision of detailed error messages in `check_call` both wait for a process to finish ... both return 0 if its all good ... I am not against this change, it seems better, but I think it will provide more information on failure rather than allow failure more quickly ...

MoralCode (approved):
`check_call` will raise an exception if something fails.
the way we use `call` here currently (not storing the results in a variable) effectively means we ignore failures

While this may lead to more crashes, they are crashes that would otherwise be maskes, so knowing about them is better

LGTM

Issue #3486, opened 2025-12-19
Dont use hardcoded version numbers in DB inserts


MoralCode:
There are many places where augur  writes a hardcoded version number to the database:

https://github.com/chaoss/augur/blob/a60ddeaada89a5708e162600bec884e635f1de0e/augur/tasks/git/dependency_tasks/core.py#L129
https://github.com/chaoss/augur/blob/a60ddeaada89a5708e162600bec884e635f1de0e/augur/tasks/git/dependency_libyear_tasks/core.py#L55
https://github.com/chaoss/augur/blob/a60ddeaada89a5708e162600bec884e635f1de0e/augur/tasks/git/util/facade_worker/facade_worker/analyzecommit.py#L191
https://github.com/chaoss/augur/blob/a60ddeaada89a5708e162600bec884e635f1de0e/augur/tasks/github/pull_requests/commits_model/core.py#L68
https://github.com/chaoss/augur/blob/a60ddeaada89a5708e162600bec884e635f1de0e/augur/tasks/github/repo_info/core.py#L226

... and many more.

some of these are clearly out of date. Others are so old that I suspect they may be referring to versions other than the augur core version.

For this issue: any of these that are supposed to refer to augur version numbers should import/read this value from `metadata.py` (and maybe it should move into the augur source tree)

himasree9:
This PR covers Phase 1 by replacing a small set of hardcoded tool_version values with imports from metadata.py.

After this PR is reviewed, I’d be happy to follow up with a Phase 2 PR to address the remaining occurrences, if that sounds reasonable to the maintainers.

guptapratykshh:
taking on this issue ,since no changes seen recently

himasree9:
Hi @guptapratykshh, I actually have a PR open for Phase 1 of this issue. My plan was to submit Phase 2 immediately after Phase 1 is merged to keep the review process manageable. Are you looking to help with Phase 2, or should we coordinate so we aren't duplicating effort?

guptapratykshh:
yes , this works as I have made changes as the hardcoded values were scattered around all the files, please can you check the PR

himasree9:
Thanks for the update!

Since Phase 1 is still under review, I think it’s best if we wait for it to merge before moving forward with Phase 2, just to avoid conflicts or duplicated changes. Once Phase 1 is merged, we can align on what’s left and decide the best way to proceed. I’ll definitely take a look at your PR in the meantime and leave any relevant feedback there.

[...]

PR #3587, opened 2026-01-16
Fix(messages): set per-message tool_source for issue/PR comments (fix 2545)

### LINKED ISSUE #2545: BUG: tool_source meta data in message table incorrect for issue messages
Issue comments are being tagged as PR comments. Example on the augur repository: 

 

also the result of `SELECT DISTINCT tool_source FROM message m WHERE repo_id = 36113` (36113 is https://github.com/chaoss/augur in our db) is: 

Pr comment task and Pr review comment task

Sukuna0007Abhi:
Fixes #2545

**Signed commits**

### DIFF
--- a/augur/tasks/github/messages.py
+++ b/augur/tasks/github/messages.py
@@ -136,7 +136,6 @@ def process_large_issue_and_pr_message_collection(repo_id, repo_git: str, logger
 
 def process_messages(messages, task_name, repo_id, logger, augur_db):
 
-    tool_source = "Pr comment task"
     tool_version = "2.0"
     data_source = "Github API"
 
@@ -174,6 +173,12 @@ def process_messages(messages, task_name, repo_id, logger, augur_db):
                 logger.info(f"{task_name}: Processing {message_len-index} messages")
 
         related_pr_or_issue_found = False
+
+        # determine whether this is an issue or PR message so we can set the correct tool_source in metadata
+        if is_issue_message(message["html_url"]):
+            tool_source = "Issue comment task"
+        else:
+            tool_source = "Pr comment task"
 
         # this adds the cntrb_id to the message data
         # the returned contributor will be added to the contributors list later, if the related issue or pr are found


Sukuna0007Abhi:
Thanks @shlokgilda for the review and pls @MoralCode any thoughts?

MoralCode:
As noted in the underlying issue, i want a better understanding of the purpose of the tool_source and tool_version fields before we make any fixes or change the way augur is logging data. That will significantly affect  Shloks first suggestion.

As for historical data: i think its probably fine to leave this as a forward-looking change - its just a simple metadata field and trying to retroactively fix stuff is more of a data corruption risk than leaving it.

My personal preference would be to close this PR until we are a) past chaosscon and have the PR backlog under control, and b) have a better understanding of the intended goal of these fields so we can properly update them (not just in this task, but in other places too, such as https://github.com/chaoss/augur/issues/3486)

sgoggins:
> The logic looks correct, but two things before we merge:
> 
>     1. **Magic strings:** `"Issue comment task"` and `"Pr comment task"` are hardcoded here and probably in other places too. Should define constants to avoid typos:
> 
> 
> ```python
> TOOL_SOURCE_ISSUE_COMMENT = "Issue comment task"
> TOOL_SOURCE_PR_COMMENT = "Pr comment task"
> ```
> 
>     2. **Historical data:** This fixes new messages going forward, but what about existing data? The entire message table currently has incorrect `tool_source` values for issue comments. Should we include a migration script to fix old data? Or open a follow-up issue to track it?
> 
> 
> Would love for other maintainers to chime in here.

Hi @shlokgilda : I don't think there is a way to fix historical data without a one time "on Augur start" style of script. From there we could discern their source based on where the bridge entity record for the message exists: `pull_request_review_msg_ref`, `pull_request_msg_ref` and `issue_msg_ref` ... there will only ever be a record in one of those bridge entities, revealing the origins. This is how I already use these tables in analaysis. FWIW.

sgoggins:
> As noted in the underlying issue, i want a better understanding of the purpose of the tool_source and tool_version fields before we make any fixes or change the way augur is logging data. That will significantly affect Shloks first suggestion.
> 
> As for historical data: i think its probably fine to leave this as a forward-looking change - its just a simple metadata field and trying to retroactively fix stuff is more of a data corruption risk than leaving it.
> 
> My personal preference would be to close this PR until we are a) past chaosscon and have the PR backlog under control, and b) have a better understanding of the intended goal of these fields so we can properly update them (not just in this task, but in other places too, such as #3486)

tool source == Which Augur collector/task/worker inserted the data
tool version == the version of that tool 

We typically only change the version when there is a significant change to its collection logic (i.e., API changes, or refactoring of collection tools).

PR #3599, opened 2026-01-16
feat(schema): add platform columns to contributors and fix data contamination (#3469)

### LINKED ISSUE #3469: Cross-contamination between github and gitlab columns in contributors table

guptapratykshh:
**Description**
 i have introduced platform column to contributors table, creating source of truth for contributor data origin. this resolvesdata contamination where gitlab data was being stuffed into github specific columns (gh_login, gh_id, etc.) and gives way for flexible multi platform support(github, gitlab, forgejo etc)

i have added platform (default: 'github') and platform_username columns to contributors table via new Alembic migration and also updated extraction logic to populate these new fields. gitlab tasks now correctly tag data as platform='gitlab' and leave gh_ columns as NULL.  refactored data_parse.py to use helper function (_extract_base_contributor_data), reducing code duplication and making it easier to add new platforms in the future and also migration automatically populates platform='github' and platform_username for all existing records.

This PR fixes #3469

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -189,6 +189,12 @@ class Contributor(Base):
         String,
         comment="Will be a double population with the same value as gh_login for github, but the local value for other systems. ",
     )
+    platform = Column(
+        String,
+        server_default=text("'github'::character varying"),
+        nullable=False
+    )
+    platform_username = Column(String)
     cntrb_email = Column(
         String,
         comment="This needs to be here for matching contributor ids, which are augur, to the commit information. ",
@@ -280,6 +286,8 @@ class Contributor(Base):
 
         contributor_obj.cntrb_id = cntrb_id.to_UUID()
         contributor_obj.cntrb_login = contributor['login']
+        contributor_obj.platform = 'github'
+        contributor_obj.platform_username = contributor['login']
         contributor_obj.cntrb_created_at = contributor['created_at'] if 'created_at' in contributor else None
         contributor_obj.cntrb_email = contributor['email'] if 'email' in contributor else None
         contributor_obj.cntrb_company = contributor['company'] if 'company' in contributor else None
--- a/augur/application/db/data_parse.py
+++ b/augur/application/db/data_parse.py
@@ -644,6 +644,21 @@ def extract_needed_message_data(comment: dict, platform_id: int, repo_id: int, t
 
     return dict_data
 
+def _extract_base_contributor_data(contributor_data, tool_source, tool_version, data_source):
+    """
+    Helper function to extract common contributor data.
+    """
+    return {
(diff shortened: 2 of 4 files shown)


augur/application/db/models/augur_data.py
```diff
@@ -189,6 +189,12 @@ class Contributor(Base):
         String,
         comment="Will be a double population with the same value as gh_login for github, but the local value for other systems. ",
     )
+    platform = Column(
+        String,
+        server_default=text("'github'::character varying"),
+        nullable=False
+    )
+    platform_username = Column(String)
```
MoralCode, line 197: what is platform username for?

MoralCode (commented):
I left a longer version of this in https://github.com/chaoss/augur/pull/3517#issuecomment-3719732574

TL;DR while the issue suggests the fix for this is to put gitlab data in the gitlab columns, I think its actually better for our future plans of adding more platforms if we use this as an opportunity to make the columns where usernames are simply be platform independent. Thats going to be an easier database migration and a better set up for a less platform-dependent future of augur.

Once things are a little more planned out as far as how platform support should work schema-wise (and the PR backlog is lower, and we arent backlogged on planned database migrations that may cause conflict with your migration 39), I think adding the platform column is a good step towards this.

Thanks for making this contribution though!

augur/application/db/models/augur_data.py
guptapratykshh: platform_username is the platform agnostic column intended to serve as unified source of truth for contributor's handle/login, regardless of source.

row with platform='github' stores login in platform_username,  row with platform='gitlab' also stores login in platform_username.

this decouples our data model from specific platforms. when we add forgejo or bitbucket later, we wo not need to add forgejo_username columns,  we will just use platform='forgejo' and platform_username.

MoralCode:

Thanks!

PR #3603, opened 2026-01-17
fix: prevent theoretical NameError from uninitialized keys variable in github_api_key_handler.py

### LINKED ISSUE #3602: Bug: Uninitialized variable in github_api_key_handler.py causes NameError
## Description

In `augur/tasks/github/util/github_api_key_handler.py`, the `get_api_keys()` method has a bug where the `keys` variable may be uninitialized.

## The Problem

```python
# Lines 95-106
attempts = 0
while attempts < 3:
    try:
        keys = self.get_api_keys_from_database()  # Only defined here
        break
    except:
        time.sleep(5)
        attempts += 1

if self.config_key is not None:
    keys += [self.config_key]  # 💥 NameError if all 3 attempts failed!
```

If `get_api_keys_from_database()` fails all 3 retry attempts, the `keys` variable is **never defined**, causing:
```
NameError: name 'keys' is not defined
```

## Additional Issues in Same Function

1. **Bare `except:`** on line 101 - should be `except Exception as e:` for better debugging
2. **Unused variable** on line 136 - `valid_now = valid_keys` is assigned but never used

## Proposed Fix

Initialize `keys = []` before the while loop, and clean up the other issues.

## I'm working on this

I'll submit a PR shortly.

ANJAN672:
## Description

This PR fixes a bug where the `keys` variable in `get_api_keys()` may be uninitialized, causing a `NameError`.

## The Problem

If `get_api_keys_from_database()` fails all 3 retry attempts, the `keys` variable was never assigned, causing a crash on line 106:

```
NameError: name 'keys' is not defined
```

## Changes Made

1. **Initialize `keys = []` before the retry loop** - Prevents NameError if all attempts fail
2. **Replace bare `except:` with `except Exception as e:`** - Avoids catching system exceptions like KeyboardInterrupt
3. **Add logging for failed attempts** - Helps with debugging
4. **Remove unused code** - Deleted unused `valid_now` variable and commented-out code

## Testing

- Code review confirms the fix addresses the initialization issue
- The change is backwards compatible and doesn't affect normal operation

## Related

Fixes #3602

## Signed commits

### DIFF
--- a/augur/tasks/github/util/github_api_key_handler.py
+++ b/augur/tasks/github/util/github_api_key_handler.py
@@ -92,13 +92,15 @@ class GithubApiKeyHandler():
         if redis_keys:
             return redis_keys
 
+        # Initialize keys to empty list to prevent NameError if all retries fail
+        keys = []
         attempts = 0
         while attempts < 3:
-
             try:
                 keys = self.get_api_keys_from_database()
                 break
-            except:
+            except Exception as e:
+                self.logger.warning(f"Failed to get API keys from database (attempt {attempts + 1}/3): {e}")
                 time.sleep(5)
                 attempts += 1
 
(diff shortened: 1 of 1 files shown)


MoralCode:
my understanding from slack was that you havent actually encountered this exception, but noticed it was theoretically possible, so i changed the issue title to reflect that

augur/tasks/github/util/github_api_key_handler.py
```diff
@@ -92,13 +92,15 @@ def get_api_keys(self) -> List[str]:
...
+        # Initialize keys to empty list to prevent NameError if all retries fail
+        keys = []
         attempts = 0
         while attempts < 3:
-
             try:
                 keys = self.get_api_keys_from_database()
                 break
-            except:
+            except Exception as e:
```
sgoggins, line 102: I think this may be more effective using `except*` to capture nested exceptions and decode them all; though I think @ABrain7710 will have the final say on whether or not this is a good idea.

sgoggins:
If this is not theoretical, can you add the stack, @ANJAN672 ?

PR #3604, opened 2026-01-17
Fix race condition causing API keys to deplete to 0 near reset time

### LINKED ISSUE #3598: first api key gets run down to 0 shortly before the keys reset
this is as we work through our backlog of secondary tasks, which are very API heavy - this key was not exhausted and stayed at around 40 queries left until all the keys were exhausted, and then it started showing up as 0. It may also be related to the timing (i.e. when keys are expired and within 15 min or so of the reset time) 

Image

this is very minor overall but wanted to file anyway

Sukuna0007Abhi:
The issue is exhausted from a timing mismatch where the local system assumed reset was imminent (within ~2 minutes), but GitHub hadn't actually refreshed the key yet, leading to rapid cycling in a tight loop. To fix this, I've implemented a configurable 120-second buffer in the __handle_github_ratelimit_response method of github_data_access.py, ensuring keys stay expired long enough for GitHub to catch up.

## Enhancement: 
The buffer uses max(epoch_when_key_resets, time.time() + GITHUB_KEY_RESET_BUFFER) to only delay when necessary, preventing unnecessary waits while avoiding the race. This resolves the depletion issue without impacting normal operation, and the buffer is adjustable via the GITHUB_KEY_RESET_BUFFER environment variable for flexibility.So it does not stick to hardcoded the 120 seconds. It allows the buffer value to be configured via an environment variable...

### How the buffer fixes it: 
By setting expire_time = max(epoch_when_key_resets, time.time() + 120), the key stays expired for at least 120 seconds from now. 

### This ensures:
If reset is imminent, the key waits extra time for GitHub to catch up.
The loop is broken because the key isn't available for reuse too soon.

### Why 120 seconds: 
Matches the "within 2 minutes" window from the issue where the race occurs. It's conservative but not excessive.

## Changes
- Added a configurable 120-second buffer in `__handle_github_ratelimit_response`.
- Keys now stay expired for at least 2 minutes when reset is imminent, preventing premature reuse.
- Buffer is set via `GITHUB_KEY_RESET_BUFFER` env var (default 120s).

## Credits
I used contributor @guptapratyksh's approach, thanks to him, but a little enhancement has done more to it.... As I fixed the hardcoded 120 seconds part..
& Thanks to @MoralCode and @cdolfi for flagging this issue..

I enhanced the contributor's 120-second buffer approach by making it conditional (only applies when reset is imminent via max()), configurable via env var, and defined as a maintainable constant—keeping it efficient and flexible without over-delaying keys.

This PR fixes #3598 

**Signed commits**

### DIFF
--- a/augur/tasks/github/util/github_data_access.py
+++ b/augur/tasks/github/util/github_data_access.py
@@ -1,11 +1,13 @@
 import logging
 import time
+import os
 import httpx
 from tenacity import retry, stop_after_attempt, wait_fixed, retry_if_exception, RetryError
 from urllib.parse import urlparse, parse_qs, urlencode
 from keyman.KeyClient import KeyClient
 
 GITHUB_RATELIMIT_REMAINING_CAP = 50
+GITHUB_KEY_RESET_BUFFER = int(os.getenv("GITHUB_KEY_RESET_BUFFER", 120))
 
 
 class RatelimitException(Exception):
@@ -219,7 +221,8 @@ class GithubDataAccess:
                 key_reset_time = 0
                 
             self.logger.info(f"\n\n\nAPI rate limit exceeded. Key resets in {key_reset_time} seconds. Informing key manager that key is expired")
-            self.key = self.key_client.expire(self.key, epoch_when_key_resets)
+            expire_time = max(epoch_when_key_resets, int(time.time()) + GITHUB_KEY_RESET_BUFFER)
+            self.key = self.key_client.expire(self.key, expire_time)
 
         else:
             self.key = self.key_client.expire(self.key, time.time() + 60)


guptapratykshh:
seems specific changes, as mentioned in the issue thread(https://github.com/chaoss/augur/issues/3598#issuecomment-3762822299), nothing is missed

augur/tasks/github/util/github_data_access.py
```diff
@@ -1,11 +1,13 @@
 import logging
 import time
+import os
 import httpx
 from tenacity import retry, stop_after_attempt, wait_fixed, retry_if_exception, RetryError
 from urllib.parse import urlparse, parse_qs, urlencode
 from keyman.KeyClient import KeyClient
 
 GITHUB_RATELIMIT_REMAINING_CAP = 50
+GITHUB_KEY_RESET_BUFFER = int(os.getenv("GITHUB_KEY_RESET_BUFFER", 120))
```
sgoggins, line 10: Using a default solves for there not being an entry in the config. I think we still need PRs that add config items to provide a pathway for those config items to make it into either the augur.json or the table.

MoralCode:
@sgoggins since you added the ready label, can you explain how you tested this?

MoralCode:
Since the underlying issue is still waiting for a more thorough root cause analysis, the issue is not super severe, and due to the large backlog,  I'm going to close this PR since I would prefer to wait for a more complete analysis of the underlying issue and a PR that is based on it.

PR #3605, opened 2026-01-17
fix: filter NULL comment URLs in github message task to prevent crash

### LINKED ISSUE #3591: Potential Worker Crash due to NULL Data
the sql union query retrieves pr_comments_url and comments_url but does not filter out NULL values. if database contains NULLs for these fields, list comprehension [x[0] for x in result] will include None.

passing None to github_data_access.paginate_resource(comment_url)(line 124) will likely raise an unchecked exception, crashing the task in messages.py

guptapratykshh:
**Description**
filtering out NULL values for pull request and issue comment urls directly in the sql union query. this ensures that only valid endpoints are retrieved. it prevents potential worker crashes by adding secondary null check in list comprehension before passing urls to github paginator.

This PR fixes #3591

**Signed commits**

### DIFF
--- a/augur/tasks/github/messages.py
+++ b/augur/tasks/github/messages.py
@@ -93,21 +93,21 @@ def process_large_issue_and_pr_message_collection(repo_id, repo_git: str, logger
 
         if since:
              query = text(f"""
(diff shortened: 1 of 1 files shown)


sgoggins (approved):
LGTM.

PR #3636, opened 2026-01-30
Fix dictionary syntax in create_section call

PARTHDESHMUKH2005:
**Description**

This PR fixes #

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -623,7 +623,7 @@ class JsonConfig(ConfigStore):
             raise NotWriteableException()
 
         if not self.has_section(section_name):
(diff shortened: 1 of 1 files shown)


MoralCode:
Does this PR solve a particular issue or bug you have experienced? It doesn't appear as though you referenced or linked one in your PR description and it would help others test this fix if theres a stack trace or some other kind of reproduction step for the underlying problem that is available

MoralCode:
Closing because:
- nonresponsive contributor
- unsigned commits
- this code hasnt been an issue since it was introduced
- GenAi confirms that this is just syntactic sugar in python. while the updated syntax is cleaner, it is not an error or a problem

PR #3639, opened 2026-01-31
[tasks] Fix deduplication key collisions and crash on nested data

### LINKED ISSUE #3632: Theoretical Key Collision in remove_duplicates_by_uniques
Description: There are two significant issues in the deduplication utility functions located in augur/tasks/util/worker_util.py:

Key Collision in remove_duplicates_by_uniques:
This function generates lookup keys by joining field values with underscores ("_".join(...)). This leads to collisions when field values contain underscores (e.g., {"a": "foo_bar", "b": "baz"} and {"a": "foo", "b": "bar_baz"} both result in the key "foo_bar_baz"). Additionally, it uses str() on values, which causes None and the string "None" to collide. Crash in remove_duplicate_dicts: This function uses set(tuple(x.items()) for x in data) to deduplicate. If any dictionary value is another dictionary (nested data), the tuple(x.items()) contains a dictionary, which is unhashable, causing a TypeError.

Create a script that calls these utility functions with the following data:
# Case 1: Collision
data = [{"a": "foo_bar", "b": "baz"}, {"a": "foo", "b": "bar_baz"}]
remove_duplicates_by_uniques(data, ["a", "b"]) # Returns 1 item instead of 2
# Case 2: Crash
data = [{"id": 1, "metadata": {"key": "value"}}]
remove_duplicate_dicts(data) # Raises TypeError: unhashable type: 'dict'

Run the script.
Observe the incorrect deduplication and the crash.

remove_duplicates_by_uniques should handle values with underscores without collision (e.g., by using a safer separator or a tuple-based key). remove_duplicate_dicts should handle nested dictionaries (e.g., by recursively converting them to hashable types or using a different deduplication approach). Screenshots N/A (Utility function logic issue)

Log files N/A (Identified during code audit and verified via standalone script)

Software versions:

Augur: 0.91.0
OS: Microsoft Windows 10 Home Single Language (10.0.19045)
Browser: N/A

atheendre130505:
This PR resolves two critical logic issues in the deduplication utilities within 
augur/tasks/util/worker_util.py:

Key Collision: Replaced underscore-joined string keys with tuple-based keys in remove_duplicates_by_uniques
 to prevent false duplicates when field values contain underscores or when distinguishing between None and "None".
TypeError Crash: Introduced a recursive _make_hashable helper function to handle nested unhashable structures (dictionaries and lists), preventing crashes in remove_duplicate_dicts and remove_duplicates_by_uniques.

This PR fixes #3632

Notes for Reviewers
The fix incorporates a new internal helper _make_hashable that ensures any nested data structure is converted to a sorted, immutable tuple before being used as a set or dictionary key. Rigorous edge case testing was performed to verify that boundary collisions (e.g., {"a": "foo_bar", "b": "baz"} vs {"a": "foo", "b": "bar_baz"}) are now handled correctly.

All original imports and the existing heading structure of the file have been preserved.
Signed commits

 Yes, I signed my commits.
Signed-off-by:atheendre130505 atheendre@gmail.com

### DIFF
--- a/augur/tasks/util/worker_util.py
+++ b/augur/tasks/util/worker_util.py
@@ -41,6 +41,14 @@ def wait_child_tasks(ids_list):
             prereq.wait()
 
 
+def _make_hashable(obj):
+    """Recursively convert unhashable objects (dict, list) to hashable tuples."""
+    if isinstance(obj, (tuple, list)):
+        return tuple(_make_hashable(e) for e in obj)
+    if isinstance(obj, dict):
+        return tuple(sorted((k, _make_hashable(v)) for k, v in obj.items()))
+    return obj
+
 def remove_duplicate_dicts(data: List[dict]) -> List[dict]:
     """Remove duplicate dicts from a list
 
@@ -52,8 +60,16 @@ def remove_duplicate_dicts(data: List[dict]) -> List[dict]:
 
     Note:
         The dicts must be perfectly the same meaning the field and data must be exactly the same to be removed
+        This handles nested unhashable types like dictionaries and lists.
     """
-    return [dict(y) for y in set(tuple(x.items()) for x in data)]
+    seen = set()
+    unique_data = []
+    for d in data:
+        h = _make_hashable(d)
+        if h not in seen:
+            seen.add(h)
+            unique_data.append(d)
+    return unique_data
 
 def remove_duplicates_by_uniques(data, uniques):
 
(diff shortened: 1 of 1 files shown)


shlokgilda:
Tuple-based keys fix the theoretical collision, `_make_hashable` handles nested structures, and the new `remove_duplicate_dicts` is actually better than the original since it preserves insertion order (the old set() approach didn't). Although, I am not sure if we even want to add these change considering the fact that this is a defensive pattern. We can discuss this further in the issue #3632.

PR #3641, opened 2026-01-31
Add configurable batch sizes with hybrid config approach

### LINKED ISSUE #3515: Batch sizes should not be hardcoded
Right now, batch sizes for comments/PRs/issues are hardcoded. They should be fecthed from the project config.

shlokgilda:
**Description:** Follow-up to #3523 which normalized hardcoded batch sizes to named constants. This PR makes those constants configurable via the `Tasks` config section using a hybrid approach: one default plus specific overrides for tasks that need different values. This PR fixes #3515.

Three config values instead of one-per-task:                                                                                        
- `default_batch_size` (1000) — used by issues, PRs, PR commits, PR files, facade contributors                                      
- `github_event_batch_size` (500) — events have higher processing overhead                                                          
- `github_message_batch_size` (20) — messages carry large text payloads                                                             
                                                                                                                                    
New `get_batch_size(task_type)` helper in `db/lib.py` handles lookup with fallback to default.                                      
                                                                                                                                    
**Notes for Reviewers**                                                                                                             
                                                                                                                                    
All default values match the previously hardcoded constants, so this is a no-op change for existing deployments.                                    
                                                                                                                                    
**Signed commits**                                                                                                                  

**AI Disclosure:** Used Claude Code to generate this PR draft.

### DIFF
--- a/augur/tasks/github/pull_requests/tasks.py
+++ b/augur/tasks/github/pull_requests/tasks.py
@@ -19,9 +19,6 @@ from augur.application.db.lib import get_secondary_data_last_collected, get_upda
 from typing import List
 
 
-# Batch sizes for PR-related data collection
-# All use default_batch_size from config (default: 1000)
-PR_BATCH_SIZE = PR_REVIEW_COMMENT_BATCH_SIZE = PR_REVIEW_BATCH_SIZE = get_batch_size()
 
 platform_id = 1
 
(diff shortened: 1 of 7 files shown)


augur/tasks/github/pull_requests/tasks.py
```diff
@@ -20,9 +20,10 @@
 
 
 # Batch sizes for PR-related data collection
-PR_BATCH_SIZE = 1000
-PR_REVIEW_COMMENT_BATCH_SIZE = 1000
-PR_REVIEW_BATCH_SIZE = 1000
+# All use default_batch_size from config (default: 1000)
+PR_BATCH_SIZE = get_batch_size()
+PR_REVIEW_COMMENT_BATCH_SIZE = get_batch_size()
+PR_REVIEW_BATCH_SIZE = get_batch_size()
```
MoralCode: nitpick: instead of three database calls to look up a config value, maybe we could chain these to triple assign all the values at once 


```suggestion
PR_BATCH_SIZE = PR_REVIEW_COMMENT_BATCH_SIZE = PR_REVIEW_BATCH_SIZE = get_batch_size()
```

MoralCode (dismissed):
code looks good! had one small nitpick idea

MoralCode:
should also probably figure out why this is failing

shlokgilda:
> should also probably figure out why this is failing

Figured it out: the `get_batch_size()` calls were at module level (as constant assignments), so they ran during import before the database was ready in the Docker environment. Gunicorn workers crashed on boot and facade tasks hit `psycopg2.DatabaseError.` Moved all those calls into the functions that actually use the batch sizes, so the DB is only queried at runtime. 

The linter failure is unrelated. It's pre-existing pylint issues (unused imports, bare excepts, etc.) in the files this PR touches. The pyling action uses `filter_mode: "file"` so it flags the whole file, not just changed lines. I can clean those up if you want but didn't want to mix it into this PR.

MoralCode (approved):
🚢

MoralCode:
Requesting a review from John since he was the driving force behind asking for these changes in a prior PR

Ulincsys (approved):

Issue #3313, opened 2025-10-15
Print messages that should be logging messages


giordano:
Lines like
https://github.com/chaoss/augur/blob/60138d072849647c815d90b65f405d7b307e1109/augur/tasks/github/contributors.py#L38
https://github.com/chaoss/augur/blob/60138d072849647c815d90b65f405d7b307e1109/augur/tasks/github/contributors.py#L53
https://github.com/chaoss/augur/blob/60138d072849647c815d90b65f405d7b307e1109/augur/tasks/github/util/github_api_key_handler.py#L122
should probably be logging messages?  More generally, all of https://github.com/search?q=repo%3Achaoss%2Faugur+%22print%28%22+language%3APython+path%3A%2F%5Eaugur%5C%2Ftasks%5C%2F%2F&type=code should be reviewed?

giordano:
I guess this is somewhat related to #3308.

PredictiveManish:
I’m considering replacing all `print()` statements in `augur/tasks/` modules with appropriate logging calls (`logger.info, logger.warning, logger.error`, etc.) to improve observability and maintain consistency.

Would this be the recommended approach, or are there any better alternatives you would suggest?

MoralCode:
> I’m considering replacing all `print()` statements in `augur/tasks/` modules with appropriate logging calls 

Thats probably what I would do yeah

PredictiveManish:
@MoralCode This is not completed yet #3536 (fixes partially) this covers a small part of `tasks` module, I'll complete all changes and like 4-5 PRs will solve this issue as it covers around 19-20 files and changing all in one PR will be risky and hard to test once so working in small chunks.  



if you can, then assign to me so that no other person confuses whether this issue is taken up or not.

MoralCode:
There shouldnt be confusion about this issue being taken because there are currently 4 PRs (one merged, one open, one closed, one draft) linked to this PR

PredictiveManish:
> There shouldnt be confusion about this issue being taken because there are currently 4 PRs (one merged, one open, one closed, one draft) linked to this PR

Closed one - older one
draft- I think it should be closed as the small ones are just split of that PR(closing now)
merged - partial fix for a particular module
open - PR 2 for partial fix and after this gets merged will move to next one.

PR #3478, opened 2025-12-18
Update: Replaced print() with logging #3313

PredictiveManish:
**Description**
This PR updates all the `print()` in `augur/tasks` to `logging` which helps in better formatting and makes it helpful in unifying logging in augur project. Solves #3313 

This PR fixes #3313 

**Notes for Reviewers**

**Signed commits**
- [x] Yes, I signed my commits.

PR #3648, opened 2026-02-05
fix: Replace print statements with logging and logging modular level  (PR2)

PredictiveManish:
**Description**
This PR standardizes the logging implementation across `libyear_util` to follow Python best practices and solve #3313:

### Changes Made:
1. **Added module-level loggers**: Each module now uses `logger = logging.getLogger(__name__)` 
2. **Eliminated print statements**: All console output now uses proper logging
3. **Fixed inconsistent patterns**: Removed logger parameter passing in favor of module-level loggers
4. **Improved import organization**: Removed duplicate imports

This PR partially solves #3313
Split of #3478 
**Notes for Reviewers**
- Consistent logging behavior across the codebase
- Better control over log levels and output
- Follows Python logging best practices
- Easier debugging and monitoring
- Cleaner separation of concerns

### Testing
- Enabled Python logging locally
- Verified log messages emit via `logger.info`
- No changes to control flow or task behavior
- Verified logging output locally with `logging.basicConfig(level=INFO)`.

**Signed commits**

### DIFF
--- a/augur/tasks/git/dependency_libyear_tasks/libyear_util/pypi_libyear_util.py
+++ b/augur/tasks/git/dependency_libyear_tasks/libyear_util/pypi_libyear_util.py
@@ -70,7 +70,7 @@ def get_latest_version(data):
     return data['info']['version']
 
 
-def get_release_date(data, version):
+def get_release_date(data, version, logger):
     if not data:
         logger.info('invalid data')
         return None


MoralCode:
> This PR partially fixes https://github.com/chaoss/augur/issues/3313

Try not to put the word "fixes" right before the issue citation since thats what triggers github to link it to the underlying issue in a way that will close it when this merges.

its a tricky balance between citing the changes/connecting the issue and still having the multi-part changes work

augur/tasks/git/dependency_libyear_tasks/libyear_util/pypi_libyear_util.py
```diff
@@ -68,7 +70,7 @@ def get_latest_version(data):
     return data['info']['version']
 
 
-def get_release_date(data, version,logger):
+def get_release_date(data, version):
```
MoralCode: This change removes a parameterized log variable to introduce a global one. This is not a good design practice

augur/tasks/git/dependency_libyear_tasks/libyear_util/pypi_parser.py
```diff
@@ -53,13 +52,13 @@ def parse_requirement_txt(file_handle):
     for encoding in ['utf-8', 'utf-16', 'latin1']:
         try:
             manifest = raw_bytes.decode(encoding)
-            logging.debug(f"[{file_name}] Successfully decoded with encoding: {encoding}")
+            logger.debug(f"[{file_name}] Successfully decoded with encoding: {encoding}")
```
MoralCode, line 55: more global variable logging that needs fixing

sgoggins: @MoralCode : @ABrain7710 can set me straight if I am not understanding correctly, but our logging is designed to create logs that are distinct to the different data collection utilities/tasks Augur performs. This is one of those tasks. It would create much more object-creation overhead if a logger instance were instantiated in each method for this task. I am assuming that by global you mean global across methods in this file, which is for one task. 

I think there may be a trade-off between the downside of all the methods in a task sharing a logger and the cost of creating and destroying logger objects in each method.

MoralCode: my feedback here is sort of addressing that problem.

youre right that we dont want to create a logger per method. But i also dont want to be using a global variable that all our methods just... use because then the dependencies/expectations each method requires to run are not clearly laid out in the signature.

What I would like to see here is that, instead of getting a `logger` variable from global scope, the logger would be passed in as a parameter like we do in many other places. Then we can instantiate it when we set up the task and pass it  to the entrypoint, which will pass it down through any other functions that get called.

This will also help us run tasks individually in isolated contexts, like unit testing, more easily.

Issue #3594, opened 2026-01-16
Epic: Flexible platform support


MoralCode:
Currently Augur is largely dependent on the github API (since that was what it was originally written for)

When considering support for new platforms, there are lots of ways we need to be flexible/things we need to do to account for other platforms that dont quite work the same as github. For example:
- We need underlying support for the platforms API (this is the obvous one)
- We need support for changing the domain we use to make requests with that api implementation (i.e. some platforms have multiple instances, i.e. selfhosted forgejo, github/github enterprise #3277, etc)
- We need to associate API keys with the combination of domain and platform that is in use (someones GHE api key prob wont work on regular github)
- Different platforms may have different specifics for handling API keys (i.e. github does rate limits for graphql and REST independently for the same key)

Also
- Different platforms provide different sets of data, so we likely need to adjust the table schema to accomodate as many useful pieces as possible (without just creating platform specific columns #3469 )
- We should probably be standardizing the way tasks are structured and set up ( #3272 )
- We should make sure that our various interfaces show information about all of the platforms (notably the CLI for checking the rate limits left on all configured keys)
- Having an admin panel in the web ui that can manage keys is probably going to help users a lot.

This is an absolutely huge task thats going to need to be broken down pretty small and done very strategically to keep us on track for adding more platforms this year.

Sukuna0007Abhi:
Hi @MoralCode if it's open for us.. I have the suggestion as I worked on #3272 and want to complete 1 part of this task via this #3272 issue..

MoralCode:
based on a slack message i posted:

im less worried about the logic of querying the APIs - its more of managing the associations between which platform implementation we use, which domain we are querying, and which keys we use to do that query, and where the data goes afterwards.

i.e. its more about getting the infrastructure and logistics of querying data in place _in a way thats reusable_ - right now our solutions to a lot of that are: 1) assume the domain never changes, 2) have one list of API keys specifically for that platform, and 3) making new table columns for new platforms.

All of that doesnt really scale well.

MoralCode:
some aspects of this are more critical than others too. I'm less worried about the API keys and domains since we can introduce a new platforms table or columns in an existing table or add it to the config or make a new config or something for storing this information and use those for new platforms while we migrate the old ones.

The biggest concern/critical item is more on the "where do we store the data" side - we are currently labeling DB columns based on the platform, and that's not scalable and cannot be fixed by making a new table and running the new solution in parallel until the old one is migrated.

[...]

PR #3653, opened 2026-02-06
Removed 'from_github' method from PullRequest class

### LINKED ISSUE #3596: Dead code and tuple bug in from_github()
**Description:**
While reviewing PR ingestion, I found that from_github() assigns many fields with a trailing comma, which makes them 1-element tuples instead of scalar values.

for example :
 pr_obj.pr_src_id = int(str(pr['id']).encode(encoding='UTF-8').decode(encoding='UTF-8')),
        pr_obj.pr_src_node_id = pr['node_id'],
        pr_obj.pr_html_url = pr['html_url'],
        pr_obj.pr_diff_url = pr['diff_url'],
        pr_obj.pr_patch_url = pr['patch_url'],
        pr_obj.pr_issue_url = pr['issue_url'],

file location :
https://github.com/chaoss/augur/blob/main/augur/application/db/models/augur_data.py
line  :  1805

This syntax creates tuples like (value,) instead of value,
This is almost certainly unintended.

However, after tracing ingestion flow, it seems that PR ingestion now uses:
extract_data_from_pr_list(...)
So this buggy function may currently be dead / unused code, which is why the tuple bug has not caused visible failures.

#question is if it is dead why this matter:
1.railing commas silently create tuples → dangerous hidden bug
2.Function appears unused → dead code increases confusion and maintenance cost
3.Future refactors might accidentally re-enable this broken path

One of the following should happen:
Option A: Fix tuple bug by removing trailing commas
Option B: Remove or deprecate from_github() if it is fully unused
Before opening a PR, I want to confirm:
Is PullRequest.from_github() completely deprecated in the current ingestion pipeline?

If yes ,then  remove the bug and add a simple comment that this complety dead 
if no ,then only remove the  trailing comma bug

KH-Coder865:
Removed the from_github class method for PullRequest.

**Description**
- Removed the dead code inside the ```PullRequest.from_github``` method.

This PR fixes #3596 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -1801,52 +1801,6 @@ class PullRequest(Base):
     assignees = relationship("PullRequestAssignee", back_populates="pull_request")
     files = relationship("PullRequestFile", back_populates="")
 
-    @classmethod
-    def from_github(cls, pr, repo_id, tool_source, tool_version):
-
-        pr_obj = cls()
-
-        pr_obj.repo_id = repo_id
-        pr_obj.pr_url = pr["url"]
-        pr_obj.pr_src_id = int(str(pr['id']).encode(encoding='UTF-8').decode(encoding='UTF-8')),
-        pr_obj.pr_src_node_id = pr['node_id'],
-        pr_obj.pr_html_url = pr['html_url'],
-        pr_obj.pr_diff_url = pr['diff_url'],
-        pr_obj.pr_patch_url = pr['patch_url'],
-        pr_obj.pr_issue_url = pr['issue_url'],
-        pr_obj.pr_augur_issue_id = None,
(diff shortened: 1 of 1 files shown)


MoralCode:
After noticing how many other versions of this method there are across all our other database classes, I think it may actually be better to fold this change into #3594 so that I can move this method, along with all the others, into a neat github API wrapper of some kind.

This is a major change that I think better lines up with @ABrain7710's intent to move the codebase forward in an Object-oriented direction when he added these "dead" methods (that never ended up getting used).

I am closing this PR, however i wanted to thank you for your contribution as it helped me notice that this code was both here in the first place and related to some work i will be doing in the near future to complete #3594

KH-Coder865:
No Problem.
Happy to contribute more ☺️

PR #3658, opened 2026-02-08
CLI: allow bootstrap commands to run without DB configuration

### LINKED ISSUE #3654: `augur --help` requires a database connection
> - `augur --help` exits early with:
> 
> `ERROR no way to get connection to the database. There is no db.config.json and the AUGUR_DB environment variable is not set`
> 
> [...] help requires DB initialization first.
> If the intended behavior is that `--help` should work without DB config, I’m happy to help implement a fix / open a PR.
>  

 _Originally posted by @mezo78902 in [#3627](https://github.com/chaoss/augur/issues/3627#issuecomment-3861949281)_

IMO the help command should just output command information, link to setup docs, etc. i see no reason it needs a database, even if running any of the commands it documents is likely to require them

mezo78902:
Fixes #3654

Problem

Running augur --help or augur config init in a fresh environment would cause the CLI to crash if the database was not configured (missing db.config.json or AUGUR_DB).

This happened because the CLI eagerly attempted to initialize the database, Redis, and RabbitMQ during module import and decorator execution, calling sys.exit() before the help menu could be displayed.

Solution

This PR introduces graceful degradation for CLI bootstrap commands (such as --help and config init), while preserving strict validation for runtime commands (such as backend start).

Changes
CLI Decorators (augur/application/cli/__init__.py)

Added an is_bootstrap_command(ctx) helper.

Updated test_db_connection and with_database decorators to:

Skip DB initialization for bootstrap commands (e.g. augur config ...)

Fail gracefully with a clear ClickException for DB-required commands when the DB is not configured.

Service Initialization (augur/tasks/init/__init__.py)

Wrapped Redis and RabbitMQ config retrieval in try/except.

Returned default values silently only during bootstrap mode to allow CLI startup.

Preserved loud failures for runtime execution to prevent misconfigured production runs.

Database Engine (augur/application/db/engine.py, session.py)

Replaced early sys.exit() calls with None returns when DB config is missing.

Updated session/engine handling to safely support None engines without raising AttributeError.

Logging (augur/application/logs.py)

Replaced noisy print() statements with logger.info() to keep CLI output clean during --help.

Verification

Tested in a clean environment (no db.config.json, no AUGUR_DB):

- augur --help — displays help cleanly without crashes or warnings

- augur config init — runs successfully and prompts for API keys

- augur backend start — fails loudly with a clear configuration error (no silent failures)

### DIFF
--- a/augur/application/cli/__init__.py
+++ b/augur/application/cli/__init__.py
@@ -1,4 +1,3 @@
-#SPDX-License-Identifier: MIT
 import click
 from functools import update_wrapper
 import os
@@ -11,6 +10,24 @@ import traceback
 from augur.application.db.engine import DatabaseEngine
 from augur.application.db import get_engine, dispose_database_engine
 from sqlalchemy.exc import OperationalError 
+
+def is_bootstrap_command(ctx):
+    """
+    Check if the current command is a bootstrap command (like config init)
+    that should run even without a database connection.
+    """
+    if "--help" in sys.argv:
+        return True
+
+    try:
+        command_path = ctx.command_path.lower()
+        if "augur config" in command_path:
+            return True
+    except AttributeError:
+        pass
+        
+    return False
+ 
 
 
 def test_connection(function_internet_connection):
@@ -49,8 +66,18 @@ def test_connection(function_internet_connection):
 def test_db_connection(function_db_connection):
     @click.pass_context
     def new_func(ctx, *args, **kwargs):
+        
+        if is_bootstrap_command(ctx):
+             return ctx.invoke(function_db_connection, *args, **kwargs)
+        
+
         engine = DatabaseEngine().engine
         usage = re.search(r"Usage:\s(.*)\s\[OPTIONS\]", str(ctx.get_usage())).groups()[0]
+        
+        if engine is None:
+             print(f"\n\n{usage} command setup failed\nERROR: connecting to database\nHint: Database not configured. Run 'augur config init'.\n")
+             sys.exit(-2)
+
         try:
             engine.connect()
             engine.dispose()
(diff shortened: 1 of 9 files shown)


MoralCode:
Thanks for the contribution! This PR seems to make a lot of changes, many of which dont seem related to the core fix (like renaming variables and other formatting)

> augur config init in a fresh environment would cause the CLI to crash if the database was not configured (missing db.config.json or AUGUR_DB).

The augur config system fundamentally relies on the database existing because config is primarily accessed from the database (yes i know its janky, thats why i introduced the hierarchical config system so there was also an option to use a file - the eventual goal being to deprecate database configs where it makes sense).

Can you please update this PR so that it is smaller and easier to review?  Im very hesitant to merge something that touches 9 different files at once. I'd rather touch fewer files and spend the extra time/characters to clearly communicate why you chose to do things the way you did (at first glance, im not sure i agree with the `is_bootstrap_command` helper)

mezo78902:
Thanks for the detailed feedback @MoralCode  — totally fair point about scope.
My goal was strictly to unblock augur --help in a fresh env without DB/Redis/RabbitMQ configured.
I agree augur config init is DB-dependent today; I’ll update the PR to focus only on the help-path and make it smaller/easier to review.

Plan: I’ll drop all unrelated refactors (renames/formatting) and revert any changes not required for --help to load. I’ll keep the minimal change that prevents eager DB/service init during CLI startup for help-only invocations.

I’ll push an updated, smaller diff shortly and add a clear rationale in the PR description for the chosen approach (and alternatives).

mezo78902:
Closing in favor of a smaller, focused PR that only addresses augur --help eager imports (per feedback). Superseded by #3661.

PR #3666, opened 2026-02-09
Feat: Make Celery periodic task scheduling configurable

### LINKED ISSUE #3406: make remaining periodic tasks configurable
We still have several periodic tasks configured in celery with fixed time intervals. These should be pulling their intervals from the config just like the materialized view task

https://github.com/chaoss/augur/blob/6b48ab60443dddc6a03ce17a61b6b395ad54de35/augur/tasks/init/celery_app.py#L231-L260

aayush-patidar:
### Description
 This PR removes hardcoded periodic Celery task intervals and makes task scheduling fully configuration-driven using AugurConfig. Users can now customize scheduling frequencies or disable specific periodic tasks via augur.config.json without modifying source code.
Default configuration values preserve the existing scheduling behavior when no custom configuration is provided, ensuring full backward compatibility.
**Fixes #3406**

### Problem
Several periodic Celery tasks in augur/tasks/init/celery_app.py were scheduled using hardcoded time intervals. This made it difficult for users to adjust task frequency or disable tasks without changing source code.

### Solution
All previously hardcoded Celery task schedules have been refactored to read their values from configuration, following the same pattern already used for materialized view scheduling.

### Changes

-  Refactored setup_periodic_tasks to retrieve scheduling intervals from AugurConfig instead of hardcoded constants.
-  Added new configuration keys under the [Tasks] section with defaults matching previous behavior:

    -     non_repo_domain_tasks_interval_in_days
    -     retry_errored_repos_cron_hour
    -     retry_errored_repos_cron_minute
    -    process_contributors_interval_in_seconds
    -    create_collection_status_records_interval_in_seconds

- Implemented logic to disable interval-based tasks when the configured value is <= 0.
- Updated configuration documentation to describe new keys, units, defaults, and disable behavior.
- Added automated tests to validate scheduling logic, default fallbacks, and disabled tasks.

### Backward Compatibility
All new configuration options include defaults that match the original hardcoded intervals. Existing installations without these keys defined will continue to operate with the same task schedules as before.

### Testing

- Added tests/tasks/test_celery_scheduler.py
- Run tests with:
    -  pytest tests/tasks/test_celery_scheduler.py

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -149,7 +149,8 @@ constraint-dependencies = [
 [tool.pytest.ini_options]
 addopts = "-ra -s"
 testpaths = [
-    "tests",
+    "test/tasks",
+    "tests/tasks/test_celery_scheduler.py",
     "tests/test_classes",
     "tests/test_application/test_cli/test_csv_utils.py",
     # "tests/test_routes", # runs, but needs a fixture for connecting to the web interface of Augur


aayush-patidar:
Hi @sgoggins  
This PR makes all periodic Celery task schedules configurable via AugurConfig, following the same pattern already used for materialized views.  
Defaults preserve existing behavior, tasks can be disabled via config, and tests + docs are included.  
Happy to adjust anything if needed — thanks for the review!

augur/tasks/init/celery_app.py
```diff
@@ -250,14 +255,32 @@ def setup_periodic_tasks(sender, **kwargs):
...
-
-        one_hour_in_seconds = 60*60
-        sender.add_periodic_task(one_hour_in_seconds, process_contributors.s())
-
-        one_day_in_seconds = 24*60*60
-        sender.add_periodic_task(one_day_in_seconds, create_collection_status_records.s())
+        # retry_errored_repos
+        retry_repos_hour_val = config.get_value('Tasks', 'retry_errored_repos_cron_hour')
+        retry_repos_minute_val = config.get_value('Tasks', 'retry_errored_repos_cron_minute')
+        retry_repos_hour = int(retry_repos_hour_val) if retry_repos_hour_val is not None else 0
```
MoralCode, line 261: im not sure if we should be defining cron parameters as individual config items

docs/source/development-guide/configuration-file-reference.rst
```diff
@@ -3,6 +3,47 @@ Configuration file reference
...
+-----------------------------
+
+The following keys in the ``[Tasks]`` section control the scheduling of periodic Celery tasks. You can adjust these intervals or disable tasks by setting the value to <= 0 (for interval-based tasks).
+
+- ``collection_interval``: Interval in seconds for the main collection monitor. Default: 30.
+- ``core_collection_interval_days``: Interval in days for core collection. Default: 15.
+- ``secondary_collection_interval_days``: Interval in days for secondary collection. Default: 10.
+- ``facade_collection_interval_days``: Interval in days for facade collection. Default: 10.
+- ``ml_collection_interval_days``: Interval in days for ML collection. Default: 40.
+
```
MoralCode, line 16: we already have docs in ReadTheDocs for many of these config items. Can you put this documentation alongside those (i know they may not be in the best place, but id prefer that than introducing more duplicate docs)

tests/tasks/test_celery_scheduler.py
MoralCode, line 1: nice job adding a unit test!

Can you add this file to the `pyproject.toml` near the bottom in the pylint section so that it can get run as part of the CI job to verify that the test works?

aayush-patidar: Thanks for the suggestion!
I’ve updated `pyproject.toml` to ensure the root `tests/` directory is included in pytest discovery so that `tests/tasks/test_celery_scheduler.py` is executed as part of CI.
Please let me know if this aligns with what you had in mind or if any additional adjustments are needed.

pyproject.toml
```diff
@@ -149,6 +149,7 @@ constraint-dependencies = [
 [tool.pytest.ini_options]
 addopts = "-ra -s"
 testpaths = [
+    "tests",
```
MoralCode: this is going to run all the tests, which is likely going to cause some of the older, not-yet-fixed ones to fail.

if your test is the only one within the tasks folder can you change this to `tests/tasks` (and if not also add the filename and the .py extension)

.github/workflows/build_docker.yml
```diff
@@ -286,7 +285,7 @@ jobs:
           podman compose -f docker-compose.yml up --no-build 2>&1 \
             | (./scripts/ci/await_all.py /tmp/regex_matches.txt \
                 && podman compose -f docker-compose.yml down)
-        timeout-minutes: 3
+        timeout-minutes: 10
```
MoralCode, line 288: This change is not related to this PR.

PR #3670, opened 2026-02-09
check for a lack of 200 status when looking up the moved repo name and log it

MoralCode:
**Description**
- This introduces a small defensive check for the HTTP status and presence of the value we need into the github move detection flow with some logging to help identify the root cause of the issue.

This PR will be helpful towards fixing #3621

**Notes for Reviewers**

**Signed commits**

AI provided the suggestion for this change and wrote the code, which was small enough for me to review manually and make some corrections to (such as printing the url from the variable we already have, removing some comments, etc.

### DIFF
--- a/augur/tasks/github/detect_move/core.py
+++ b/augur/tasks/github/detect_move/core.py
@@ -59,10 +59,12 @@ def extract_owner_and_repo_from_endpoint(key_auth, url, logger):
     response_from_gh = hit_api(key_auth, url, logger)
 
     page_data = parse_json_response(logger, response_from_gh)
+    # Note: if parse_json_response fails, it will return a string, causing `in` to do a substring check.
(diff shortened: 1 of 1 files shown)


MoralCode:
> Also might be worth logging `response_from_gh.text` instead of `page_data` in the error? If `parse_json_response` mangles the response, you'd lose the raw output that'd actually help debug what GitHub sent back.

Is that still available if the json data has been consumed?

I feel like I've seen weird python bugs before where accessing the response of a http call returns something like a generator/reader object that, once accessed, needs to be stored in a variable in order to be reused

sgoggins (approved):
LGTM. I verified that this gets called inside of our retry block. FYI, the retry block is very critical for not marking a repository as problematic simply because the GitHub API choked for a few minutes.

PR #3375, opened 2025-11-05
Add flexible column order support for CSV imports (Fixes #3310)

shlokgilda:
**Description**
This PR adds flexible column order support for CSV imports in the `augur db add-repos` and `augur db add-repo-groups` CLI commands. This PR fixes #3310.

### Summary of Changes

 - **New module:** `augur/application/cli/csv_utils.py` containing CSV processing utilities with intelligent header detection
 - **Refactored:** `add-repos` and `add-repo-groups` commands to use the new CSV parser
 - **Features added:**
   - Support for CSVs with headers (column order can be arbitrary)
   - Backward compatibility with existing headerless CSVs
   - Automatic column type detection for headerless CSVs (distinguishes URL columns from ID columns)
   - 10MB file size limit with clear error messages
   - Rejection file generation (`.rejected.csv`) for failed imports
 - **Updated:** Sample CSV files to include headers as examples

 ### Testing

 I have manually tested the implementation with various CSV formats:
 - CSV files with headers in different column orders
 - Headerless CSV files (backward compatibility)
 - Headerless CSV files with swapped columns (auto-detection)
 - Invalid data scenarios (rejection file generation)
 - File size limit enforcement

 **Note on Test Suite:** I attempted to run the existing test suite but encountered import errors unrelated to this PR (`ORG_REPOS_ENDPOINT` import path issues: `test_add_cli_repos.py`, `test_repo_group.py`, `test_user.py`, etc. use the constant but it was moved from `augur.util.repo_load_controller` to `augur.application.db.models.augur_operations`). This is a pre-existing issue in the codebase that should be addressed in a separate PR. I intentionally did not modify any existing test files to avoid scope creep.

**Notes for Reviewers**

- The new `csv_utils.py` module is fully self-contained and has no external dependencies beyond Python stdlib and SQLAlchemy (already in the project)
- Changes to `db.py` are minimal and focused on integrating the new CSV parser
- All original functionality is preserved - existing headerless CSVs will continue to work
- The implementation adds new capabilities without breaking existing workflows

**Signed commits**
- [x] Yes, I signed my commits.

PR #3683, opened 2026-02-12
enable the (new ish) CSV utils tests

MoralCode:
**Description**
in https://github.com/chaoss/augur/pull/3375, @shlokgilda improved the csv importer to be flexible with column order (since our CSVs are only 2 columns). This PR came with unit tests that were never enabled.

This PR fixes that by adding those tests to the list of tests that get run when `uv run pytest` is run (also in CI)
**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,7 @@ constraint-dependencies = [
 addopts = "-ra -s"
 testpaths = [
     "tests/test_classes",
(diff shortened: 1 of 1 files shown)


shlokgilda (approved):
LGTM

MoralCode:
bypass merge because trivial change and already tested with CI

PR #3685, opened 2026-02-12
Start consolodating Github REST API URLs into a central place

MoralCode:
**Description**
This takes inspiration from the discussion in #3597 and is a step towards resolving #3277

This pr:
- refactors a few commonly reused functions that generate github API URLs (specifically for the issues, contributors, and user endpoints) so that the url building happens within the existing GithubDataAccess class.
- creates a function that brings the fetching of github user API data into this class as well (a model for future plans for the other endpoints)
- creates a user_endpoint_urls function that, given a username, can recreate the urls that the github API responds with. This change compliments #3747.
- creates a currently-unused general-purpose function to generate URLs for querying Github's search API, removing the need to manually assemble the URL in an API-compliant way from clients that use it. clients just provide the query string as it would be typed into githubs search box. This change also enabled some current legacy facade code for doing these searches to be replaced with this general-purpose implementation.

**Notes for Reviewers**
Augur collection appears to run smoothly with this PR on my local dev instance, indicating this change doesnt cause breakage.

**Signed commits**

### DIFF
--- a/augur/tasks/github/util/github_data_access.py
+++ b/augur/tasks/github/util/github_data_access.py
@@ -37,6 +37,9 @@ class ResourceGoneException(Exception):
 
 class GithubDataAccess:
     """Utilities for accessing the GitHub REST API
+    
+    Public facing functions in this class should refrain from returning data in a structure
+    that is derived from githubs API responses to keep all platform-specific parsing here.
     """
 
     def _base_domain(self) -> str:


augur/tasks/github/util/github_data_access.py
```diff
@@ -34,7 +34,18 @@ def __init__(self, message="Resource returned HTTP 410 Gone. It is likely intent
...
 class GithubDataAccess:
+    """Utilities for accessing the GitHub REST API
+    """
+
+    def _base_url(self):
+        return "https://api.github.com/"
+
+    def issues_endpoint_url(self, owner, repo, trailing_slash = True):
+        """ https://api.github.com/repos/{owner}/{repo}/issues/ """
+        return f"https://api.github.com/repos/{owner}/{repo}/issues" + "/" if trailing_slash else ""
```
shlokgilda: This has an operator precedence issue. `+` binds tighter than `if/else`, so this evaluates as:
```python
return (f"...issues" + "/") if trailing_slash else ""
```
When `trailing_slash=False,` this returns `""` and not the URL without a slash. `issues.py` calls this with trailing_slash=False, so it'd produce `"?state=all"` as the full URL.

Needs parens:
```python
return f"https://api.github.com/repos/{owner}/{repo}/issues" + ("/" if trailing_slash else "")
```

augur/tasks/github/util/github_data_access.py
```diff
@@ -34,7 +34,18 @@ def __init__(self, message="Resource returned HTTP 410 Gone. It is likely intent
         super().__init__(message)
 
 class GithubDataAccess:
+    """Utilities for accessing the GitHub REST API
+    """
+
+    def _base_url(self):
+        return "https://api.github.com/"
```
shlokgilda: This method exists but isn't used? `issues_endpoint_url` below still hardcodes https://api.github.com/. Should probably be: 
```python
def issues_endpoint_url(self, owner, repo, trailing_slash=True):
    return f"{self._base_url()}repos/{owner}/{repo}/issues" + ("/" if trailing_slash else "")
```

shlokgilda (changes requested):
Like the idea. 

One thing that needs fixing before merge: there's an operator precedence bug in `issues_endpoint_url` that'll break the `trailing_slash=False` path. Python evaluates the ternary after the `+`, so when `trailing_slash=False` you get an empty string back instead of the URL without a slash. The issues endpoint actually hits this path, so it'd break issue collection. 

Also, `_base_url()` is defined but never called. `issues_endpoint_url` still hardcodes the domain. Worth wiring that in since that's the whole point of the refactor.

Details in the inline comments.

augur/tasks/github/util/github_data_access.py
MoralCode: sounds good. BTW if youre going to effectively write the solution in a review comment, use the "suggestion" button (first icon in the top bar) itll pre-populate the line that the review comment was made on and will allow the suggestion to be quickly added as a new commit


```suggestion
        return f"https://api.github.com/repos/{owner}/{repo}/issues" + ("/" if trailing_slash else "")
```

augur/tasks/github/util/github_data_access.py
MoralCode: lol u right, oops

augur/tasks/github/util/github_data_access.py
shlokgilda: Aaah, thank you!! I've been looking for this for such a long time. Finally figured out how you do that! Good to know.

sgoggins:
Needs to be tested before merge.

augur/tasks/github/facade_github/core.py
```diff
@@ -41,6 +35,9 @@ def query_github_contributors(logger, key_auth, github_url):
 
     github_data_access = GithubDataAccess(key_auth, logger)
 
+    # Set the base of the url and place to hold contributors to insert
+    contributors_url = github_data_access.contributors_endpoint_url(owner, repo) + "?state=all"
```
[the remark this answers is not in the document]
MoralCode: this commit shouldnt be here....

MoralCode:
currently testing on my dev instance

PR #3686, opened 2026-02-12
access GraphQL urls from the designated graphql helper class

MoralCode:
**Description**
This is essentially the same as #3685, but for the github graphQL API (another small step towards resolving https://github.com/chaoss/augur/issues/3277)

**Notes for Reviewers**
No testing done yet, pretty trivial change to set the pattern for how i want things to work with building URLs in the future

**Signed commits**

### DIFF
--- a/augur/tasks/github/releases/core.py
+++ b/augur/tasks/github/releases/core.py
@@ -1,5 +1,6 @@
 #SPDX-License-Identifier: MIT
 from augur.tasks.github.util.github_task_session import *
+from augur.tasks.github.util.github_graphql_data_access import GithubGraphQlDataAccess
 from augur.application.db.models import *
 from augur.tasks.github.util.util import get_owner_repo
 from augur.tasks.github.util.gh_graphql_entities import request_graphql_dict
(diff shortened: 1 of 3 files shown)


MoralCode:
So i did graphQL in a different way because graphQL queries dont really rely on url paths or parameters, they send the information for the request in the body.

I worry that by exposing the baseURL for REST in a similar way, consumers of this class may be encouraged to build their own queries (although i guess if theyre using the correct baseurl its probably fine????

shlokgilda:
Yeah makes sense, keeping them different is the right call. LGTM.

shlokgilda (approved):

sgoggins:
> So i did graphQL in a different way because graphQL queries dont really rely on url paths or parameters, they send the information for the request in the body.
> 
> I worry that by exposing the baseURL for REST in a similar way, consumers of this class may be encouraged to build their own queries (although i guess if theyre using the correct baseurl its probably fine????

This is not a comment with any valence. The REST API's rate limits are published and clear. The GraphQL API limits are generally greater, however we have observed they are also severely throttled from time to time. So, this is a "the more you know" sort of comment. This is the kind of thing we learn when collecting mass amounts of data for a while.

Issue #3662, opened 2026-02-08
Standardize our linter rules


MoralCode:
The maintainers need to decide on what our canonical code style is and adjust the pylint rules to enforce that.

PR #3695, opened 2026-02-14
Populate contributor profile fields from GitHub user API (#3694)

### LINKED ISSUE #3694: Contributor user information from profile - emails as well as other available info
This is where information from a users GH profile should be getting populated from: 

https://github.com/chaoss/augur/blob/main/augur/tasks/github/facade_github/core.py

Right now, the vast majority of the available information (most importantly the listed email) is getting missed. 

There seems to be some minor edits needed to get the all of the available info populated into the contributors table 

Note: Note: contributor resolution is a different limit (50/hr)

antcybersec:
- Fix NameError when email is missing: set canonical_email from email (both can be None) so contributor insert never fails on missing email.
- Use .get() for all optional profile fields (email, name, company, location, created_at, updated_at, and all gh_* URL fields) so every available value from the user endpoint is stored in the contributors table.

**Description**

This PR fixes #3694

**Notes for Reviewers**

**Signed commits**


augur/tasks/github/facade_github/core.py
sgoggins, line 117: @MoralCode : Is this how we are propagating errors now? I'm asking because I know it's how we did that originally, and I know @ABrain7710 had at one point advised a different strategy. Time for a style guide.

sgoggins (commented):
Do you have specific logs without this fix to illustrate it fixes the bug? Have you tested this? Really appreciate you diving in on this one, it's just more efficient to ask this question than to run it myself. 🫠

augur/tasks/github/facade_github/core.py
antcybersec: I kept the existing pattern in this file (log then raise e) and didn’t change the error handling. Happy to follow whatever strategy you and @ABrain7710 prefer for propagating errors and to update this PR if you decide on a style.

sgoggins: I don't think we've edited this file in a spell, so I'm more questioning if there's a new style we are using that we should implement while we are here. So, that's not at all a critique of what you did. Following what's already in file is exactly what I would have done. I'm more asking other maintainers so we take the opportunity of this pr to pay down tech debt. This might be how we are doing it now ... I've been on augur for too long to be sure without asking Andrew and Adrian.

MoralCode: I don't have a style guide currently available that i know about. I think that would make a good compliment to #3662. I don't think we should enforce style guidelines that are undocumented as that isn't fair to new contributors.

MoralCode:
@antcybersec are you on slack? 
I'd like to chat about this in more detail over on the [CHAOSS Slack](https://chaoss.community/kb-getting-started/) in the #wg-augur-8knot channel. I think its a good change, but want to make sure everyone is on the same page

iGufrankhan:
@MoralCode 

I checked out the PR branch locally and tested ingestion on psf/requests.
 Contributor resolution completed successfully, contributors without public email were inserted with canonical_email set to 
 NULL, and I did not observe any NameError or crashes during ingestion.

image


all runnning good 

now let me verify before the test log if error happening or not

augur/tasks/github/facade_github/core.py
```diff
@@ -62,58 +62,52 @@ def query_github_contributors(logger, key_auth, github_url):
...
-                "gh_type": contributor['type'],
-                "gh_site_admin": contributor['site_admin'],
-                "cntrb_last_used" : None if 'updated_at' not in contributor else contributor['updated_at'],
-                "cntrb_full_name" : None if 'name' not in contributor else contributor['name'],
-                #"tool_source": session.tool_source,
-                #"tool_version": session.tool_version,
-                #"data_source": session.data_source
+                "gh_url": contributor.get('url'),
+                "gh_html_url": contributor.get('html_url'),
+                "gh_node_id": contributor.get('node_id'),
```
shlokgilda, line 97: The old code had a comment here noting this field is used for duplicate checking. Might be worth keeping that context since it's not obvious from the field name alone.
```suggestion
                "gh_node_id": contributor.get('node_id'),  # Used for duplicate checking
```

shlokgilda (approved):
Solid fix. Left a minor inline comment, nothing blocking. Looks good to merge once that's considered.

augur/tasks/github/facade_github/core.py
MoralCode: this is documenting a usecase in another part of augur's code, rather than something from github and is minor enough that im inclined to just let it go. if its being used, im sure a ctrl-f for it in the augur code will help locate the code using it.

PR #3696, opened 2026-02-15
update the environment variable example to be more intuitive

MoralCode:
**Description**
This updates the environment sample file based on feedback from CHAOSScon to be more intuitive and in line with what was presented on the slides

**Notes for Reviewers**
Nonfuctional change that's essentially a docs fix

**Signed commits**


environment.txt
```diff
@@ -1,7 +1,11 @@
...
 AUGUR_GITHUB_API_KEY=<PUT YOUR GITHUB API KEY HERE>
 AUGUR_GITHUB_USERNAME=<PUT YOUR GITHUB USERNAME HERE> 
-AUGUR_GITLAB_API_KEY=<PUT YOUR GITLAB KEY HERE>
-AUGUR_GITLAB_USERNAME=<PUT YOUR GITLAB USERNAME HERE>
-AUGUR_DB=<PUT YOUR DATABASE CONNECTION STRING HERE>
+AUGUR_GITLAB_API_KEY=<PUT YOUR GITLAB KEY HERE or make something up>
+AUGUR_GITLAB_USERNAME=<PUT YOUR GITLAB USERNAME HERE or make something up>
+
 AUGUR_RABBITMQ_USERNAME=<OPTIONAL - SET DESIRED RABBITMQ USER>
 AUGUR_RABBITMQ_PASSWORD=<OPTIONAL - SET DESIRED RABBITMQ PASSWORD>
```
MoralCode, line 12: This is used at build time, which i guess can stay here.... its weird

shlokgilda:
Previously `AUGUR_DB` was a full connection string like `postgresql+psycopg2://user:pass@host:5432/augur`. Now it's just `augur.`  What's the intended usage here? The Python code in `engine.py` and a few other places reads `AUGUR_DB` directly as a SQLAlchemy URL, so won't setting it to just a database name would break things outside Docker?

Am I misunderstanding the flow here?

shlokgilda:
Also, `entrypoint.sh` constructs the connection string from `AUGUR_DB_HOST,` `AUGUR_DB_USER,` `AUGUR_DB_PASSWORD,` and `AUGUR_DB_NAME`. But `AUGUR_DB_HOST` isn't listed here. Wouldn't that check fail silently and leave `AUGUR_DB` unset?

MoralCode:
ah yeah i missed that, maybe defaults were saving me.

that augur_DB at the top should be AUGUR_DB_HOST and AUGUR_DB_NAME, defaulting to the same value

shlokgilda:
Aah okay! I'll wait for your changes.

shlokgilda:
Looks good overall. 

One thing: `AUGUR_DB_HOST=augur` should be `AUGUR_DB_HOST=augur-`db to match the Docker service name in `docker-compose.yml.` Right now, it doesn't matter because `docker-compose.yml` hardcodes the full `AUGUR_DB` string (so `entrypoint.sh` never constructs it from the components), but if someone removes that line expecting the component vars to work, `augur` won't resolve, only `augur-db` will.

Might also be worth adding `AUGUR_DB_PORT=5432` since `docker-compose.yml` already references  `{AUGUR_DB_PORT:-5432}` for the port mapping, and a short comment at the top noting these are used by Docker's `entrypoint.sh` to build the connection string.

environment.txt
sgoggins: I am usure what the GitLab "or make something up" is intended to communicate. :)

PR #3701, opened 2026-02-16
convert comments near contributor resolution functions into actual docstrings

MoralCode:
**Description**
non-functional change to re-format our existing function docs in the contributor resolution file to be in proper python syntax so that they show up in peoples IDE's

**Notes for Reviewers**
no functional changes made. no testing done.
**Signed commits**

### DIFF
--- a/augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
+++ b/augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
@@ -289,7 +289,7 @@ def update_contributor(self, cntrb, max_attempts=3):
 
 
 
-def fetch_username_from_email(logger, auth, commit) -> dict:
+def fetch_username_from_email(logger, auth, commit) -> dict | None:
     """Try every distinct email found within a commit for possible username resolution.
     Add email to garbage table if can't be resolved.
     


augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -138,8 +143,18 @@ def create_endpoint_from_commit_sha(logger, commit_sha, repo_id):
...
+
+    Args:
+        contributor (_type_): _description_
+
+    Raises:
+        ValueError: raises when name is one word or none.
+
+    Returns:
+        str: url
+    """
```
shlokgilda, line 157: The _type_ / _description_ placeholders look like auto-generated IDE stubs that weren't filled in. Maybe something like:

```suggestion
    """Try to construct the best url to ping GitHub's API for a username given a full name.

    Args:
        contributor (dict): Contributor record containing name fields.

    Raises:
        ValueError: If the contributor's name is one word or empty.

    Returns:
        str: GitHub search API url for the contributor's full name.
    """
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -196,10 +213,13 @@ def insert_alias(logger, contributor, email):
...
-# Updates the alias table if the login is already in the contributor's table with the new email.
-# Returns whether the login was found in the contributors table
-def resolve_if_login_existing(logger, contributor):
+def resolve_if_login_existing(logger, contributor) -> bool:
+    """ Takes the user data from the endpoint as arg
+    Updates the alias table if the login is already in the contributor's table with the new email.
+    Return:
+        bool: whether the login was found in the contributors table
+    """
+
```
shlokgilda: `Return:` should be `Returns:` to match the other docstrings in this same PR.

```suggestion
    """Takes the user data from the endpoint as arg.
    Updates the alias table if the login is already in the contributor's table with the new email.

    Returns:
        bool: Whether the login was found in the contributors table.
    """
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -262,13 +282,23 @@ def update_contributor(self, cntrb, max_attempts=3):
...
+ 
+    Args:
+        logger (_type_): logging instance to use
+        auth (_type_): Keymanager instance to use for auth
+        commit (_type_): _description_
+
+    Returns:
+        dict: A dictionary of response data from github with potential logins on success.
+#           None on failure
+    """
```
shlokgilda, line 306: This one has a few issues stacked up:
- `#   \param contributor is the raw database entry`: seems like leftover Doxygen comment syntax inside the docstring? 
- `_type_ / _description_` placeholders not filled in

```suggestion
    """Try every distinct email found within a commit for possible username resolution.
    Add email to garbage table if can't be resolved.
  
    Args:
        logger: Logger instance.
        auth: GithubRandomKeyAuth instance for API authentication.
        commit (dict): Raw commit database entry containing 'email_raw'.
  
    Returns:
        dict: Response data from GitHub with potential logins on success. None on failure.
    """
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -262,13 +282,23 @@ def update_contributor(self, cntrb, max_attempts=3):
...
 
-# Try every distinct email found within a commit for possible username resolution.
-# Add email to garbage table if can't be resolved.
-#   \param contributor is the raw database entry
-#   \return A dictionary of response data from github with potential logins on success.
-#           None on failure
 
-def fetch_username_from_email(logger, auth, commit):
+
+def fetch_username_from_email(logger, auth, commit) -> dict:
```
shlokgilda: Nit: return type should be `Optional[dict]` since this function returns `None` on failure.

```suggestion
def fetch_username_from_email(logger, auth, commit) -> dict | None:
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -27,13 +27,18 @@ def clean_dict(d):
...
+    """Hit the endpoint specified by the url and return the json that it returns if it returns a dict.
+    
+    NOTE: This function is being deprecated in favor of GithubDataAcess.get_resource()
+    No functional change in this function body; logic preserved fully.
+    This function still attempts to hit a GitHub API endpoint and return a dictionary if successful.
+
+
+    Returns:
+        None on failure. Dict on success
+    """
```
shlokgilda, line 41: The `# deprecated` comment above the function is redundant now that the docstring says the same thing. Also two blank lines between the note and `Returns:`.

```suggestion
def request_dict_from_endpoint(logger, session, url, timeout_wait=10):
    """Hit the endpoint specified by the url and return the json that it returns if it returns a dict.

    Deprecated: Use GithubDataAccess.get_resource() instead.
    No functional change in this function body; logic preserved fully.

    Returns:
        None on failure. Dict on success.
    """
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -157,9 +172,11 @@ def create_endpoint_from_name(contributor):
...
 
 def insert_alias(logger, contributor, email):
-    # Insert cntrb_id and email of the corresponding record into the alias table
-    # Another database call to get the contributor id is needed because it's an autokeyincrement accessed by multiple workers
-    # Same principle as enrich_cntrb_id method.
+    """
+    Insert cntrb_id and email of the corresponding record into the alias table
+    Another database call to get the contributor id is needed because it's an autokeyincrement accessed by multiple workers
+    Same principle as enrich_cntrb_id method.
+    """
```
shlokgilda, line 184: Should also add `Args` and `Raises` here for consistency with the other docstrings in this PR.

```suggestion
    """Insert cntrb_id and email of the corresponding record into the alias table.

    Another database call to get the contributor id is needed because it's an
    autokeyincrement accessed by multiple workers. Same principle as enrich_cntrb_id method.

    Args:
        logger: Logger instance.
        contributor (dict): Contributor record containing 'gh_user_id' and optionally 'cntrb_canonical'.
        email (str): Email address to create the alias for.

    Raises:
        LookupError: If the contributor cannot be found in the database by gh_user_id.
    """
```

sgoggins (dismissed):
LGMT

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -306,10 +336,10 @@ def fetch_username_from_email(logger, auth, commit):
 
     return login_json
 
-# Method to return the login given commit data using the supplemental data in the commit
-#   -email
-#   -name
+
 def get_login_with_supplemental_data(logger, auth, commit_data):
+    """Method to return the login given commit data using the supplemental data in the commit (email and name)
+    """
```
shlokgilda, line 355: Should have `Args` and `Returns` for consistency with the other docstrings.

```suggestion
    """Return the login given commit data using the supplemental data in the commit (email and name).

    Args:
        logger: Logger instance.
        auth: GithubRandomKeyAuth instance for API authentication.
        commit_data (dict): Commit record containing 'email_raw' and 'name'.

    Returns:
        str: GitHub login username on success. None if unresolved.
    """
```

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: I left this like this on purpose as a flag to future us to research this. i was going to fill it in with dict but wasnt super confident in it since im pretty sure augur throws random stuff around.

 i want to be more detailed wiht the structure of dict thats expected but i couldnt infer much from the if statement at the top of this function and gave up since this is just a quick PR that wasnt focused too hard on getting the typings perfect

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: that may require importing optional and that could end up being a functional change that breaks stuff

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: the param was left there because that was what the comments before this fix used. id like to understand where it comes from/why it was there first before we totally remove it

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: i did this on purpose because i was planning to research and do a follow up PR that would actually use the correct method of marking a function as deprecated (thats compatble back to 3.10) so suitably intelligent IDEs will show it as struck out everywhere we use it.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
iGufrankhan: Args:
Contributor (dict): Contributor record containing either  'cmt_author_name' or 'name' as a full name string.
Raises:
    ValueError: If the selected name field contains fewer than two words.
 Returns:
        str: GitHub search API url for the contributor's full name.



beause i found that 


name_field = 'cmt_author_name' if 'commit_name' in contributor else 'name'

    # Deal with case where name is one word or none.
    if len(contributor[name_field].split()) < 2:
        raise ValueError

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
shlokgilda: Sounds good. ` warnings.warn(..., DeprecationWarning)` works all the way back to Python 2, so 3.10 compat won't be an issue there.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
shlokgilda: Makes sense. A TODO comment might be clearer than IDE stubs as a flag though. Something like `contributor: TODO: determine expected dict structure` reads more intentionally than the placeholders. Up to you.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
shlokgilda: Fair enough on wanting to understand the history. What about the `_type_/_description_` stubs in the Args?

shlokgilda (approved):
Looks good mostly barring some style issues. But no blockers as such

MoralCode:
sean previously approved, Shlok just acknowledged his feedback is addressed. bypass merging

PR #3704, opened 2026-02-17
[keyman] Remove exception in Keyman on uninitialized platform

### LINKED ISSUE #3700: keyman crashing with Invalid platform: github_rest
<details>
<summary>Stack Trace</summary>

```
augur-keyman-1  | 2026-02-15 21:24:29,882 - KeyOrchestrator - ERROR - Error during REQUEST
augur-keyman-1  | Traceback (most recent call last):
augur-keyman-1  |   File "/augur/keyman/Orchestrator.py", line 188, in run
augur-keyman-1  |     new_key = self.new_key(request["key_platform"])
augur-keyman-1  |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
augur-keyman-1  |   File "/augur/keyman/Orchestrator.py", line 102, in new_key
augur-keyman-1  |     raise InvalidRequest(f"Invalid platform: {platform}")
augur-keyman-1  | keyman.KeyOrchestrationAPI.InvalidRequest: Invalid platform: github_rest

```
</details>

this was observed when bringing augur back up after deleting some files as a temporary remediaton of #3699

AryanMishra1789:
**Description**
- This PR prevents Keyman from crashing during startup when a key request is received for a platform whose key pools have not yet been initialized.

- After state cleanup or service restarts, workers may request keys before any keys have been published, which previously caused Keyman to raise InvalidRequest and crash.

This PR fixes #3700

**Notes for Reviewers**
This change treats the condition as a recoverable startup scenario rather than an invalid request. The fix is intentionally minimal and only adds a warning log while preventing the crash.

**Signed commits**

### DIFF
--- a/keyman/Orchestrator.py
+++ b/keyman/Orchestrator.py
@@ -12,7 +12,12 @@ if os.environ.get("KEYMAN_DOCKER"):
 
     sys.path.append("/augur")
 
-    conn = redis.Redis.from_url(os.environ.get("REDIS_CONN_STRING"))
+    # Only create a Redis connection if a connection string is provided.
+    redis_url = os.environ.get("REDIS_CONN_STRING")
+    if redis_url:
+        conn = redis.Redis.from_url(redis_url)
+    else:
+        conn = None
 
     # Just log to stdout if we're running in docker
     logger = logging.Logger("KeyOrchestrator")
@@ -151,7 +156,10 @@ class KeyOrchestrator:
             WaitKeyTimeout: If no fresh keys available (includes wait duration)
         """
         if platform not in self.fresh_keys:
-            raise InvalidRequest(f"Invalid platform: {platform}")
+            self.logger.warning(
+                f"Key requested for uninitialized platform '{platform}'; this may occur during startup after state cleanup."
+            )
+            return
 
         if not len(self.fresh_keys[platform]):
             if not len(self.expired_keys[platform]):


keyman/Orchestrator.py
```diff
@@ -12,7 +12,15 @@
...
     sys.path.append("/augur")
 
-    conn = redis.Redis.from_url(os.environ.get("REDIS_CONN_STRING"))
+    # Only create a Redis connection if a connection string is provided.
+    redis_url = os.environ.get("REDIS_CONN_STRING")
+    if redis_url:
+        try:
+            conn = redis.Redis.from_url(redis_url)
+        except Exception:
+            conn = None
```
MoralCode: Im not sure if swallowing/hiding the exception here is the best way to handle this

AryanMishra1789: Thanks for pointing this out. 

The core fix in this PR is the change in new_key() to prevent Keyman from crashing when a key request arrives before platform state is initialized. The try/catch around Redis initialization was added defensively during debugging and isn’t essential to addressing the InvalidRequest crash.

I can remove that change and keep this PR focused solely on the key request handling.

AryanMishra1789:
I have removed the redis try/catch and updated the PR to reflect the requested change.

MoralCode:
Sounds good! i'd like to talk to John (the maintainer who wrote keyman) to see whether removing this exception is likely the best way to handle this underlying issue, or whether something else (like adding a try catch around the invocation in keyclient) would be better

PR #3711, opened 2026-02-18
Add Job to Automatically Clear Unresolved Emails

IsaacMilarky:
**Description**
- Add Method to automatically clear emails in the unresolved table
- Automatically run this method every 48 hours by default
- Add a config option to control interval

**Notes for Reviewers**
Currently being tested

**Signed commits**

### DIFF
--- a/augur/application/config.py
+++ b/augur/application/config.py
@@ -57,6 +57,7 @@ default_config = {
                 "run_facade_contributors": 1,
                 "facade_contributor_full_recollect": 0,
                 "commit_messages": 1,
+                "unresolved_commit_emails_refresh_interval_hours": 48,
             },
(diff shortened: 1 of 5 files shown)


MoralCode:
im a little skeptical about the concept of this change. I see what its going for, but i think id rather do something like:
- add a column for number of attempts made to resolve that contributor, and only ignore the contributor if its greater than some threshold
- introduce a task to explicitly go through and re-attempt to resolve unresolved contributors (since github downtime is a known way that this list can pile up)


IMO automatically clearing this, especially when there are legitimately unresolveable things in there, seems likely to waste a lot more API keys re-attempting to resolve truly unresolveable contributors for relatively little benefit IMO

PR #3715, opened 2026-02-18
[tasks/data_analysis] Guard against missing repo key in contributor_breadth_worker

### LINKED ISSUE #3468: KeyError: 'url' in contributor_breadth_worker when processing GitHub events

devs6186:
## Changeset
- Skip non-public events in the collection loop; GitHub omits the repo key entirely for private/deleted-repo events
- Guard event["repo"] access in process_contributor_events with a .get() check before reading url/name/id fields

## Notes
The GitHub Events API does not guarantee a repo key on every event. Events tied to private or deleted repos come back without it, causing a KeyError crash that stops all contributor breadth collection for that user. Both defensive layers are needed: filtering at ingestion keeps the list clean, the guard in process_contributor_events is a safety net for any non-public event that slips through.

## Related issues/PRs
- Fixes #3468

**Description**
- Added public-event filter and repo key guard in contributor_breadth_worker

This PR fixes #3468

**Notes for Reviewers**
Check that non-public events are skipped cleanly and that the rest of the user loop continues.

**Signed commits**

### DIFF
--- a/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
+++ b/augur/tasks/data_analysis/contributor_breadth_worker/contributor_breadth_worker.py
@@ -103,6 +103,10 @@ def contributor_breadth_model(self) -> None:
         try:
             for event in github_data_access.paginate_resource(repo_cntrb_url):
 
+                # skip non-public events — their 'repo' key may be absent (issue #3468)
+                if not event.get("public", True):
+                    continue
+
                 cntrb_events.append(event)
 
                 event_age = datetime.strptime(event["created_at"], "%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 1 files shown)


MoralCode:
Thanks for your contribution!

Per the comments in the linked issue and discussion on other, earlier PRs addressing that linked issue, Skipping public events is not the correct solution to this problem. Closing

Issue #3457, opened 2025-12-11
augur.tasks.github.events.collect_events: duplicate key value violates unique constraint "issue-insert-unique"


cdolfi:
I think this error will get solved the same way as #3192 

`IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "issue-insert-unique"\nDETAIL: Key (issue_url)=(https://api.github.com/repos/pulp/pulp-ui/issues/1) already exists.\n')`

`Traceback (most recent call last):
  File "/augur/.venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1938, in _exec_single_context
    self.dialect.do_executemany(
  File "/augur/.venv/lib/python3.11/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 794, in do_executemany
    cursor.executemany(statement, parameters)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "issue-insert-unique"
DETAIL:  Key (issue_url)=(https://api.github.com/repos/pulp/pulp-ui/issues/1) already exists.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/augur/.venv/lib/python3.11/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/augur/.venv/lib/python3.11/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/augur/augur/tasks/github/events.py", line 46, in collect_events
    collection_strategy.collect(repo_git, key_auth, core_data_last_collected)
  File "/augur/augur/tasks/github/events.py", line 128, in collect
    self._process_events(events, repo_id)
  File "/augur/augur/tasks/github/events.py", line 167, in _process_events
    update_issue_closed_cntrbs_by_repo_id(repo_id)
  File "/augur/augur/application/db/lib.py", line 555, in update_issue_closed_cntrbs_by_repo_id
    connection.execute(update_stmt, update_data)
  File "/augur/.venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1416, in execute
    return meth(
           ^^^^^
  File "/augur/.venv/lib/python3.11/site-packages/sqlalchemy/sql/elements.py", line 516, in _execute_on_connection
[...]

PR #3716, opened 2026-02-18
[tasks/github] Fix uncommitted UPDATE in update_issue_closed_cntrbs_by_repo_id

devs6186:
## Changeset
- Replace engine.connect() with engine.begin() in update_issue_closed_cntrbs_by_repo_id so the UPDATE on issues.cntrb_id is committed rather than silently rolled back
- Wrap the call site in events.py with try/except so a DB error here does not abort the whole batch of event collection

## Notes
In SQLAlchemy 2.0 engine.connect() uses autobegin but NOT autocommit. Every call to update_issue_closed_cntrbs_by_repo_id was rolling back at context-manager exit, leaving cntrb_id stale. Switching to engine.begin() makes the transaction commit on success and roll back only on exception, which is the correct behavior. The try/except guard in _process_events means a transient failure here logs an error rather than killing the Celery task mid-batch.

## Related issues/PRs
- related to #3457
- Related to #3192

**Description**
- Fixed silent transaction rollback in update_issue_closed_cntrbs_by_repo_id

**Notes for Reviewers**
The engine.connect() -> engine.begin() change is the core fix. Confirm the UPDATE path in lib.py and the error-handling wrapper in events.py look correct.

**Signed commits**

### DIFF
--- a/augur/application/cli/collection.py
+++ b/augur/application/cli/collection.py
@@ -200,7 +200,7 @@ def repo_reset(ctx):
     """
     Refresh repo collection to force data collection
     """
-    with ctx.obj.engine.connect() as connection:
+    with ctx.obj.engine.begin() as connection:
         connection.execute(s.sql.text("""
             UPDATE augur_operations.collection_status 
             SET core_status='Pending',core_task_id = NULL, core_data_last_collected = NULL;
--- a/augur/application/cli/github.py
+++ b/augur/application/cli/github.py
@@ -26,7 +26,7 @@ def update_api_key():
     Get the ratelimit of Github API keys
     """
 
-    with DatabaseEngine() as engine, engine.connect() as connection:
+    with DatabaseEngine() as engine, engine.begin() as connection:
 
         get_api_keys_sql = s.sql.text(
             """
--- a/augur/tasks/data_analysis/insight_worker/tasks.py
+++ b/augur/tasks/data_analysis/insight_worker/tasks.py
@@ -106,7 +106,7 @@ def insight_model(repo_git: str,logger,engine) -> None:
                 AND ri_date < :min_date
     """)
 
-    with engine.connect() as conn:
+    with engine.begin() as conn:
         result = conn.execute(delete_record_SQL, parameters=dict(repo_id=repo_id, min_date=min_date))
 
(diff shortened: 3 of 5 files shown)


MoralCode:
Thanks for the contribution!

Can you clarify how this PR relates to the two issues you linked?

augur/application/db/lib.py
```diff
@@ -566,7 +566,10 @@ def update_issue_closed_cntrbs_by_repo_id(repo_id):
         )
 
     if update_data:
-        with engine.connect() as connection:
+        # engine.begin() auto-commits (or rolls back on exception).
```
sgoggins, line 563: Holy moly. This comment is right. Nice catch.

sgoggins (approved):

devs6186:
> Thanks for the contribution!
> 
> Can you clarify how this PR relates to the two issues you linked?

Sure. The traceback in #3457 shows the crash path going through _process_events to update_issue_closed_cntrbs_by_repo_id. The root cause is that engine.connect() in SQLAlchemy 2.0 uses autobegin but not autocommit — so the UPDATE on cntrb_id was being silently rolled back on every single call, leaving the field perpetually stale. The try/except wrapper means a failure there no longer kills the entire event collection batch for that repo.

The link to #3192 is more indirect — the original reporter of #3457 noted the bug pattern looks similar (data
consistency errors in the collection pipeline caused by incomplete transactions). #3192 is specifically      about PRs, not events, so this PR doesn't close it, just touches the same kind of problem in a different     
path. I kept it as "Related to" rather than "Fixes" for that reason.

augur/application/db/lib.py
devs6186: > Holy moly. This comment is right. Nice catch.

Thanks! Took me a while to trace why the field was always null — the engine.connect() vs engine.begin()      
distinction in 2.0 is easy to miss.

MoralCode: this comment also tracks with a fix that caio identified in 8Knot.

MoralCode:
Taking over this PR since the contributor has abandoned it and the changes it makes are fairly minimal and seemingly salvage-able

PR #3718, opened 2026-02-18
[tasks/git] Add Celery heartbeat to analyze_commits_in_parallel

### LINKED ISSUE #3319: Augur "forgets" about a running task

devs6186:
## Changeset
- Add bind=True to the analyze_commits_in_parallel task decorator
- Call self.update_state(PROGRESS) at each quarter-queue milestone so the worker can process broker heartbeat frames

## Notes
analyze_commits_in_parallel runs a tight Python loop over thousands of commits without yielding to Celery. On repos with many commits this loop runs for minutes, long enough for RabbitMQ to close the connection due to missed heartbeats. When the connection drops, the task success/failure result can never be sent back, leaving the task permanently stuck in started state in Flower and the repo stuck in collecting in the DB. Adding self.update_state() at regular intervals gives the Celery event loop a chance to process heartbeat frames. bind=True is required to access self; callers still use .si(repo_git, True) unchanged.

## Related issues/PRs
- Fixes #3319

**Description**
- Added bind=True and periodic update_state calls to prevent RabbitMQ heartbeat timeout during long commit analysis

This PR fixes #3319

**Notes for Reviewers**
The update_state interval is every quarter of the queue. For very large repos (100k+ commits) this still means minutes between calls. If that proves insufficient, the interval could be reduced.

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -297,15 +297,17 @@ def analyze_commits_in_parallel(self, repo_git, multithreaded: bool)-> None:
     bulk_insert_dicts(logger,pendingCommitMessageRecordsToInsert, CommitMessage, ["repo_id","cmt_hash"])
     facade_bulk_insert_commits(logger,pendingCommitRecordsToInsert)
     """
+    last_heartbeat = datetime.datetime.utcnow()
     for count, commitTuple in enumerate(queue):
-        quarterQueue = int(len(queue) / 4) or 1
-
-        if (count + 1) % quarterQueue == 0:
+        now = datetime.datetime.utcnow()
+        if (now - last_heartbeat).total_seconds() >= 30:
             percent = (count / len(queue)) * 100
-            logger.info(f"Progress through current analysis queue is {percent}%")
-            # Notify Celery of progress so the broker heartbeat is maintained
-            # for long-running repos (issue #3319).
-            self.update_state(state='PROGRESS', meta={'done': count, 'total': len(queue)})
+            logger.info(f"Progress through current analysis queue is {percent:.1f}%")
+            try:
+                self.update_state(state='PROGRESS', meta={'done': count, 'total': len(queue)})
+            except Exception as e:
+                logger.warning(f"Failed to send Celery heartbeat: {e}")
+            last_heartbeat = now
 
         commitRecords, commit_msg = analyze_commit(logger, repo_id, repo_loc, commitTuple)
         if commitRecords:


MoralCode:
Interesting solution - i like where this is heading if the work leading up to it proves to be correct.

What led you to determine that the solution was analyze_commits_in_parallel specifically? the linked issue does discuss that task, but its in the context of a different failure (that has since been addressed) that should have been moved to its own issue sooner.

Is there additional reason to believe that this task is still causing celery to forget/intentionally lose tasks?

sgoggins:
> Interesting solution - i like where this is heading if the work leading up to it proves to be correct.
> 
> What led you to determine that the solution was analyze_commits_in_parallel specifically? the linked issue does discuss that task, but its in the context of a different failure (that has since been addressed) that should have been moved to its own issue sooner.
> 
> Is there additional reason to believe that this task is still causing celery to forget/intentionally lose tasks?

I think the intuition or perhaps empirical experience is correct. `analyze_commits_in_parallel` is the long running method on new repos, and can go on so long that I believe, but do not have the resources to prove, that these jobs are getting lost. I think this makes sense. I'll do a full review now.

augur/tasks/git/facade_tasks.py
```diff
@@ -301,7 +301,11 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
         quarterQueue = int(len(queue) / 4) or 1
 
         if (count + 1) % quarterQueue == 0:
-            logger.info(f"Progress through current analysis queue is {(count / len(queue)) * 100}%")
+            percent = (count / len(queue)) * 100
```
sgoggins, line 304: I think for large repos we probably just want to have this doing an update every 30-60 seconds instead of relying on queue length.

augur/tasks/git/facade_tasks.py
```diff
@@ -301,7 +301,11 @@ def analyze_commits_in_parallel(repo_git, multithreaded: bool)-> None:
...
+            logger.info(f"Progress through current analysis queue is {percent}%")
+            # Notify Celery of progress so the broker heartbeat is maintained
+            # for long-running repos (issue #3319).
+            self.update_state(state='PROGRESS', meta={'done': count, 'total': len(queue)})
```
sgoggins: I want to know what @MoralCode / @ABrain7710  / @IsaacMilarky think of this, but I think the self.update_state should be wrapped in a try/except block so that if for some reason this heartbeat fails we don't halt Augur altogether. I'd need to run it to determine if that does happen, but I think it could happen.

devs6186:
> Interesting solution - i like where this is heading if the work leading up to it proves to be correct.
> 
> What led you to determine that the solution was analyze_commits_in_parallel specifically? the linked issue does discuss that task, but its in the context of a different failure (that has since been addressed) that should have been moved to its own issue sooner.
> 
> Is there additional reason to believe that this task is still causing celery to forget/intentionally lose tasks?

The reason I focused on analyze_commits_in_parallel is that it's the only   
place in the facade pipeline that runs a tight blocking loop over potentially tens of thousands of commits
with zero yield points. Everything else — cloning, fetching, git subprocess calls — has natural I/O breaks.  
This loop is pure Python iteration, so the worker never gets a chance to process broker heartbeat frames. The
fact that shlokgilda confirmed repos stuck in Collecting where actual_commits == expected_commits (task     
completed but success handler never fired) lines up exactly with what you'd expect from a dropped broker     
connection mid-task.

augur/tasks/git/facade_tasks.py
devs6186: > I think for large repos we probably just want to have this doing an update every 30-60 seconds instead of relying on queue length.

based on what MoralCode found in the RabbitMQ discussion and the log pattern, it looks like the broker is killing connections that go quiet for too long, not an actual broker restart. The task keeps running and the commits get written, but the connection drop means the success handler never fires and the DB stays at Collecting. So yes, keeping heartbeats alive during the loop should address the root cause.

For the two code suggestions — both done in the latest commit:

  - Switched from the quarter-queue milestone check to a 30-second wall-clock interval using
  datetime.datetime.utcnow() (already imported). This means large repos get consistent heartbeats regardless of queue size.
  - Wrapped self.update_state() in try/except with a warning log so a transient failure there doesn't abort the
   whole commit loop.

  Let me know if the interval feels too aggressive or if you'd prefer 60 seconds instead.

MoralCode:
given the speed at which this was developed alongside 8+ other prs, i am very suspicious of the quality of this code and would want to do some independent testing to confirm that this task still has issues with not sending heartbeats (and also what the correct way to yield to celery is)

with andrews scheduling rewrite too, I suspect this may become irrelevant before it is actually solved.

closing. IMO theres not enough of a useful code change here to keep a PR open over.

PR #3719, opened 2026-02-18
[db/models] Deduplicate repos by repo_src_id on CLI add

### LINKED ISSUE #3056: Change repo uniqueness to be based on repo_src_id not url
When repositories move like https://github.com/openai/triton to https://github.com/triton-lang/triton, both can be added and cause neither to fully complete collection. If the check before adding to the repository was based on repo_src_id I believe this problem could be prevented

devs6186:
## Changeset
- is_valid_github_repo now returns repo_src_id (GitHub numeric ID) alongside repo_type
- add_cli_repo passes repo_src_id through to insert_github_repo and insert_gitlab_repo
- insert_github_repo checks for an existing Repo row with the same repo_src_id before inserting; returns its repo_id if found

## Notes
When a GitHub repo moves (e.g. openai/triton -> triton-lang/triton), both the old and new URLs can be added via CLI. Because deduplication was URL-only, two separate rows would be created, causing neither to finish collection. The GitHub numeric repo ID is stable across renames and transfers, so we now capture it from the validation API call and check it before inserting. No schema change is needed — repo_src_id already exists on the table; we just start using it for lookups.

## Related issues/PRs
- Fixes #3056

**Description**
- Added repo_src_id capture and pre-insert lookup to prevent duplicate rows for renamed repos

This PR fixes #3056

**Notes for Reviewers**
This only applies to the CLI add path. The web/frontend path was fixed separately in PR #2929.

**Signed commits**

### DIFF
--- a/augur/application/db/models/augur_data.py
+++ b/augur/application/db/models/augur_data.py
@@ -954,7 +954,7 @@ class Repo(Base):
 
                 return False, {"status": f"Github Error: {data['message']}"}
 
-            return True, {"status": "Valid repo", "repo_type": data["owner"]["type"]}
+            return True, {"status": "Valid repo", "repo_type": data["owner"]["type"], "repo_src_id": data["id"]}
         
         return False, {"status": "Failed to validate repo after multiple attempts"}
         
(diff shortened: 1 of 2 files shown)


MoralCode:
seems like an okay change. not reviewing in detail since IMO this needs a deeper look to see exactly where/how the frontend handles this behavior in case there is a chance for code reuse here.

Contributor indicated this PR is abandoned so marking as draft and keeping the stale tag.

PR #3726, opened 2026-02-19
[scripts] Skip blank_confirm loop in non-interactive mode

### LINKED ISSUE #3633: docker compose: make GitHub/GitLab user and key effectively optional

Follow up to https://github.com/chaoss/augur/pull/3300#pullrequestreview-3319439365, it should be possible to make the user and keys effectively optional.

**Potential solutions:**

Untested, but I'm mildly sure that something like
```diff
diff --git a/scripts/install/config.sh b/scripts/install/config.sh
index 0408b0aa0..6134c378d 100755
--- a/scripts/install/config.sh
+++ b/scripts/install/config.sh
@@ -15,6 +15,10 @@ function blank_confirm() {
 
     confirm_placeholder=${!1}
 
+    if [ -t 0 ]; then
+       return
+    fi
+
     while [ -z "${confirm_placeholder}" ]; do
         echo "You entered a blank line, are you sure?"
         read -p "enter 'yes' to continue, or enter the intended value: " confirm_placeholder
```
should do the trick.

**Additional context:**

This is a CHAOSS Con rapid prototyping experiment.

devs6186:
## Changeset
- Added `[ ! -t 0 ]` check in `blank_confirm()` in `scripts/install/config.sh` to skip the confirmation loop when stdin is not a terminal

## Notes
When `docker compose up` runs `config.sh`, stdin is not connected to a TTY. The `blank_confirm` function was looping on `while [ -z "${confirm_placeholder}" ]` waiting for user input that could never arrive, causing the container startup to hang indefinitely.

The fix checks `[ ! -t 0 ]` (stdin is not a terminal) at the top of `blank_confirm` and returns immediately in that case. Interactive runs (normal terminal usage) are completely unaffected since the condition is false when a TTY is present.

This was the solution proposed in the issue itself — tested the logic and confirmed the shell syntax is valid.

## Related issues/PRs
- Fixes #3633

**Description**
- Added a TTY check so blank_confirm doesn't block when running non-interactively (docker compose, CI)

This PR fixes #3633

**Notes for Reviewers**
Four-line addition inside `blank_confirm`. The rest of the function and all callers are unchanged. The guard fires only when stdin is not a TTY, so it doesn't change any interactive behaviour.

**Signed commits**

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -171,7 +171,7 @@ jobs:
           docker compose -f docker-compose.yml up --no-build 2>&1 \
             | (./scripts/ci/await_all.py /tmp/regex_matches.txt \
                 && docker compose -f docker-compose.yml down)
-        timeout-minutes: 3
+        timeout-minutes: 10
         env:
           AUGUR_GITHUB_API_KEY: ${{ secrets.GITHUB_TOKEN }}
           AUGUR_GITHUB_USERNAME: ${{ github.repository_owner }}
@@ -286,7 +286,7 @@ jobs:
           podman compose -f docker-compose.yml up --no-build 2>&1 \
             | (./scripts/ci/await_all.py /tmp/regex_matches.txt \
                 && podman compose -f docker-compose.yml down)
-        timeout-minutes: 3
+        timeout-minutes: 10
         env:
           AUGUR_GITHUB_API_KEY: ${{ secrets.GITHUB_TOKEN }}
           AUGUR_GITHUB_USERNAME: ${{ github.repository_owner }}


.github/workflows/build_docker.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           docker compose -f docker-compose.yml up --no-build 2>&1 \
             | (./scripts/ci/await_all.py /tmp/regex_matches.txt \
                 && docker compose -f docker-compose.yml down)
-        timeout-minutes: 3
```
Ulincsys, line 174: Changing the timeout was not mentioned in your PR description. Was there a reason for this change?

MoralCode:
Per OP's slack messages, it sounds like they were mainly interested in contributing for GSoC and were planning to move on to other projects after being informed Augur isn't participating this year.

@Ulincsys do you think theres still value in having someone else take over this PR to get it merged?

MoralCode:
If this change merges, i think it would be more of a bandaid than a solution. The block and wait for input is a mechanism to gaurantee there is some input and should not happen if the environment variables contain all the correct values.

I think this is part of the solution (i.e. removing interactive code bits when the environment is often non interactive), but i would like to see a more foundational change first (i.e. have at least the gitlab keys fall back to a sensible default value or not be required at all, or maybe allow augur to start up with no keys configured and print a warning to the console or admin page to that effect or something like that - whatever is gong to be most user-friendly and easiest to debug)

MoralCode:
as-is this solution is not complete and the submitter has not responded for several weeks.  the core feedback has been copied to the linked issue so the ideas here can be incorporated in a more complete PR at a later time. Closing

PR #3729, opened 2026-02-20
PR Fixes a runtime issue for tensorflow on OSX

sgoggins:
**Description**
- Tensorflow was not getting installed on OSX (on Apple Silicon)
- Conditional OS based tensorflow library selection in pyproject.toml fixes this issue.


MoralCode:
so youre saying that tensorflow on arm macs requires a separate `tensorflow-macos` dependency?

sgoggins:
Maybe. I'm still working on what's breaking. I think it's one factor. There may also be a psycopg2 issue with celery. Not finished. Draft mode.

PR #3735, opened 2026-02-24
chown the definitions to the rabbit user after our script edits them

### LINKED ISSUE #3734: podman: Rabbit config update script doesnt have permissions to modify the config file
In our rabbitmq container image, we have a script that populates the credentials in the config based on the environment variables available at build time.

I recently saw this error when rebuilding my local instance's container image in podman

```
STEP 15/16: COPY docker/rabbitmq/update_config.py /
--> Using cache 711eede66212dc63de0c796c17811c96cd2dfec5f98c5ab608ea2259c77debd5
--> 711eede66212
STEP 16/16: RUN exec python3 update_config.py
  File "//update_config.py", line 40, in <module>
    with config_file.open("w") as file:
         ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/etc/rabbitmq/definitions.json'
Error: building at STEP "RUN exec python3 update_config.py": while running runtime: exit status 1

```

MoralCode:
**Description**
- This makes a small adjustment to our container build order for rabbitmq such that the permissions are changed after our script that adjusts rabbit config is run

This PR fixes #3734

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/docker/rabbitmq/Dockerfile
+++ b/docker/rabbitmq/Dockerfile
@@ -12,7 +12,6 @@ ARG RABBIT_MQ_DEFAULT_VHOST=augur_vhost
 COPY --chown=rabbitmq:rabbitmq ./docker/rabbitmq/augur.conf /etc/rabbitmq/conf.d/
 
 ADD docker/rabbitmq/definitions.json /etc/rabbitmq/
-RUN chown rabbitmq:rabbitmq /etc/rabbitmq/definitions.json
 
 ADD docker/rabbitmq/advanced.config /etc/rabbitmq/
 RUN chown rabbitmq:rabbitmq /etc/rabbitmq/advanced.config
@@ -23,3 +22,5 @@ RUN apk add --no-cache python3
 COPY docker/rabbitmq/update_config.py /
 
 RUN exec python3 update_config.py
+
+RUN chown rabbitmq:rabbitmq /etc/rabbitmq/definitions.json


docker/rabbitmq/Dockerfile
```diff
@@ -23,3 +22,5 @@ RUN apk add --no-cache python3
 COPY docker/rabbitmq/update_config.py /
 
 RUN exec python3 update_config.py
+
+RUN chown rabbitmq:rabbitmq /etc/rabbitmq/definitions.json
```
sgoggins, line 26: If the issue is that update_config.py cannot write to definitions.json ... just trying to wrap my head around whether or not this "should" work. 

Given this error: 

```
PermissionError: [Errno 13] Permission denied: '/etc/rabbitmq/definitions.json'
Error: building at STEP "RUN exec python3 update_config.py": while running runtime: exit status 
```

it would **seem** like `/etc/rabbitmq`, if owned by the rabbitmq user, would then allow for the definitions.json file to be written. 

I have not tested this, just thinking about the logic I see, @MoralCode .

sgoggins: It would again seem to me that changing permissions on a file that can't be written won't work ... but this may be something I don't understand.

MoralCode: What are you asking?

MoralCode: If the file can't be written due to permissions issues adjusting the permissions seems like it would help

sgoggins: Does the file already exist though?

MoralCode: Yeah, it's in the same folder as the Dockerfile. I think one of the previous Dockerfile steps is that we copy the file into the container before we run this script to make the final adjustments to it.


The problem here is essentially that if we run this chown line before we run the script that changes the file, we change the permissions such that the script can no longer edit the file, causing the crash.

I still need to run some builds with this new ordering of the build steps to make sure that the problem goes away, but I'm pretty confident that this is what's happening.

MoralCode:
can confirm this fixes the build on my machine. will run it while testing other stuff to make sure rabbit behavior is unchanged as anticipated

docker/rabbitmq/Dockerfile
sgoggins: Are you comfortable its working?

MoralCode: yep

MoralCode:
This is a relatively important as it would likely prevent a newcomer from successfully building the container. But its also simple, well tested fix that only affects container users. Merging (bypass), especially since its been pretty well tested and restores the intended behavior of the rabbit config editing script

PR #3737, opened 2026-02-25
Admin changes

Ulincsys:
**Description**
- Add admin dashboard
- Add admin API routes
- Add `admin_required` decorator to protect admin routes

**Notes for Reviewers**

- There's a lot more work to do on this interface, but I'd like to get it into main so it stays up to date instead of lagging behind again.
- Admin users cannot be created via the frontend, and must be added via the `augur user add` command with the `--admin` flag

**Signed commits**


augur/api/routes/admin.py
```diff
@@ -0,0 +1,27 @@
...
+from subprocess import run, PIPE, Popen
+from flask import Response, current_app, jsonify
+
+from augur.application.db.lib import get_value
+from augur.application.logs import AugurLogger
+from augur.api.util import admin_required
+
+logger = AugurLogger("augur").get_logger()
+
+@app.route(f"/{AUGUR_API_VERSION}/admin/shutdown")
```
shlokgilda, line 14: Defaults to GET requests. Shouldn't this be POST only? Would prevent CSRF.

augur/api/routes/admin.py
```diff
@@ -0,0 +1,27 @@
...
+@admin_required
+def shutdown_system():
+    run("augur backend stop-collection-blocking".split(), stdin=PIPE, stdout=PIPE, stderr=PIPE)
+    Popen("augur backend stop", shell=True, stdin=PIPE, stdout=PIPE, stderr=PIPE)
+    
+    return jsonify({"status": "shutdown"})
+
+@app.route(f"/{AUGUR_API_VERSION}/admin/restart")
+@admin_required
```
shlokgilda, line 23: Same as `admin/shudown`. I think this should also be POST only.

augur/api/routes/admin.py
```diff
@@ -0,0 +1,27 @@
...
+def restart_system():
+    Popen("python scripts/control/restart.py", shell=True)
```
shlokgilda, line 25: Can we use absolute path derived from `__file__` or the project root instead of relative path?

augur/api/routes/config.py
```diff
@@ -7,43 +7,52 @@
...
 
-def generate_upgrade_request():
-    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/426
-    response = jsonify({"status": "SSL Required"})
-    response.headers["Upgrade"] = "TLS"
-    response.headers["Connection"] = "Upgrade"
-
-    return response, 426
-
 @app.route(f"/{AUGUR_API_VERSION}/config/get", methods=['GET', 'POST'])
```
shlokgilda, line 21: Why are we allowing POST here?

augur/api/routes/config.py
```diff
@@ -7,43 +7,52 @@
...
-    if not development and not request.is_secure:
-        return generate_upgrade_request()
-
     with DatabaseSession(logger, engine=current_app.engine) as session:
         
         config_dict = AugurConfig(logger, session).config.load_config()
 
     return jsonify(config_dict), 200
 
+@app.route(f"/{AUGUR_API_VERSION}/config/set", methods=['GET', 'POST'])
```
shlokgilda, line 30: Similar to above. Config mutations via GET are a CSRF risk and leak values into server logs, browser history, and referer headers. This should be POST-only with a JSON body.

augur/api/routes/config.py
```diff
@@ -7,43 +7,52 @@
...
+    if not setting or not section or not value:
+        return jsonify({"status": "Missing argument"}), 400
+    
+    with get_session() as session:
+        config = AugurConfig(logger, session)
+        config.add_or_update_settings([result])
+    
+    return jsonify({"status": "success"})
 
 @app.route(f"/{AUGUR_API_VERSION}/config/update", methods=['POST'])
```
shlokgilda, line 53: `update_config` is missing `@admin_required`. The new `set_config_item` above it has admin protection, but this existing endpoint that can overwrite the entire config is unprotected. Should be:

```suggestion
@app.route(f"/{AUGUR_API_VERSION}/config/update", methods=['POST'])
@ssl_required
@admin_required
def update_config():
```

augur/api/util.py
```diff
@@ -155,4 +156,22 @@ def wrapper(*args, **kwargs):
...
+
+def admin_required(func):
+    @login_required
+    @wraps(func)
+    def inner_function(*args, **kwargs):
+        if current_user.admin:
+            return func(*args, **kwargs)
+        else:
+            abort(403)
+    return inner_function
```
shlokgilda, line 169: The `admin_required` decorator has a subtle ordering issue. `@login_required` is applied to `inner_function` *after* `@wraps(func)`, but `login_required` returns its own wrapper that doesn't preserve the metadata set by `@wraps`. This means the final returned function loses `func`'s `__name__` and `__module__`. Maybe we can use this approach: 

```suggestion
def admin_required(func):
    @wraps(func)
    def inner_function(*args, **kwargs):
        if not current_user.is_authenticated:
            return current_app.login_manager.unauthorized()
        if current_user.admin:
            return func(*args, **kwargs)
        else:
            abort(403)
    return inner_function
```

augur/api/view/augur_view.py
```diff
@@ -52,8 +61,21 @@ def internal_server_error(error):
         errout.close()
     except Exception as e:
         logger.error(e)
+        raise e
```
shlokgilda, line 64: `raise e` inside the 500 error handler will crash the handler itself, meaning the user never sees the friendly error page. Is that the intended idea?

augur/api/view/routes.py
```diff
@@ -337,6 +350,14 @@ def dashboard_view():
         ]}
     ]
 
-    backend_config = requestJson("config/get", False)
+    backend_config = AugurConfig(logger, db_session).load_config()
+    
+    with get_session() as session:
```
shlokgilda, line 355: `session` here shadows Flask's `session` import at the top of the file. Not a bug right now but could cause confusion. Consider `db_session` or `db_sess` instead?

augur/templates/admin-dashboard.j2
shlokgilda, line 1: The IDs `stats-section` is reused on three different `<h1>` elements (Stats, User List, Config). Should be `stats-section`, `user-section`, and `config-section` to avoid issues down the line?

shlokgilda, line 1: The `fetch()` calls for restart/shutdown don't specify `{ method: 'POST' }`. They default to GET. If the backend endpoints get fixed to POST-only (as suggested above), these need updating too:
```js
fetch(url, { method: 'POST' }).then(response => { ... });
```

augur/api/routes/admin.py
MoralCode: also is this script going to work for docker installs?

MoralCode:
@shlokgilda some context: i think the point of this PR is mainly to bring the admin page back (it used to be merged but somehow got unmerged)

definitely agree with a lot of the feedback though

im very hesitant about including "start and stop your instance" controls in our mixed manual/docker setup environment

MoralCode:
As discussed in the March 10 maintainers meeting:

Most of these API endpoints, especially for restarting the Instance, assume a non-docker install. Thus this restarting only works on bare metal
This also relies on the jumpstart part if the code still existing (consider removing if exists as its only useful outside docker)

Many of these endpoints are likely disabled anyway and Augur’s core functionality has changed since this was merged


Id like to avoid merging endpoints that are disabled and dont have a future so we dont need a followup PR to remove them later

sgoggins:
> As discussed in the March 10 maintainers meeting:
> 
> Most of these API endpoints, especially for restarting the Instance, assume a non-docker install. Thus this restarting only works on bare metal This also relies on the jumpstart part if the code still existing (consider removing if exists as its only useful outside docker)
> 
> Many of these endpoints are likely disabled anyway and Augur’s core functionality has changed since this was merged
> 
> Id like to avoid merging endpoints that are disabled and dont have a future so we dont need a followup PR to remove them later

I will wait for @Ulincsys to address these points. Have we verified the restart does not work on Docker? I may have missed that discussion point at the last meeting.

Ulincsys:
@sgoggins I'm planning to address the comments on this tonight after I get home.

We did discuss the restart endpoints during the last meeting. They make the assumption that they are not running in Docker, and so should are not suitable for attempting a restart from within Docker

MoralCode:
I see potentially two forward paths:

1. if jumpstart is still present in the code and we intend to keep it, then we can potentially set these routes up to only be present on the wbe server/in the admin UI when we are not running in docker.

2. if jumpstart isnt present/these routes arent enabled anyway, we should remove them and their UI elements since they wont be functional and wont work.


It seems like the benefit of this UI was for admin convenience when augur was not yet dockerized and startup/shutdown was a special sequence of commands that needed remembering/referring to. with docker, it can be as simple as running  `docker compose stop`  from the same folder as augurs containerfile, or `docker stop <one or more container names>` to stop pieces of augur.

I worry that, if we have endpoints that allow instances to be restarted in the UI, that could present a much larger DoS attack surface, or general user inconvenience in multi-admin/multi-tenancy scenarios, than the convenience it provides.

I can see a benefit to controlling augur tasks within the admin dashboard, but im not sure we need shutdown and restart endpoints

MoralCode:
Jumpstart does appear to still be in the repo. last touched 2 years ago.

Im very tempted to exclude it from the docker builds. maybe we can have the code check if jumpstart is available before trying to use it and make these endpoints no-ops on docker?

PR #3738, opened 2026-02-25
Rename csv_utils to _csv_utils and fix imports

### LINKED ISSUE #3627: Bug: augur --help crashes with AttributeError: module 'csv_utils' has no attribute 'cli'
**Description:**
Running `augur --help` crashes with an `AttributeError` because the CLI multicommand loader imports `csv_utils.py` (a utility module) that does not define a `cli` attribute/command.

This prevents users from viewing any help/usage information via the CLI, making basic usage/debugging difficult.

**How to reproduce:**
1. Clone the repo and install Augur
2. Run `augur --help`

**Expected behavior:**
`augur --help` should display the list of available commands, options, and general usage information without crashing.

**Actual behavior / Error:**
AttributeError: module 'augur.application.cli.csv_utils' has no attribute 'cli'

Full traceback typically points to `augur/application/cli/_multicommand.py` in the `list_commands()` or `get_command()` logic, where it dynamically imports every non-underscored `.py` file in `augur/application/cli/` and expects each to have a `cli` object/Click command group.

**Cause (likely):**
`csv_utils.py` is a helper/utility module (likely for CSV-related functions) and not intended as a CLI command group, but `_multicommand.py` blindly includes it.

**Software versions:**
- Augur: v0.92.0
- OS: Ubuntu 24.04.3 LTS

**Additional context:**
Observed after a fresh clone + install with no modifications to CLI files.

Ulincsys:
**Description**
- Rename `csv_utils.py` to `_csv_utils.py`
    - Any PY file in the `augur/application/cli` directory is assumed by Click to be a CLI multicommand submodule, unless that file's name begins with an underscore `_`. The `csv_utils.py` file is not a multicommand submodule, and so must be prefixed with an underscore to exclude it from dynamic loading during multicommand processing.
- Fix util imports
    - The `augur.application.cli.tasks` multicommand submodule previously imported `clear_rabbitmq_messages` and `raise_open_file_limit` from the base `backend` multicommand submodule, but they have since moved to `augur.application.cli._cli_util`

With this patch, running `augur --help` successfully loads all multicommand submodules and displays the global help message.

This PR fixes #3627 

**Signed commits**


shlokgilda (approved):
LGTM

MoralCode:
I dont *like* the idea of renaming a shared utility like this because of a unique issue with click. I didnt see another way until i read:

> but they have since moved to augur.application.cli._cli_util

This makes me think: could we import the needed csv utils in this already-prefixed CLI util file instead and then wherever its needed in the CLI can import it from that already-prefixed file?

Ulincsys:
> I dont _like_ the idea of renaming a shared utility like this because of a unique issue with click. 

Is there a specific objection you have as to why renaming the file is improper?

It doesn't really matter to me how this issue gets resolved, I just took the approach that required the least changes to achieve the desired effect.

MoralCode:
> Is there a specific objection you have as to why renaming the file is improper?

A bad assumption that this utils file existed somewhere else in the code and that we were having the CSV module dictate filenames outside of its module.

I think this is fine for now. Long term I think id like to find a better way to discover click modules (maybe an AST parsing based system to lightly parse each file?), but this is fine for now

PR #3743, opened 2026-02-25
Remove quotes from format string passed to non shell=true subcommand

### LINKED ISSUE #3740: Facade not correctly populating author name in commits table
While investigating #3693, We (the maintainers) noticed that far too many entries in the commits table were missing author names. Since this table has a NOT NULL constraint on this column, "missing" here means "is set to empty string"

**Observations**
- we manually ran the git CLI used to grab the commit log data against a local clone of the augur repo and a specific unresolved commit and confirmed the git CLI was returning the information correctly.
- we didnt see any logic errors that seemed problematic in the text parsing of the author name (i.e. up until it gets added to the dict that presumably gets INSERTed eventually)
    - we did however, notice that there is part of the text parsing code that includes an `author_name or ''` line, which could be what is inserting the empty string, we just werent sure how
- we briefly checked that the column names match between the dict and the table they were going into 
- This seemed to correlate a fair bit with having the committer information be attributed to a bot, but dont read into this too much

MoralCode:
The lack of shell=true means this is just passed through literally, leading to the first line not starting with `author_name` but `'author_name`, creating a bug because the commit parsing checks for `startswith('author_name')`

This PR fixes #3740

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/db/lib.py
+++ b/augur/application/db/lib.py
@@ -525,6 +525,8 @@ def get_unresolved_commit_emails_by_name(name):
         return session.query(UnresolvedCommitEmail).filter_by(name=name).all()
  
 def get_contributors_by_full_name(full_name):
+    if not full_name or full_name == '':
+        return None 
 
     with get_session() as session:
 


Ulincsys (dismissed):

MoralCode:
oh wow oops, looks like i pushed shortly after you approved haha

PR #3754, opened 2026-03-04
Port prior updates to central contributor extractor

MoralCode:
**Description**
in #3695 the contributor profile fields were updated for additional robustness

This only changed one of the many places we build this object due to our current tech debt

this PR updates the central function we built for this purpose to reflect these changes (and some small ones from what i saw in the contributor resolution process

**Notes for Reviewers**
This is part of a chain of PRs. while they should be largely independent,  id suggest merging this first

**Signed commits**

### DIFF
--- a/augur/application/db/data_parse.py
+++ b/augur/application/db/data_parse.py
@@ -645,8 +645,6 @@ def extract_needed_message_data(comment: dict, platform_id: int, repo_id: int, t
     return dict_data
 
 def extract_needed_contributor_data(contributor, tool_source, tool_version, data_source):
-    # Set platform id to 1 since it is a github method
-    platform_id = 1
 
     if not contributor:
         return None


augur/application/db/data_parse.py
```diff
@@ -645,42 +645,52 @@ def extract_needed_message_data(comment: dict, platform_id: int, repo_id: int, t
...
+    # TODO get and store an owner id
 
     contributor = {
             "cntrb_id": cntrb_id.to_UUID(),
             "cntrb_login": contributor['login'],
-            "cntrb_created_at": contributor['created_at'] if 'created_at' in contributor else None,
-            "cntrb_email": contributor['email'] if 'email' in contributor else None,
-            "cntrb_company": contributor['company'] if 'company' in contributor else None,
-            "cntrb_location": contributor['location'] if 'location' in contributor else None,
+            "cntrb_created_at": contributor.get('created_at'),
```
sgoggins, line 665: Why are we removing the conditionals to acount for the possibility of none values? Those are in there because there are contributors on GitHub that don't return those values. Typically ones that have been on the platform from early on.

shlokgilda: The None handling is the same. `.get('key')` returns `None` when the key is missing, same as the old `if 'key' in contributor else None` pattern. Just shorter syntax.

MoralCode: What shlok said.

The .get syntax is easier to understand/maintain and helps reduce some code inconsistency with the ternary if (some were written with a negated condition and swapped cases, leading to the same result with more confusing code)

augur/application/db/data_parse.py
```diff
@@ -645,42 +645,52 @@ def extract_needed_message_data(comment: dict, platform_id: int, repo_id: int, t
...
+    platform_id = 1
 
     if not contributor:
         return None
 
-    cntrb_id = GithubUUID()   
-    cntrb_id["user"] = contributor["id"]
+    cntrb_id = GithubUUID()
+    cntrb_id["user"] = int(contributor["id"])
+    cntrb_id["platform"] = platform_id
```
MoralCode:
```suggestion
```

augur/application/db/data_parse.py
```diff
@@ -645,42 +645,51 @@ def extract_needed_message_data(comment: dict, platform_id: int, repo_id: int, t
     return dict_data
 
 def extract_needed_contributor_data(contributor, tool_source, tool_version, data_source):
+    # Set platform id to 1 since it is a github method
+    platform_id = 1
```
MoralCode:
```suggestion
```

ABrain7710 (approved):

sgoggins (approved):

PR #3756, opened 2026-03-04
remove name check from facade contributor resolution

cdolfi:
**Description**
This change removes checking for name in contributor resolution. There is too much room for collisions 

This PR fixes #

**Notes for Reviewers**
This doesnt need to be merged if it would be better in a different pr/context

**Signed commits**

### DIFF
--- a/augur/tasks/github/facade_github/tasks.py
+++ b/augur/tasks/github/facade_github/tasks.py
@@ -35,7 +35,6 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
             continue
         
         #Check the unresolved_commits table to avoid hitting endpoints that we know don't have relevant data needlessly
-        
             
         unresolved_query_result = get_unresolved_commit_emails_by_email(email)
 
@@ -52,11 +51,11 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
     
         if login == None or login == "":
             logger.warning("Failed to get login from commit hash")
-            # Try to get the login from supplemental data if not found with the commit hash
-            login = get_login_with_supplemental_data(logger, auth,contributor)
+            # Try to get the login from email data if not found with the commit hash
+            login = fetch_username_from_email(logger, auth,contributor)
     
         if login == None or login == "":
-            logger.error("Failed to get login from supplemental data!")
+            logger.error("Failed to get login from email!")
             continue
 
         url = ("https://api.github.com/users/" + login)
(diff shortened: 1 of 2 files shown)


MoralCode:
Did you conclude there just is no other reasonable way to query usernames from the existing data in the contributors table based on a value that we have accessible at the time of resolution (i.e. commit author name and email)

The solution we used for unresolved contributors was to use the email instead. ~~would an email lookup work here? (i suspect this has been covered prior to this function being called, supporting the theory that this is unnecessary)~~ not sure if i thought this through fully - it may not do what i thought

augur/tasks/github/facade_github/tasks.py
```diff
@@ -46,27 +45,17 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
...
         # Try to get the login from the commit sha
         if login == None or login == "":
             login = get_login_with_commit_hash(logger, auth, contributor, repo_id)
     
         if login == None or login == "":
             logger.warning("Failed to get login from commit hash")
-            # Try to get the login from supplemental data if not found with the commit hash
-            login = get_login_with_supplemental_data(logger, auth,contributor)
+            # Try to get the login from email data if not found with the commit hash
+            login = fetch_username_from_email(logger, auth,contributor)
```
MoralCode, line 55: this wont work as written. `fetch_username_from_email` returns the JSON results of the search query. code is needed to parse through that and identify the most likely result (using the provided `score`) - similar logic to this exists somewhere in get_login_with_supplemental_data or the functions it calls.

also the supplemental data function does try a search by email first already, falling back to a name search (arguably worth removing) if that didnt work

augur/tasks/github/facade_github/tasks.py
```diff
@@ -177,11 +163,11 @@ def link_commits_to_contributor(logger, facade_helper, contributorQueue):
...
                 UPDATE commits 
                 SET cmt_ght_author_id=:cntrb_id
+                cmt_author_platform_username = :cntrb_login
                 WHERE 
-                (cmt_author_raw_email=:cntrb_email
-                OR cmt_author_email=:cntrb_email)
+                cmt_author_email=:cntrb_email
                 AND cmt_ght_author_id is NULL
-        """).bindparams(cntrb_id=cntrb["cntrb_id"],cntrb_email=cntrb["email"])
+        """).bindparams(cntrb_id=cntrb["cntrb_id"],cntrb_email=cntrb["email"],cntrb_login=cntrb["login"])
```
MoralCode, line 170: why are we storing the same data in two places? by this point the contributors login should already be in the contributors table - do we need to duplicate it to the commits table?

augur/tasks/github/facade_github/tasks.py
```diff
@@ -94,7 +83,7 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
             "cntrb_id" : cntrb_id.to_UUID(),
             "cntrb_login": user_data['login'],
             "cntrb_created_at": user_data['created_at'],
-            "cntrb_email": user_data['email'] if 'email' in user_data else None,
+            "cntrb_email": user_data['email'] if 'email' in user_data and user_data['email'] is not None else emailFromCommitData,
             "cntrb_company": user_data['company'] if 'company' in user_data else None,
```
MoralCode, line 87: this will cause a merge conflict with some of the PRs i just filed to deduplicate this whole dict block with several other places we also define it.

augur/tasks/github/facade_github/tasks.py
cdolfi: yeah my goal with this is just to change it to be based on the email only, let me fix this better

augur/tasks/github/facade_github/tasks.py
cdolfi: yeah i mentioned this in slack

PR #3757, opened 2026-03-04
deduplicate contributor object creation in core task process_contributors

MoralCode:
**Description**
The Augur repo suffers from massive code duplication. one place this happens is when creating dicts to populate the contributors table.

We already have a function for this that isnt being used.

This updates the `process_contributors` celery task such that it doesnt reinvent the wheel and properly reuses the shared code (which means it will benefit from the improvements in #3754)

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/github/contributors.py
+++ b/augur/tasks/github/contributors.py
@@ -18,7 +18,7 @@ def process_contributors():
 
     logger = logging.getLogger(process_contributors.__name__)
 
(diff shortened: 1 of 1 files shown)


augur/tasks/github/facade_github/core.py
shlokgilda, line 20: We can remove this assignment from here.

Issue #3746, opened 2026-03-02
make contributor_aliases the single source of truth for contributor emails


MoralCode:
I think the aliases table would make more sense as a single source of truth for emails.

- 8Knot queries are more complicated because of needing to compare between the "primary" email and the aliases (having to get the email from two places
    - this could potentially have performance benefits for 8knot to fix this
- we currently have no logic for which email is the "primary one" other than "whichever one happened first"
- this will probably greatly simplify augur collection process (find a new email, check if it exists already, if not insert it into the aliases table).

### Data Migration

data migration for this DB change is simple since we allow duplicate aliases (i see some in our prod now) - we can just add an alias unconditionally for every current contributors primary address, and then delete the column for it in contributors (it is already foreign key referenced)

PR #3758, opened 2026-03-04
deduplicate contributor creation in facade contributor resolution

MoralCode:
**Description**
The Augur repo suffers from massive code duplication. one place this happens is when creating dicts to populate the contributors table.

We already have a function for this that isnt being used.

This updates the facade task responsible for contributor resolution such that it doesn't reinvent the wheel and properly reuses the shared code (which means it will benefit from the improvements in #3754)

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/github/facade_github/tasks.py
+++ b/augur/tasks/github/facade_github/tasks.py
@@ -111,7 +111,7 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
         if not cntrb.get('cntrb_canonical'):
             cntrb['cntrb_canonical'] = emailFromCommitData
         if not cntrb.get('cntrb_email'):
-            cntrb['cntrb_canonical'] = emailFromCommitData
+            cntrb['cntrb_email'] = emailFromCommitData
         
         if not cntrb.get('cntrb_full_name'):
             cntrb['cntrb_full_name'] = name_field


augur/tasks/github/facade_github/tasks.py
```diff
@@ -84,47 +85,16 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
...
-            #"tool_source": interface.tool_source,
-            #"tool_version": interface.tool_version,
-            #"data_source": interface.data_source
-        }
-
+        # extra processing unique to facade based contributor collection
+        if not cntrb.get('cntrb_canonical'):
+            cntrb['cntrb_canonical'] = emailFromCommitData
+        if not cntrb.get('cntrb_email'):
+            cntrb['cntrb_canonical'] = emailFromCommitData
```
shlokgilda: This sets `cntrb_canonical` again instead of `cntrb_email`. Looks like a copy-paste from the line above.

The old code didn't fall back `cntrb_email` to the commit email (only `cntrb_canonical` got that treatment), so this block is either:
- A typo and should be `cntrb['cntrb_email'] = emailFromCommitData` (new behavior, which might be intentional?)
- Redundant with the first if and should just be removed

augur/tasks/github/facade_github/tasks.py
```diff
@@ -84,47 +85,16 @@ def process_commit_metadata(logger, auth, contributorQueue, repo_id, platform_id
         # Get name from commit if not found by GitHub
         name_field = contributor['commit_name'] if 'commit_name' in contributor else contributor['name']
 
+        cntrb = extract_github_contributor(user_data, tool_source, tool_version, data_source)
```
shlokgilda, line 106: Worth noting: `extract_needed_contributor_data` returns `None` when `user_data` is falsy. The old code didn't guard for that. It would just crash on `user_data['id'].` But now the fallback blocks below (lines 91-97) will crash with `TypeError: 'NoneType'` if `extract_github_contributor` returns `None`.

Probably fine in practice since `user_data` comes from `github_data_access.get_resource()` which raises on 404, but might be worth a quick guard:
```suggestion
        cntrb = extract_github_contributor(user_data, tool_source, tool_version, data_source)
        if cntrb is None:
            continue
```

augur/tasks/github/facade_github/tasks.py
```diff
@@ -194,6 +164,10 @@ def link_commits_to_contributor(logger, facade_helper, contributorQueue):
 @celery.task(base=AugurFacadeRepoCollectionTask, bind=True)
 def insert_facade_contributors(self, repo_git):
 
+    tool_source = "Insert Contributors task"
+    tool_version = "2.0"
+    data_source = "Github API"
+
     # Set platform id to 1 since this task is github specific
     platform_id = 1
```
shlokgilda, line 192: We can remove this `platform_id` assignment from here.

augur/tasks/github/facade_github/tasks.py
cdolfi: So there has been a lot of discussion about cntrb_canonical vs cntrb_email. In the current state cntrb_canonical gets populated with the commit email and cntrb_email is populated with the email from commits. After diving into the code a bit more im thinking just add all the commit emails to the alias table and call it a day lol

cdolfi:
```suggestion
           
```

MoralCode: i think i didnt change this after pasting. im intending to set both. deduping the two email fields is a separate pr and db migration issue

augur/tasks/github/facade_github/tasks.py
MoralCode: its probably worth keeping as a reminder that this hasnt been made platform independent yet

augur/tasks/github/facade_github/tasks.py
MoralCode: this will mean fewer emails get populated

cdolfi: Those emails will still go to alias. This would just make it where commits emails will be only stored in the alias table. Makes it where there is only one place to check for the info/ less duplicate data . This is what I meant by my "thoughts on just putting the emails from commits in the contributor alias table? im thinking only do that and not worry about the commits email in regards to the contributor table" comment in slack

sgoggins: Here is a case where we have canonical emails so you can get contributor data from a single row in a single table. 

Testing will be a lot more prone to error if we don't keep cntrb_canonical because it is relied on in a number of places. @MoralCode : You suggested an aliasing strategy in the maintainers call last night so there there's no duplicate data. I'm good with that, or with populating cntrb_email as well.

sgoggins: We need the emails in the contributor rows.

MoralCode: this was not intended to be written this way. the intent here is to populate both

MoralCode: resolved

MoralCode: so the way this discussion is going makes me think its talking about #3746 , making it out of scope for this pr. theres clearly a perspective difference here that I think needs more time to dive into

MoralCode: resolving as not relevant to this pr getting merged

shlokgilda (approved):
LGTM. One note: #3758 depends on this, so this should go in first.

sgoggins:
@MoralCode : There's a merge conflict which I think is just dropping a line you added down because another line was added in a previous PR ... BUT, I don't want to resolve it because I didn't work on it. If you also aren't sure I will dig in.

PR #3763, opened 2026-03-04
Remove name from contributor login resolution

cdolfi:
**Description**
This PR removes name check from contributor login resolution. This was causing commits from generic names to get grouped together under the same cntrb_id 
This PR fixes #

**Notes for Reviewers**

**Signed commits**
- [ ] Yes, I signed my commits.

PR #3765, opened 2026-03-05
remove remaining name-based resolution from contributor resolution

### LINKED ISSUE #3764: commits with authors using generic names get grouped together as the same contributor
Because Augur does a name-based pre-check for existing contributors in process_commit_metadata, it seems like we may be matching anyone with the public name "dave", "adam", etc to the same person.

MoralCode:
**Description**
- This change removes both the name-based database lookup for existing contributors AND a name-only search of the person using GitHub's search API

name-based searches create issues where bugs (such as one previously addressed involving an empty string) can cause many different people to end up associated to the same contributor in augur. 

This PR fixes #3764 preventing an important data corruption bug. 

Thanks to @cdolfi for helping collaborate on this fix. This fix is based on her pr from https://github.com/chaoss/augur/pull/3763 (reimplemented a slightly different way with the same effect and DCO signed off by me) to help keep it focused and easy to test and ship. 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
+++ b/augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
@@ -143,34 +143,6 @@ def create_endpoint_from_commit_sha(logger, commit_sha, repo_id):
     return url
 
 
-def create_endpoint_from_name(contributor):
-    """Try to construct the best url to ping GitHub's API for a username given a full name.
-
-    Args:
-        contributor (_type_): _description_
-
-    Raises:
-        ValueError: raises when name is one word or none.
-
-    Returns:
-        str: url
-    """
-    # Try to get the 'names' field if 'commit_name' field is not present in contributor data.
-    name_field = 'cmt_author_name' if 'commit_name' in contributor else 'name'
-
-    # Deal with case where name is one word or none.
-    if len(contributor[name_field].split()) < 2:
-        raise ValueError
-    cmt_cntrb = {
-        'fname': contributor[name_field].split()[0],
-        # Pythonic way to get the end of a list so that we truely get the last name.
-        'lname': contributor[name_field].split()[-1]
-    }
-    url = 'https://api.github.com/search/users?q=fullname:{}+{}'.format(
-        cmt_cntrb['fname'], cmt_cntrb['lname'])
-
-    return url
-
 def insert_alias(logger, contributor, email):
     """Insert cntrb_id and email of the corresponding record into the alias table.
     Another database call to get the contributor id is needed because it's an


cdolfi:
From my POV looks good to me! Same impact, better implementation

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -143,34 +143,6 @@ def create_endpoint_from_commit_sha(logger, commit_sha, repo_id):
...
-
-    # Deal with case where name is one word or none.
-    if len(contributor[name_field].split()) < 2:
-        raise ValueError
-    cmt_cntrb = {
-        'fname': contributor[name_field].split()[0],
-        # Pythonic way to get the end of a list so that we truely get the last name.
-        'lname': contributor[name_field].split()[-1]
-    }
-    url = 'https://api.github.com/search/users?q=fullname:{}+{}'.format(
```
sgoggins, line 169: The contributors_aliases bug as it has been described to me, and I acknowledge I've heard a number of different perspectives on this, exists as a result of comparing the contributor full name stored in our database to something simple like "Dave" or an empty string in the commit log. And so the issue as I can see it so far is that a bunch of junk aliases can get generated that way. 

This is accessing the search API. I have not seen or heard about how searching for the name is impacting this issue of aliases.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -358,38 +330,12 @@ def get_login_with_supplemental_data(logger, auth, commit_data):
...
-        
-        login_json = GithubDataAccess(auth, logger, "search").get_resource(url)
-
     # total_count is the count of username's found by the endpoint.
+    # This Checks if the email result got anything.
+    # If it fails, dont place it in unresolved yet as we may add more steps later
     if login_json is None or 'total_count' not in login_json:
         logger.error(
-            "Search query returned an empty response, moving on...\n")
+            "Search query returned an empty response. Could not resolve the username from the email. Moving on...\n")
```
sgoggins, line 338: Where are we searching the commits API? I think we need to do that before we throw them into unresolved.

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
```diff
@@ -343,12 +315,12 @@ def fetch_username_from_email(logger, auth, commit) -> dict | None:
 
 
 def get_login_with_supplemental_data(logger, auth, commit_data):
-    """Return the login given commit data using the supplemental data in the commit (email and name).
+    """Return the login given commit data using the supplemental data in the commit (email).
```
sgoggins, line 318: If we are going to make this change we should use the github ID, which is an immutable integer, while the logins can now change. That wasn't the case when we wrote it but it is the case now. I also want @ABrain7710 to review this because the rules we are following need to be fully explained at this point.

sgoggins (changes requested):
It is important at this point to ensure there is clarity about what we are fixing, and the bug described to date involves incorrect matching on full names from the commit log and full names stored in the contributors table, specifically blank names being the most significant one noted. 

Am I wrong in recalling that we already fixed the blank name issue? 

That leaves incorrect mappings on things like "Dave" leading to incorrect aliases. 

This PR is removing access to the search API to find the name, which is a different path of analysis than what has been explained to me to date. 

What I think we have learned through this is that our highest probability of reconciling the two ground truths of identify in Augur: 
- Commit Log : Commit Identity
- GitHub API : All rows in contributors begin here
- The issue arises in the aliases table, which is a map between the two ground truths. 
- The commit API is on GitHub is resolving identities on commits at a 60%-77% rate in my testing, varying heuristically by age of repsitory and Git use practices mostly. 

I do not believe we are checking the commit API right now? Are we? 

I also think @ABrain7710 needs to review this, as he's got a deeper knowledge of our current intentions for not only this component of identity resolution but others as well.

sgoggins (commented):
I removed the "data loss and corruption" tag because it is coarse. The issue we are experiencing is in the contributors_aliases table.

cdolfi:
@sgoggins this also impacts the commits table

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: We already do that in the order you described, both before and after this change

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: yes. this is a separate, less severe but still serious issue we have identified.

For the same reason that matching based on names when checking the unresolved aliases table results in errors and mismatches (an issue that was fixed) we also should not be checking for existing contributors by name or using the search API with just a name (thats what this PR does)

cdolfi: I gave a thorough explanation in slack. This means that commits are getting incorrect cmt_ght_author_id field and therefore linked to the wrong contributor

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
cdolfi: That is in the task.py file below where all of this gets called in process_commit_metadata

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
MoralCode: all of contributor resolution currently works on the username and this change is being made in the scope of "fix things with a small fix to get it working".

I agree with the need to use the github ID and can address that in a followup PR/more thorough reorganizing of this process within augur

MoralCode:
@sgoggins we dont need a label for every single item of data corruption IMO. I think this is a valid use of the catch all data corruption tag

contributor aliases are still data we care about, else it wouldn't be there. This bug means aliases are being incorrectly assigned, which sounds like a textbook definition of data loss

augur/tasks/github/facade_github/contributor_interfaceable/contributor_interface.py
cdolfi: Okay. Do we store that in the contributors table? I am not seeing it. Also from reading the GithubUUID() code I am not seeing it used there either

sgoggins (approved):
LGTM

PR #3769, opened 2026-03-10
Facade credentials augur startup

### LINKED ISSUE #2928: Cache git credentials in facade on augur startup instead of installation

IsaacMilarky:
**Description**
Move credential storing logic to when augur is started instead of storing them on install. 

Also, move the `.git-credentials` file to a temporary file and symbolic link the file to the facade directory. 

This PR fixes #2928 

**Notes for Reviewers**

**Signed commits**

### DIFF
--- a/augur/application/cli/backend.py
+++ b/augur/application/cli/backend.py
@@ -49,7 +49,7 @@ def create_git_credential_file(github_key, gitlab_key):
         print("Found AUGUR_GITLAB_USERNAME env variable")
         gitlab_username = os.environ.get("AUGUR_GITLAB_USERNAME")
     else:
-        gitlab_username = input("Please input your GitHub username")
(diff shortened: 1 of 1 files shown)


MoralCode:
Can you document the core reason for making this change in the PR as well?

As far as i noticed recently, it seems like we are already writing this file.

if we can use the timeout parameter to have git keep these credentials in memory (https://stackoverflow.com/a/35943882/), or some other, more robust secret storage method like GCM (https://stackoverflow.com/questions/36585496/error-when-using-git-credential-helper-with-gnome-keyring-as-sudo/40312117#40312117) that would be better than storing keys on disk IMO.

If we need to have the augur instance hosts come up with a secret value (like other self hosted open source apps have) to use as the credential to unlock the secret store/decrypt credentials, that may help us be a little safer with how we handle git credentials

Issue #3770, opened 2026-03-16
process_contributors consuming memory every hour until OOM


MoralCode:
Our instance was within 10 GB of its 124 GB memory capacity (and thats after adding 8G swap space) and still experiencing processes being killed due to the kernel OOM killer. htop revealed that the processes using the most memory were all core task workers (with a couple of facade ones thrown in).

This could probably use some additional investigation to see if the fix could be similar to what we have seen before in facade.

I also think setting a high memory limit per task and maybe slowly lowering it to something reasonable could also assist with keeping augur resource usage within predictable margins.

MoralCode:
mem usage seems to be deteriorating on an hourly basis post-collection-reset

<img width="537" height="221" alt="Image" src="https://github.com/user-attachments/assets/c13216a1-41cd-4517-98df-63ac352ebafc" />

MoralCode:
Flower shows that collect_contributors is taking up all of the core task slots. Presumably spawning a new one each hour.
<img width="1698" height="392" alt="Image" src="https://github.com/user-attachments/assets/f4909038-9fa7-4c0e-95d4-f9b39c77a11a" />

Confirmed by a htop run from within the augur container, showing matching PIDs
<img width="1665" height="257" alt="Image" src="https://github.com/user-attachments/assets/e965fd76-9522-4118-9edf-536025d9b276" />

Issue #3452, opened 2025-12-10
This function is one of the worst for database leaks


MoralCode:
https://github.com/chaoss/augur/blob/34f5644e2738c91d2f6b2a4ce81035fa201d9094/augur/tasks/init/__init__.py#L7

every time this is called, a DB connection is opened that never gets closed. This should not be happening.
This has been looked at by several maintainers several times

We need tests and stuff to confirm this

sgoggins:
I think co-pilot found the missing indentation for this problem: 

https://github.com/chaoss/augur/pull/3456/commits/2f1a56142690a69e46e039e3be82374033051b12

MoralCode:
I really dont think thats the answer

MoralCode:
My plan:

1. Wait for database stuff to be fixed/merged
2. Fix database test fixtures in unit tests
3. Create unit test confirming one connection is leaked every time the function in this issue runs
4. Experiment with fixes until the test passes

SOHAMPAL23:
Hey @MoralCode  can I work on this issue

MoralCode:
This issue is currently on hold until some other changes land first. Its also a very deeply technical change and is not going to be very friendly for a newcomer.

MoralCode:
https://docs.sqlalchemy.org/en/20/orm/session_basics.html

This seems like useful prerequisite understanding to evaluate this issue

PR #3771, opened 2026-03-16
inject augur and version into engine connection args

MoralCode:
When diagnosing #3770, i investigated the seemingly high number of idle postgres tasks. while confirming whether these were a problem, i used dbeaver to view the open sessions on the database. This view in dbeaver has a column for application name (a couple of which were populated with dbeavers own name, for its own db connections).

I think setting augur to identify its own connections this way ( which can be done using https://stackoverflow.com/questions/15685861/setting-application-name-on-postgres-sqlalchemy) is a small change that can would help debugging efforts, especially when multiple things (grafana stack, augur, dbeaver or other user-preferred database tools, researchers) can be accessing an augur database at a time. This may also lightly help debug our "connection leaks" issue (#3452 etc)

This PR adds the string `augur` to this field by passing it as a connection argument when we create a database engine.

This string may also be followed by some other text indicating which part of augur the database connections are coming from (such as api, cli, etc). While I initially tried to include the version number, This turned out to not be possible as testing revealed that the `metadata` file this version is in is not an actual module that is accessible

**Notes for Reviewers**
Currently testing this in my local dev instance

**Signed commits**

### DIFF
--- a/augur/application/db/__init__.py
+++ b/augur/application/db/__init__.py
@@ -4,8 +4,6 @@ from contextlib import contextmanager
 
 from augur.application.db.engine import create_database_engine, get_database_string
 
-from metadata import __version__
-
 engine = None
 Session = None
 
@@ -14,7 +12,7 @@ def get_engine():
 
     if engine is None:
         url = get_database_string()
-        engine = create_database_engine(url=url, poolclass=StaticPool, connect_args={"application_name": f"augur v{__version__}"})  
+        engine = create_database_engine(url=url, poolclass=StaticPool, connect_args={"application_name": f"augur"})  
         Session = sessionmaker(bind=engine)
     
     return engine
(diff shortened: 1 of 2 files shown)


MoralCode:
Screenshot_20260317_145646
can confirm it works

augur/application/db/__init__.py
```diff
@@ -42,7 +42,7 @@ def get_session():
 def temporary_database_engine():
 
     url = get_database_string()
-    temporary_database_engine = create_database_engine(url=url, poolclass=StaticPool)  
+    temporary_database_engine = create_database_engine(url=url, poolclass=StaticPool, connect_args={"application_name": f"augur v{__version__} temporary"})
```
shlokgilda, line 45: Is `__version__` defined or imported anywhere in this file? I don't see it. Wouldn't this raise a NameError at runtime?

MoralCode: Oh that should probably be removed if it hasn't already. I gave up importing the version except where already available because its in the metadata.py, outside the module structure and therefore hard to properly import

shlokgilda: Yeah! Probably should remove this

PR #3792, opened 2026-03-25
[facade] fix contributor resolution permanently skipping commits with null cmt_ght_author_id

### LINKED ISSUE #3779: current insert_facade_contributors logic not linking previously resolved contributors
https://github.com/chaoss/augur/blob/49a008ab97c43472339e400cb316a5323110d78d/augur/tasks/github/facade_github/tasks.py#L210-L246

**Query error**
this query, per the comments and what i understand of the implementation is getting all of the commit data's emails and names from the commit table that do not appear in the contributors table or the contributors_aliases table.

This works fine for new contributors that are not yet in the contributors table.

Where this fails is if a record slips through (maybe commits were made using an email that later got linked to a github account). In that case, the records with NULL `cmt_ght_author_id` never get revisited and properly linked to a contributor that IS resolved, just not for the commits that slipped through.

**Last Collection Date**

since we added a filter for last collection date (added by @IsaacMilarky added it in 8539825bb217c388735dfa1bc43d25dc4cee0d51, PR #3253 ), this problem is worse, since anything that slipped through or didnt get properly linked will now be systematically ignored by its older last collection date.

When this PR was filed, I called out that the change it was making seemed unrelated to the analyze_commits_in_parallel problem that was being solved at the time (https://github.com/chaoss/augur/pull/3253#issuecomment-3357922494).

Since this query is being run as essentially a precondition check to establish what records we should attempt to run contributor resolution on.

currently our logic is to find all email addresses (looking up against two sources, the contributors table and the aliases table, see #3746) haven't been matched yet

That is way less simple than "we should run contributor resolution on all commits we don't currently have linked to a contributor" AKA "is the `cmt_ght_author_id` NULL?"

mn-ram:
## Changeset
- Replace the `new_contrib_sql` query in `insert_facade_contributors` — the old query selected unresolved commits by checking whether the commit's email was **absent** from the `contributors`/`contributors_aliases` tables, gated behind a `last_collection_date` window. Now uses `cmt_ght_author_id IS NULL` as the sole eligibility signal, excluding known-dead emails via `unresolved_commit_emails`.
- Replace the `resolve_email_to_cntrb_id_sql` query — the old query had the same date window. New version uses a CTE that unions all three contributor email columns (`cntrb_email`, `cntrb_canonical`, `alias_email`), deduplicates with `DISTINCT ON`, then inner-joins against commits where `cmt_ght_author_id IS NULL AND repo_id = :repo_id`.
- Remove the `CollectionStatus` lookup block and `last_collected_date` variable — no longer needed after removing both `since_date` bind params.
- Remove the now-unused `get_session, execute_session_query` import.

## Notes
The old email-NOT-EXISTS logic had two compounding failure modes:

1. **Silent permanent skip** — if a commit email was *later* linked to a GitHub account (contributor or alias row inserted after the commit was collected), the email now EXISTS in the contributors table and the commit silently falls out of the selection set. `cmt_ght_author_id` stays `NULL` forever with no error raised.
2. **Date window makes it permanent** — the `last_collection_date` cutoff added in PR #3253 meant any commit that slipped through with an older `data_collection_date` was excluded on *every* subsequent run, making recovery impossible without a manual full-recollect flag.

`cmt_ght_author_id IS NULL` is the only semantically correct question: "has this commit been linked to a contributor yet?" It's also what enables the self-recovery path described in #3779 — setting `cmt_ght_author_id = NULL` on corrupted rows (from #3740) is now sufficient to re-queue them on the next collection cycle.

The `unresolved_commit_emails` exclusion in query 1 preserves the existing short-circuit: emails that have already been confirmed unresolvable via GitHub API aren't re-queried needlessly.

## Related issues/PRs
- Fixes #3779
- Enables self-recovery from secondary effects of #3740
- Supersedes the `since_date` change introduced in #3253 for this specific query

### DIFF
--- a/augur/tasks/github/facade_github/tasks.py
+++ b/augur/tasks/github/facade_github/tasks.py
@@ -212,7 +212,7 @@ def insert_facade_contributors(self, repo_git):
                 commits.repo_id = :repo_id
                 AND commits.cmt_ght_author_id IS NULL
                 AND commits.cmt_author_raw_email NOT IN (
-                    SELECT email FROM unresolved_commit_emails
+                    SELECT email FROM augur_data.unresolved_commit_emails
                 )
             ORDER BY hash
     """).bindparams(repo_id=repo_id)
@@ -276,11 +276,11 @@ def insert_facade_contributors(self, repo_git):
         )
         SELECT
             d.cntrb_id,
-            d.email
+            c.cmt_author_email AS email
         FROM
             commits c
         INNER JOIN
-            deduplicated d ON c.cmt_author_raw_email = d.email
+            deduplicated d ON c.cmt_author_email = d.email
         WHERE
             c.cmt_ght_author_id IS NULL
             AND c.repo_id = :repo_id


augur/tasks/github/facade_github/tasks.py
```diff
@@ -278,37 +255,46 @@ def insert_facade_contributors(self, repo_git):
...
     logger.debug("DEBUG: Got through the new_contribs")
     
-    # sql query used to find corresponding cntrb_id's of emails found in the contributor's table
-    # i.e., if a contributor already exists, we use it!
+    # Build the full email→contributor mapping by unioning all three email
+    # columns (cntrb_email, cntrb_canonical, alias_email) then deduplicating.
+    # Only target commits that are still unlinked (cmt_ght_author_id IS NULL)
+    # so we never re-process already-resolved records. Removing the
+    # last_collection_date guard means historical commits that slipped through
+    # on first pass are finally eligible for resolution.
```
MoralCode: this comment is very verbose and describes almost line by line what the below query does. Can you make the comment more high level so it complements, rather than duplicating, the understanding someone would get by reading the code?

augur/tasks/github/facade_github/tasks.py
```diff
@@ -278,37 +255,46 @@ def insert_facade_contributors(self, repo_git):
...
+            SELECT alias_email AS email, cntrb_id
+            FROM contributors_aliases
+            WHERE alias_email IS NOT NULL
+        ),
+        deduplicated AS (
+            SELECT DISTINCT ON (email) email, cntrb_id
+            FROM email_to_contributor
+            ORDER BY email
+        )
         SELECT DISTINCT
```
MoralCode: the suggested query cali wrote in the issue doesnt have a distinct here. Can you elaborate on why this was added to the query?

augur/tasks/github/facade_github/tasks.py
```diff
@@ -196,54 +196,31 @@ def insert_facade_contributors(self, repo_git):
...
-
-    # Get all of the commit data's emails and names from the commit table that do not appear
-    # in the contributors table or the contributors_aliases table.
+    # Find all commits not yet linked to a contributor. The correct signal for
+    # "needs resolution" is a NULL cmt_ght_author_id — not a date window or an
+    # email-table cross-check. The old email-join approach silently skipped
+    # commits whose emails were later linked to a GitHub account, and the
+    # last_collection_date cutoff (PR #3253) made that permanent. Commits
+    # already marked unresolvable are excluded via the unresolved_commit_emails
+    # table so we don't hammer the GitHub API on known dead-ends.
```
MoralCode: This comment is fairly verbose - could we simplify it?

augur/tasks/github/facade_github/tasks.py
mn-ram: Removed `DISTINCT` from the outer `SELECT` — the `deduplicated` CTE already ensures one `cntrb_id` per email via `DISTINCT ON (email)`, so the outer `DISTINCT` was redundant.

augur/tasks/github/facade_github/tasks.py
mn-ram: Simplified the comment.

augur/tasks/github/facade_github/tasks.py
mn-ram: Simplified the comment.

mn-ram:
Hi @sgoggins, I’ve addressed all of MoralCode’s review feedback and also fixed the schema prefix for unresolved_commit_emails and corrected the email column to cmt_author_email. All checks are passing now — would appreciate a re-review. Thanks!

MoralCode:
@mn-ram thanks for addressing the feedback so quickly.

I may have been a little premature to review your PR since I have been actively working on resolving the same issue that this PR solves.

I have pulled some of the changes you made here that I hadnt gotten to yet, such as being a little more specific about the schema being used in these queries, and the text of some of your comments, and have pulled them over to my PR #3797 with you credited as a co-author.

I'm going to close this PR so we don't end up duplicating any more work from each other. Thanks for your contribution though!

PR #3793, opened 2026-03-25
fix: raise Ignore() when issues are disabled to prevent chord cascade failure

### LINKED ISSUE #3461: re-check handling of ResourceGoneException
I introduced a ResourceGoneException for when augur tries to query issues when a repo has them disabled

It seems like this may not being handled.

mn-ram:
## Summary

Fixes #3461

\`ResourceGoneException\` was introduced in commit \`04ef1ef\` and correctly excluded from Celery's retry policy in \`_decide_retry_policy\`. However it was never given a graceful skip path at the task level, so any repo with GitHub Issues disabled would silently abort its entire core collection phase — not just issue collection.

**This PR applies one fix — catch \`ResourceGoneException\` and raise \`Ignore()\` — across three tasks that all access Issues-dependent endpoints:**

| Task | Endpoint | Problem before this fix |
|---|---|---|
| \`collect_issues\` | \`repos/{owner}/{repo}/issues\` | Caught by bare \`except Exception\`, returned \`-1\` |
| \`collect_events\` | \`repos/{owner}/{repo}/issues/events\` | No exception handling at all |
| \`collect_github_messages\` | \`repos/{owner}/{repo}/issues/comments\` | No exception handling at all |

All three endpoints return HTTP 410 Gone when Issues are disabled on a repo — so all three need the same fix. They are grouped in one PR because the change is identical in each file and splitting them would just produce three single-line PRs for the same root cause.

Because these tasks run inside a Celery chord/chain, an unhandled exception in any one of them aborted the entire collection phase — commits, pull requests, contributors, and releases were all skipped for the affected repo. Raising \`Ignore()\` lets Celery treat the task as intentionally skipped so the rest of the chord continues normally.

**Changes:**
- Catch \`ResourceGoneException\` in \`collect_issues\`, \`collect_events\`, and \`collect_github_messages\` and raise \`celery.exceptions.Ignore()\`, following the same pattern already used for \`RepoGoneException\` → \`Reject()\` in \`detect_move/tasks.py\`
- Add unit tests for \`_decide_retry_policy\` to explicitly verify that \`ResourceGoneException\` is excluded from retries
- Add unit tests for all three tasks confirming \`Ignore()\` is raised when the GitHub API returns 410

## Test Plan

- [ ] \`pytest tests/test_tasks/test_task_utilities/test_paginators/test_github_data_access.py -v\`
- [ ] \`pytest tests/test_tasks/test_github_tasks/test_issues.py -v\`
- [ ] Manually add a repo with GitHub Issues disabled (e.g. a docs mirror) and confirm commits, PRs, and contributors still collect while the three tasks log a \`WARNING\` and no \`ERROR\` status is set on the repo

### DIFF
--- a/augur/tasks/github/messages.py
+++ b/augur/tasks/github/messages.py
@@ -24,24 +24,23 @@ def collect_github_messages(repo_git: str, full_collection: bool) -> None:
 
     logger = logging.getLogger(collect_github_messages.__name__)
 
-    try:
-        with GithubTaskManifest(logger) as manifest:
+    with GithubTaskManifest(logger) as manifest:
 
-            augur_db = manifest.augur_db
+        augur_db = manifest.augur_db
 
-            repo_id = augur_db.session.query(Repo).filter(
-                Repo.repo_git == repo_git).one().repo_id
+        repo_id = augur_db.session.query(Repo).filter(
+            Repo.repo_git == repo_git).one().repo_id
 
-            owner, repo = get_owner_repo(repo_git)
-            task_name = f"{owner}/{repo}: Message Task"
+        owner, repo = get_owner_repo(repo_git)
+        task_name = f"{owner}/{repo}: Message Task"
 
-            if full_collection:
-                core_data_last_collected = None
-            else:
-                # subtract 2 days to ensure all data is collected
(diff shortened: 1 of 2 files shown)


MoralCode:
this seems like a PR that does a lot of different things

Lets chat about this in more detail over on the [CHAOSS Slack](https://chaoss.community/kb-getting-started/) in the #wg-augur-8knot channel

PR #3796, opened 2026-03-26
reschedule clone_repos regardless of how many repos were found

### LINKED ISSUE #3321: Facade is not picking up jobs until Augur is restarted following core collection
> This has been the case for years, I believe... and is probably something that should have an issue open if we don't already have one.
- @sgoggins in [slack](https://chaoss-workspace.slack.com/archives/C0226ELG6R4/p1761080449039899?thread_ts=1760713058.666319&cid=C0226ELG6R4)

to me it sounds like the repro steps for this are:
1. start augur fresh
2. add a new repo via the UI
3. wait for core collection to finish
4. observe secondary collection finish, but facade does not start
5. restart augur
6. observe facade collect on the newly added repo

i believe the expected behavior is that facade runs parallel with secondary (assuming sufficient available workers)

MoralCode:
**Description**
This PR fixes #3321 - a bug where, especially on smaller instances, facade appears to not clone repos until the instance is restarted after core and facade are ready.

This was due to an indentation bug with the self-rescheduling behavior of this task where it only reschedules itself if it processed at least one repo.

given how this is implemented, we could maybe simplify it further by making this a properly scheduled task alongside the same lines as collection_monitor, rather than DIYing it

**Notes for Reviewers**
both the conversation in the issue and AI helped diagnose this issue but all code is human written

**Signed commits**

### DIFF
--- a/augur/tasks/git/facade_tasks.py
+++ b/augur/tasks/git/facade_tasks.py
@@ -384,7 +384,7 @@ def clone_repos():
                 setattr(repoStatus,"facade_status", CollectionState.ERROR.value)
                 session.commit()
 
-            clone_repos.si().apply_async(countdown=60*5)
+    clone_repos.si().apply_async(countdown=60*5)
 
 
 #@celery.task(bind=True)


MoralCode:
i cant test this for a while, but it should be relatively easy to test

PR #3797, opened 2026-03-26
Update contributor resolution preparation query to consider all unresolved contributors

### LINKED ISSUE #3779: current insert_facade_contributors logic not linking previously resolved contributors

MoralCode:
**Description**
This PR applies the suggested query changes to the two SQL queries that govern what commit changes are pulled in for contributor resolution

It improves upon the changes suggested in #3792 and those proposed in the underlying issue by making some small query corrections:
- the canonical email column is called cntrb_canonical
- the name of the email output in the second query is relied upon by later functions, and should not change

This PR fixes #3779

**Notes for Reviewers**
This has been run locally and appears to work (at least in terms of not crashing)

Still waiting to see whether it helps resolve more contributors and get the % of null `cmt_ght_author_id`'s to be much lower than the 10% observed in both large augur instances.
 
**Signed commits**

### DIFF
--- a/augur/tasks/github/facade_github/tasks.py
+++ b/augur/tasks/github/facade_github/tasks.py
@@ -196,8 +196,8 @@ def insert_facade_contributors(self, repo_git):
     repo_id = repo.repo_id
     facade_helper = FacadeHelper(logger)
 
-    # Get all of the commit data's emails and names from the commit table that do not appear
-    # in the contributors table or the contributors_aliases table.
+    # Find commits not yet linked to a contributor (cmt_ght_author_id IS NULL),
+    # skipping emails already marked unresolvable.
 
     logger.info(
     "Beginning process to insert contributors from facade commits for repo w entry info: {}\n".format(repo_id))
@@ -207,7 +207,7 @@ def insert_facade_contributors(self, repo_git):
             commits.cmt_commit_hash AS hash,
             commits.cmt_author_raw_email AS email_raw
         FROM
-            commits
+            augur_data.commits
         WHERE
             commits.repo_id = :repo_id AND
             commits.cmt_ght_author_id IS NULL AND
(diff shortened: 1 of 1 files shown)


sgoggins (dismissed):
For bare metal this would require a new parameter in the database config: 

facade_contributor_full_recollect | True or False.

LGTM

PR #3802, opened 2026-03-27
[detect_move] fix: reset collection status on repo move to prevent scheduler stall

### LINKED ISSUE #3667: all Core tasks stop getting scheduled if there are more than 40 repos that need renaming
Once Augur is brought up Core tasks are only getting scheduled for the first few hours, I believe less than 12 but for sure less than 24 hours. Its usually about 1,500-3,000 repos that get through before it stops scheduling again. If you restart the containers the tasks get scheduled again and move as expected. I have been able to repeat this behavior multiple times. 

Once this behavior was noticed on our instance all of our core repos were outside of the collection interval we have set. We still have about half of our repos outside the collection interval even with restarting it consistently the last few days. 

My suspicion is that this bug was introduced in the last release 

Update: this is happening at exactly 40 retries each time

mn-ram:
## Description

When `detect_github_repo_move_core` or `detect_github_repo_move_secondary` detects a 301-redirected repository, the task previously raised `Retry()` with no arguments. Celery does not forward positional args on a bare `Retry`, so the task would restart with the original (now-stale) `repo_git`, crash, and leave `core_status` stuck in `COLLECTING`. Once all 40 scheduler slots were occupied by these stalled tasks, `augur_collection_monitor` stopped dispatching any new Core collection work — requiring manual container restarts to recover.

This PR fixes the root cause by using `self.retry(args=[new_url], countdown=0, max_retries=1)` so the task restarts immediately with the correct, updated `repo_git`. All downstream tasks in the chain then run against the new URL. No collection-status columns are modified; the existing `on_success` handler manages state cleanup as normal.

Fixes #3667

## Changes

- `augur/tasks/github/detect_move/tasks.py`:
  - Add `bind=True` to both task decorators so `self` is available
  - Replace bare `pass` / `Retry()` with `self.retry(args=[e.new_url], countdown=0, max_retries=1)` on `RepoMovedException`
  - Fall back to `Reject` if `new_url` is not present

## Notes for Reviewers

`self.retry` raises `celery.exceptions.Retry` internally, so the `raise` is the standard Celery pattern. `countdown=0` means the retry is enqueued immediately with no delay. `max_retries=1` prevents an infinite loop if the new URL also returns 301.

## Signed commits

### DIFF
--- a/augur/tasks/github/detect_move/tasks.py
+++ b/augur/tasks/github/detect_move/tasks.py
@@ -4,7 +4,9 @@ from augur.tasks.github.detect_move.core import ping_github_for_repo_move, RepoM
 from augur.tasks.init.celery_app import celery_app as celery
 from augur.tasks.init.celery_app import AugurCoreRepoCollectionTask, AugurSecondaryRepoCollectionTask
 from augur.application.db.lib import get_repo_by_repo_git, get_session
+from augur.application.db.models import Repo
 from augur.tasks.github.util.github_random_key_auth import GithubRandomKeyAuth
+from augur.tasks.util.collection_state import CollectionState
 
 from celery.exceptions import Reject
 
@@ -24,13 +26,23 @@ def detect_github_repo_move_core(self, repo_git : str) -> None:
 
     with get_session() as session:
 
-        #Ping each repo with the given repo_git to make sure
-        #that they are still in place.
         try:
             ping_github_for_repo_move(session, key_auth, repo, logger)
         except RepoMovedException as e:
             if e.new_url:
-                raise self.retry(args=[e.new_url], countdown=0, max_retries=1)
+                # Cancel downstream tasks — they were built with the old URL baked
+                # into their .si() signatures and are not yet on the broker.
+                self.request.chain = None
+
+                # DB already has the new URL; reset status so the scheduler
+                # re-collects this repo with the correct URL in the next cycle.
+                moved_repo = session.query(Repo).filter(Repo.repo_git == e.new_url).first()
+                if moved_repo:
+                    status = moved_repo.collection_status[0]
+                    status.core_status = CollectionState.PENDING.value
+                    status.core_task_id = None
+                    session.commit()
+                return
             raise Reject(e)
         except RepoGoneException as e:
             raise Reject(e)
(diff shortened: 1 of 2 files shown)


augur/tasks/github/detect_move/core.py
```diff
@@ -116,6 +116,20 @@ def ping_github_for_repo_move(session, key_auth, repo, logger,collection_hook='c
...
+
+        statusQuery = session.query(CollectionStatus).filter(CollectionStatus.repo_id == repo.repo_id)
+        collectionRecord = execute_session_query(statusQuery, 'one')
+        setattr(collectionRecord, task_id_field, None)
+        if getattr(collectionRecord, last_collected_field) is not None:
+            setattr(collectionRecord, status_field, CollectionState.SUCCESS.value)
+        else:
+            setattr(collectionRecord, status_field, CollectionState.PENDING.value)
+        session.commit()
+
```
MoralCode: I really dont think we should be changing the task status things as a solution to this problem.

the whole reason for calling retry is to cause the task to start over with the updated  repo name so that all the downstream tasks are run on the correct/newly updated repository name

mn-ram:
Hi @MoralCode, I've addressed your feedback:

* Removed all collection-status changes — `core.py` is untouched by this PR
* Using `self.retry(args=[new_url], countdown=0, max_retries=1)` so the task restarts immediately with the correct URL
* Added unit tests covering the retry and reject paths

Could you take another look when you get a chance? Thanks!

MoralCode:
If the task resterts, does the entire rest of the chain get the new URL?

mn-ram:
Yes, the rest of the chain gets the new URL. `ping_github_for_repo_move`calls`update_repo_with_dict()`, which updates `repo_git` in the DB before the exception is raised. So when `self.retry(args=[new_url])` re-executes the current task, the DB already has the new URL—and all downstream tasks read from the DB, so they automatically pick it up.

MoralCode:
> executes the current task, the DB already has the new URL—and all downstream tasks read from the DB, so they automatically pick it up.

I'm fairly sure that many of the tasks would already have been scheduled in celery, potentially before the move detection ever runs. Was there particular documentation you found that led you to your conclusion?

mn-ram:
You're right — I investigated further and confirmed the issue.

All downstream collection tasks are built with `.si(repo_git)` (Celery's immutable signature), so the old URL is baked into their callbacks at chain-construction time. Since `detect_github_repo_move_core` is the **first** task in the chain, the downstream tasks are stored as callbacks in the task message (not yet on the broker), so `self.retry()` only re-ran the detection step while leaving those stale callbacks in place.

**Fix applied:**

Instead of `self.retry()`, the handler now:
1. Sets `self.request.chain = None` — discards the downstream callbacks before Celery can dispatch them (same pattern already used in `task_failed_util`).
2. Resets `core_status` / `secondary_status` to `PENDING` in the DB so the scheduler picks the repo up again in the next cycle with the correct new URL already in the DB.
3. Returns cleanly.

Updated the test to assert `request.chain is None` and `core_status == PENDING` instead of the old retry assertion.

